PR #6, opened 2017-06-07
Use default phabricator api key to query public revisions

purelogiq:
After a long talk with our security folks, we came to a conclusion that having every user of Lando (1000+) enter their Phabricator API keys would be a possible security risk, even if stored in a secure cookie.

This patch allows Lando to use 1 key that it owns itself to query for public revisions on behalf of users. If the user does provide an api key, that key will be used instead, which allows users who have access to secure revisions to only input their API key when needed.

### DIFF
--- a/landoapi/api/revisions.py
+++ b/landoapi/api/revisions.py
@@ -9,7 +9,7 @@ from connexion import problem
 from landoapi.phabricator_client import PhabricatorClient
 
 
-def get(api_key, revision_id):
+def get(revision_id, api_key=None):
     """ API endpoint at /revisions/{id} to get revision data. """
     phab = PhabricatorClient(api_key)
     revision = phab.get_revision(id=revision_id)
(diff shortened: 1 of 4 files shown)


landoapi/spec/swagger.yml
```diff
@@ -34,7 +34,7 @@ paths:
           type: string
           description: |
             A Phabricator Conduit API key to use to get the revision.
```
mars-f: Could you add a sentence here to say that, if this value is left out, a key only capable of working with public revisions will be used instead?

landoapi/api/revisions.py
```diff
@@ -5,11 +5,13 @@
 Revision API
 See the OpenAPI Specification for this API in the spec/swagger.yml file.
 """
+import os
```
mars-f: The `os` module isn't used in the rest of the patch - an unused import?

PR #8, opened 2017-06-08
 Land revision to a stub Transplant API (bug 1372538) 

zalun:


landoapi/api/revisions.py
```diff
@@ -26,6 +27,28 @@ def get(revision_id, api_key=None):
     return _format_revision(phab, revision, include_parents=True), 200
 
 
+def land(revision_id, api_key=None):
+    """ API endpoint at /revisions/{id}/transplants to land revision. """
+    phab = PhabricatorClient(api_key)
+    revision = phab.get_revision(id=revision_id)
+    revision = _format_revision(phab, revision, include_parents=True)
```
purelogiq: Are the parent revisions needed here? Based on the google doc about the transplant service, we can only land 1 patch at a time anyway. Having the extra info doesn't hurt, but, keep in mind that getting the parents is an expensive operation as it could make upwards of 10+ calls to Phabricator for people that like to make chains of very small commits.

landoapi/api/revisions.py
```diff
@@ -26,6 +27,28 @@ def get(revision_id, api_key=None):
...
+        # We could not find a matching revision.
+        return problem(
+            404,
+            'Revision not found',
+            'The requested revision does not exist',
+            type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404'
+        )
+
+    trans = TransplantClient()
+    id = trans.land(
```
purelogiq: Can we make the TransplantClient#land method accept named arguments. Having to know the position of the arguments makes things harder to remember.

landoapi/api/revisions.py
```diff
@@ -26,6 +27,28 @@ def get(revision_id, api_key=None):
...
+        'ldap_username@example.com', revision['repo']['url'], 'patch',
+        'destination', 'push_bookmark', 'http://pingback.url'
+    )
+    return {'request_id': id}, 202
```
purelogiq: Ok, here is a big comment - let me know if I am just misunderstanding something.

I think this is a flaw in how the trello card was designed. Returning a transplant request id will serve no purpose in letting the UI or any other client know the status of the land request. The UI has no database and no way to store state - and so if the UI ever asks to know whether a revision has been landed, it will ask using the revision ID which comes from the URL of the page the user loads (e.g. lando.com/D1234).

The request ID is important for lando-api to have of course, since it can store state and track all the landing statuses internally. Lando-api can save that request_id to its database and later use it to look up its status. But, returning this info via the API won't help anyone, since no one other than lando can talk to the transplant service. Is this API just a intermediate step for to get the card done?

I propose that this endpoint just return a 200 status code if it succeeded to be added to the landing queue, or a bad status code if transplant failed immediately. Then later there will be another endpoint that accept a revision id to see whether that revision has been actually been landed.

Taking this into account the `/revisions/{id}/transplants` route doesn't really make as much sense. Routes such as `/revisions/{id}/land` and `/revisions/{id}/landing_status` would make more sense imo.

cc @mars-f

landoapi/spec/swagger.yml
```diff
@@ -52,7 +52,48 @@ paths:
           schema:
             allOf:
               - $ref: '#/definitions/Error'
+  /revisions/{revision_id}/transplants:
+    post:
+      operationId: landoapi.api.revisions.land
+      description: |
+        Sends request to transplant service and responds with id received
```
purelogiq: See the long comment above about whether this should actually work this way.

landoapi/transplant_client.py
```diff
@@ -0,0 +1,83 @@
...
+    def _request(
+        self, url, data=None, params=None, method='GET', success_code=200
+    ):
+        data = data if data else {}
+        response = requests.request(
+            method=method,
+            url=self.api_url + url,
+            params=params,
+            data=data,
+            timeout=10
```
purelogiq: Too early to know, but, we should ask around to make sure this timeout value isn't too short. I don't know how long transplant takes to do things :)

landoapi/transplant_client.py
```diff
@@ -0,0 +1,83 @@
...
+        request.post(
+            self.api_url + '/autoland',
+            text=json.dumps({
+                'request_id': 1
+            }),
+            status_code=202
+        )
+
+        result = self._POST(
+            '/autoland', {
```
purelogiq: Some of these values aren't needed, e.g. 'rev', 'commit descriptions', etc. See the second page of https://docs.google.com/document/d/1cRIdpVqYcn0Mle7vyjJePNw5TcFoZ20N42dLVy5xVEk/edit#heading=h.rfxbgux0479w for details.

For the ones that aren't needed, lets remove them as arguments to the function. Then others that will never change like `pingback_url`, lets set them here are remove them as arguments to the function.

landoapi/transplant_client.py
```diff
@@ -0,0 +1,83 @@
...
+        commit_descriptions=''
+    ):
+        """ Sends a push request to Transplant API to land a revision.
+
+        Returns request_id received from Transplant API.
+        """
+        # Connect to stubbed Transplant service
+        request.post(
+            self.api_url + '/autoland',
+            text=json.dumps({
```
purelogiq: You can just use `json={...}`, requests will do the dump for you. If you do that then you can remove the import and requirement.txt for json.

landoapi/api/revisions.py
purelogiq: Now that I think about it, it wouldn't really hurt to return that ID, so maybe I am overthinking a small problem!

zalun: But no need indeed :)

landoapi/transplant_client.py
zalun: I'd keep it like that until we will decide. The creation of the job should be quick. It's just adding to the queue.

landoapi/spec/swagger.yml
```diff
@@ -52,7 +52,48 @@ paths:
...
+          type: string
+          description: |
+            The id of the revision to land. This can either be in form of
+            'D12345' (as seen in Phabricator URLs), or just the id number.
+          required: true
+        - name: api_key
+          in: query
+          type: string
+          description: |
+            A Phabricator Conduit API key to use to get the revision.
```
mars-f: Does Lando use a default public API key if this argument is left out?

landoapi/transplant_client.py
```diff
@@ -0,0 +1,90 @@
...
+                'rev': rev,
+                'destination': destination,
+                'push_bookmark': push_bookmark,
+                'commit_descriptions': commit_descriptions,
+                'pingback_url': pingback_url
+            }
+        )
+
+        # Transplant API is responding with a created request_id of the job
+        return result.get('request_id') if result else None
```
zalun: Sure, done

landoapi/transplant_client.py
```diff
@@ -0,0 +1,90 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+import os
+import requests
+import requests_mock
+
+
+class TransplantClient:
```
zalun: It's a to be default Transplant client. Currently mocked as we didn't wanted to make tha request yet.

landoapi/spec/swagger.yml
zalun: yes, if it's the one defined in `docker-compose.yml` as `PHABRICATOR_UNPRIVILEGED_API_KEY=api-123456789`

purelogiq:
This LGTM, the only thing left is to rename the commit message.

PR #9, opened 2017-06-14
Check the status of landing jobs

zalun:


landoapi/api/transplants.py
```diff
@@ -0,0 +1,137 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""
+Revision API
```
zalun: Transplant API :)

landoapi/api/transplants.py
```diff
@@ -0,0 +1,137 @@
...
+            'Landing job not found',
+            'The requested landing job does not exist',
+            type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404'
+        )
+
+    transplant.check_status()
+    return transplant.json(), 200
+
+
+def _format_revision(
```
zalun: This should be removed

docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
       - PHABRICATOR_URL=https://mozphab.dev.mozaws.net
       - PHABRICATOR_UNPRIVILEGED_API_KEY=api-123456789
       - TRANSPLANT_URL=https://stub.transplant.example.com
+      - DATABASE_URL=sqlite:////sqlite.db
```
purelogiq: Not really an issue now, but, at some point we should start using postgres in the development environment since that what will be used in production.

landoapi/api/revisions.py
```diff
@@ -7,30 +7,24 @@
 """
 from connexion import problem
 from landoapi.phabricator_client import PhabricatorClient
-from landoapi.transplant_client import TransplantClient
 
 
-def get(revision_id, api_key=None):
-    """ API endpoint at /revisions/{id} to get revision data. """
+def get_revision(revision_id, api_key=None):
```
purelogiq: I get the idea behind this method, but, in my view it would be better to keep this file to 2 things: 1) the api functions for revisions, 2) helper methods for only those api functions.

If the `transplants.py` needs to get revisions, create a `_get_revision` method within that file specific to its needs.

landoapi/api/revisions.py
```diff
@@ -7,30 +7,24 @@
...
-        return problem(
-            404,
-            'Revision not found',
-            'The requested revision does not exist',
-            type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404'
-        )
+        return None
 
-    return _format_revision(phab, revision, include_parents=True), 200
+    return _format_revision(phab, revision, include_parents=True)
```
purelogiq: If you do make a helper method for `transplants.py` remember that setting include_parents=True is very expensive because it will make several network requests. In fact, if you only need to get the revision id, then you won't need to call_format_revision at all.

landoapi/api/revisions.py
```diff
@@ -7,30 +7,24 @@
...
 
 
-def land(revision_id, api_key=None):
-    """ API endpoint at /revisions/{id}/transplants to land revision. """
-    phab = PhabricatorClient(api_key)
-    revision = phab.get_revision(id=revision_id)
+def get(revision_id, api_key=None):
+    """ API endpoint at /revisions/{id} to get revision data. """
+    revision = get_revision(revision_id, api_key)
```
purelogiq: Let's revert this and have transplants.py use its own private helper method.

landoapi/app.py
```diff
@@ -7,21 +7,35 @@
...
+            'sqlite:////%ssqlite.db' % ('test_' if testing else '')
+        )
+    )
 
+    flask_app.register_blueprint(dockerflow)
+    db.init_app(flask_app)
     return app
 
 
+def create_tables():
```
purelogiq: This function is never used (?)

landoapi/api/transplants.py
```diff
@@ -0,0 +1,55 @@
...
+            'Transplant not created',
+            'The requested revision does exist, but landing failed',
+            type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502'
+        )
+
+    return {'id': transplant.id}, 202
+
+
+def check_status(transplant_id):
+    """ API endpoint at /revisions/{id}/transplants to list landing jobs
```
purelogiq: 2 things. 

This comment doesn't reflect what this function does. The comment says it lists all landing jobs for a revision, but, infact it only retrieves one landing job based on a given id.

Like I mentioned in the last PR. Having an endpoint to get an individual transplant job based on its id will serve no purpose. This is because external services (e.g. Lando UI) should not (or in the case of Lando UI literally cannot) save the transplant ids. The only way for them to query the status of a landing is by revision id. 

I propose these endpoints
`/revisions/{id}/land` - This lands a revision like you have already implemented. The change here is to remove `transplants` from the url because it doesn't really matter for external clients to know the specifics.

`/revisions/{id}/landings` - For each transplant job on that revision returns: ldap username of person that started it, date started, date finished, landing status and target repo. 

This is just a proposal, but, I think we should have a discussion about it @zalun & @mars-f.

landoapi/models/transplant.py
```diff
@@ -0,0 +1,87 @@
...
+TRANSPLANT_JOB_FINISHED = 2
+
+
+class Transplant(db.Model):
+    __tablename__ = "transplants"
+
+    id = db.Column(db.Integer, primary_key=True)
+    request_id = db.Column(db.Integer, unique=True)
+    revision_id = db.Column(db.String(30))
+    status = db.Column(db.Integer)
```
purelogiq: Personally I feel using a string here would be much better, e.g 'started', and 'finished'. 
In the API it is not clear what 0 or 2 means without looking at the source code, for an external client that is very confusing. In addition if we ever have to debug the database it is nice to see text of the statuses instead of a number.

What do you think?

tests/test_revisions.py
```diff
@@ -3,8 +3,12 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 from urllib.parse import parse_qs
 
+import json
```
purelogiq: The imports here on line 6, 9, and 11 are never used (?)

tests/test_transplants.py
```diff
@@ -0,0 +1,113 @@
...
+from tests.canned_responses.phabricator.users import *
+from tests.canned_responses.phabricator.repos import *
+from tests.canned_responses.lando_api.revisions import *
+
+
+@pytest.mark.usefixtures('client_class')
+class TestTransplants:
+    def setup_method(self):
+        conn = _app.create_app('/version.json')
+        with conn.app.app_context():
```
purelogiq: I actually think this is ok, or at least not something we have to worry about for now. Hopefully we find a solution soon :)

purelogiq: https://github.com/kevinbeaty/flask-todomvc/blob/master/tests/todos_tests.py#L17 does something similar.

purelogiq:
I also tested this in the browser and it worked! So I think we're on the right track.

tests/test_transplants.py
zalun: cool, I've been also asking around and people are saying it's OK

tests/test_revisions.py
zalun: yes - leftover

landoapi/models/transplant.py
zalun: Agreed - I was also thinking that it should be changed.

landoapi/api/transplants.py
mars-f: I'd drop the "land" endpoint and just POST directly to `/revision/{id}/landings`.

landoapi/models/transplant.py
mars-f: For the job statuses I suggest we copy the [taskcluster task statuses](https://docs.taskcluster.net/reference/platform/taskcluster-queue/docs/task-life-cycle).  "pending", "running", and "completed", "failed", and possibly "exception".

landoapi/api/transplants.py
purelogiq: I agree :)

purelogiq:
```
•smacleod> mars (/whoever is touching transplant): please see https://docs.google.com/document/d/1cRIdpVqYcn0Mle7vyjJePNw5TcFoZ20N42dLVy5xVEk/edit#heading=h.lteitxoe85mb
2:26:04 PM mars: as someone who has been bit by nesting api endpoints, I highly advise against it (CC imadueme)
2:26:14 PM (re the landings stuff above)
2:26:56 PM e.g. go with /landings/?revision=<blah> vs /revisions/<blah>/landings
2:27:10 PM 
<mars> smacleod: so you would rather have us post to /revision/{id}/landings, and that takes you to /landings/{landing_id} ?
2:27:13 PM 
<•smacleod> keep your ids global, endpoints flat, with query arguments to filter
```
smacleod proposes that we do not use nested api endpoints.

purelogiq: Revised: 
```
All endpoints can have an optional Phabricator API Key in Request Header (which will be used when working with secure revisions.)

Get revision data
GET /revisions 
	?=id
	
Return all the current and past landings for a revision
GET /landings
	?=revision_id
	?=filter=(all,inprogress,failed,etc)
	
Land a revision
POST /landings
	?=revision_id

```

purelogiq: Revised 😤 😃 
```
All endpoints can have an optional Phabricator API Key in Request Header (which will be used when working with secure revisions.)

Get revision data
GET /revisions/{id}
	
Return all the current and past landings for a revision
GET /landings
	?=revision_id
	?=status=(all,inprogress,failed,etc)
	
Land a revision
POST /landings
    Body:
	    revision_id
```

mars-f: @purelogiq I moved the design bit to https://public.etherpad-mozilla.org/p/lando-api-urls

docker-compose.yml
zalun: Confirmed

landoapi/models/landing.py
```diff
@@ -0,0 +1,115 @@
...
+        'date_created': int(revision['dateCreated']),
+        'date_modified': int(revision['dateModified']),
+        'status': int(revision['status']),
+        'status_name': revision['statusName'],
+        'summary': revision['summary'],
+        'test_plan': revision['testPlan'],
+    }
+
+
+class Landing(db.Model):
```
purelogiq: We should store the ldap_username, date_started, and date_landed. We can make another card for it so that we can land this one quicker.

landoapi/spec/swagger.yml
```diff
@@ -84,14 +121,53 @@ paths:
...
     properties:
+      id:
+        type: integer
+        description: |
+          The id of the Landing
       request_id:
         type: integer
         description: |
-          The id of the request in Transplant service
+          The id of the Request in Transplant service
```
purelogiq: I don't think we should expose the transplant request_id to the public. What do you think?

docker/Dockerfile-dev
```diff
@@ -4,7 +4,10 @@
 
 FROM python:3.5-alpine
 
+RUN apk --update --no-cache add \
+	sqlite > /dev/null 2>&1
```
mars-f: I haven't seen the bit about coyping stderr to stdout done in Dockerfiles before, especially for 'apk add'.  Is that a best practice?

docker/Dockerfile-dev
```diff
@@ -4,7 +4,10 @@
...
 ADD requirements.txt /requirements.txt
+ADD docker/db/schema.db /sqlite.db
```
mars-f: Do we need to worry about database file permissions?  In prod the application server runs under a different user.

landoapi/api/landings.py
```diff
@@ -0,0 +1,72 @@
...
+            'Revision not found',
+            'The requested revision does not exist',
+            type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404'
+        )
+    except LandingNotCreatedException:
+        # We could not find a matching revision.
+        return problem(
+            502,
+            'Landing not created',
+            'The requested revision does exist, but landing failed',
```
mars-f: It would be nice if the error message said something like "Please retry your request at a later time."

landoapi/models/landing.py
```diff
@@ -0,0 +1,115 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+from landoapi.models.storage import db
+from landoapi.phabricator_client import PhabricatorClient
+from landoapi.transplant_client import TransplantClient
+
+TRANSPLANT_JOB_STARTED = 'started'
```
mars-f: Could you add a comment explaining that these strings are known values copied from the Transplant service API?

landoapi/models/landing.py
```diff
@@ -0,0 +1,115 @@
...
+        self.request_id = request_id
+        self.revision_id = revision_id
+        self.status = status
+
+    @classmethod
+    def create(cls, revision_id, phabricator_api_key=None, save=True):
+        """ Land revision and create a Transplant item in storage. """
+        revision = _get_revision(revision_id, phabricator_api_key)
+        if not revision:
+            raise RevisionNotFoundException()
```
mars-f: We will need the missing Revision's ID in the exception message for debugging.

landoapi/models/landing.py
```diff
@@ -0,0 +1,115 @@
...
+
+        trans = TransplantClient()
+        request_id = trans.land(
+            'ldap_username@example.com', revision['repo_url']
+        )
+        if not request_id:
+            raise LandingNotCreatedException
+
+        t = cls(
```
mars-f: Could you expand this variable name?  It's too short to understand in context.

landoapi/models/landing.py
```diff
@@ -0,0 +1,115 @@
...
+        )
+        if save:
+            t.save(create=True)
+
+        return t
+
+    @classmethod
+    def get(cls, landing_id):
+        """ Get Landing object from storage. """
+        t = cls.query.get(landing_id)
```
mars-f: Same, could you please expand this variable name?

landoapi/models/landing.py
```diff
@@ -0,0 +1,115 @@
...
+            db.session.add(self)
+
+        db.session.commit()
+        return self
+
+    def __repr__(self):
+        return '<Landing: %s>' % self.id
+
+    def serialize(self):
+        """ Serialize """
```
mars-f: Docstring needs more detail.  What target format are we serializing into?  JSON compatible? sqlite compatible?

landoapi/models/landing.py
```diff
@@ -0,0 +1,115 @@
...
+        return {
+            'id': self.id,
+            'revision_id': self.revision_id,
+            'request_id': self.request_id,
+            'status': self.status
+        }
+
+
+class LandingNotCreatedException(Exception):
+    pass
```
mars-f: Needs a docstring to explain that this exception represents downstream service errors.

landoapi/models/landing.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class LandingNotFoundException(Exception):
```
mars-f: Needs a docstring to explain that this exception represents a specific downstream service response (I think?).

tests/test_landings.py
```diff
@@ -0,0 +1,94 @@
...
+from landoapi.models.landing import (Landing, TRANSPLANT_JOB_STARTED)
+from tests.canned_responses.phabricator.revisions import *
+from tests.canned_responses.lando_api.revisions import *
+from tests.canned_responses.lando_api.landings import *
+
+
+@pytest.mark.usefixtures('client_class')
+@pytest.mark.usefixtures('phabfactory_class')
+@pytest.mark.usefixtures('docker_env_vars')
+class TestLanding:
```
mars-f: These tests should be converted from xUnit-style to pytest-style.

landoapi/spec/swagger.yml
zalun: It's not needed, might be useful for debugging.

docker/Dockerfile-dev
zalun: copied blindly from another Dockerfile, I'll remove it

docker/Dockerfile-dev
zalun: This is not ready for production - I guess it would rather link the directory. Also manager is needed to run create_all on an empty database.

tests/test_landings.py
```diff
@@ -0,0 +1,95 @@
...
+        _db.init_app(app)
+        _db.create_all()
+        # Yield the connexion app so other fixtures can get the app instance
+        # we just created.
+        yield _db
+        _db.session.remove()
+        _db.drop_all()
+
+
+def test_landing_revision(db, client, phabfactory):
```
mars-f: This test name should be something like `test_landing_request_is_saved_to_database()`.

docker/Dockerfile-prod
```diff
@@ -7,6 +7,10 @@ FROM python:3.5-alpine
 RUN addgroup -g 1001 app && \
     adduser -D -u 1001 -G app -s /usr/sbin/nologin app
 
+RUN apk --update --no-cache add \
+	sqlite > /dev/null 2>&1
```
mars-f: The `2>&1` should be removed to match Dockerfile-dev.

tests/test_landings.py
```diff
@@ -0,0 +1,95 @@
...
+from landoapi.models.landing import (Landing, TRANSPLANT_JOB_STARTED)
+
+from tests.canned_responses.phabricator.revisions import *
+from tests.canned_responses.lando_api.revisions import *
+from tests.canned_responses.lando_api.landings import *
+
+
+@pytest.fixture
+def db(app):
+    """Monkeypatch environment variables that we'd get running under docker."""
```
mars-f: Docstring needs rewording, it doesn't match the fixture's purpose.

tests/test_landings.py
```diff
@@ -0,0 +1,95 @@
...
+    with app.app_context():
+        _db.init_app(app)
+        _db.create_all()
+        # Yield the connexion app so other fixtures can get the app instance
```
mars-f: Comment needs removal.

PR #10, opened 2017-06-15
Break the unit test suite's reliance on os.environ

mars-f:
Some unit tests won't run unless os.environ is set.  Mock out the
required environment variables so that the unit tests pass.


tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+
+@pytest.fixture
+def docker_env_vars(monkeypatch):
```
purelogiq: There is a file `tests/utils.py` which contains utility and setup functions that every test uses. Can you move this function to that file? Unless there is a purpose for this file in the near future. I just want to avoid having a bunch of test setup across different files.

tests/test_phabricator_client.py
```diff
@@ -16,6 +16,8 @@
 from tests.canned_responses.phabricator.repos import *
 from tests.canned_responses.phabricator.errors import *
 
+pytestmark = pytest.mark.usefixtures('docker_env_vars')
```
purelogiq: I believe that if you also move this to `test/utils.py` then we won't have to add it at the top of every test file, every test file can just import the utils file as they already do.

tests/test_revisions.py
```diff
@@ -11,6 +11,8 @@
 from tests.canned_responses.phabricator.repos import *
 from tests.canned_responses.lando_api.revisions import *
 
+pytestmark = pytest.mark.usefixtures('docker_env_vars')
```
purelogiq: Same

tests/conftest.py
```diff
@@ -0,0 +1,12 @@
...
+    """Monkeypatch environment variables that we'd get running under docker."""
+    monkeypatch.setenv('PHABRICATOR_URL', 'http://phabricator.test')
```
purelogiq: If you're going to change the phab url, can you also go through and update all the canned responses to use this URL? Right now they use the default mozphab.aws URL which came from the original docker-compose.yml.

tests/conftest.py
purelogiq: Nevermind! I see that conftest.py is a common pattern with pytest.

tests/test_phabricator_client.py
purelogiq: Ignore the specifics above, but, otherwise is there a better way to have this be done automatically for every test?

mars-f: Importing the fixture won't apply it - you need to include a fixture in the module's `pytestmark` to apply it to every test in a file.

tests/conftest.py
mars-f: Done

PR #12, opened 2017-06-19
Use data factories for testing

mars-f:
Using carefully-crafted test data fixtures can lead to brittle tests and
difficult-to-construct test scenarios in our integration tests.

Switch the test suite to use a data factory instead of static fixtures
to mock Phabricator responses.


purelogiq:
Does this invalidate #10?

mars-f:
#10 is still valid.

PR #14, opened 2017-06-22
Use uWSGI to serve the app in production

mars-f:
Run the lando-api application under uWSGI.

To build and run the production image locally add the following to `docker-compose.override.yml`:

```yaml
version: '2'
services:
  lando-api:
    build:
      context: ./
      dockerfile: ./docker/Dockerfile-prod
    ports:
      - "9000:9000"
    environment:
      - UWSGI_WORKERS=1
      - UWSGI_SOCKET=:9001
      - UWSGI_HTTP=:9000
```

### DIFF
--- a/docker/Dockerfile-prod
+++ b/docker/Dockerfile-prod
@@ -4,8 +4,30 @@
 
 FROM python:3.5-alpine
 
-RUN addgroup -g 1001 app && \
-    adduser -D -u 1001 -G app -s /usr/sbin/nologin app
+MAINTAINER mars@mozilla.com
+# These are unlikely to change from version to version of the container
+EXPOSE 9000
+CMD ["/usr/local/bin/uwsgi"]
+
+RUN addgroup -g 10001 app && adduser -D -u 10001 -G app -h /app app
+
+# uWSGI configuration
+ENV UWSGI_MODULE=landoapi.wsgi:app \
+	UWSGI_SOCKET=:9000 \
+	UWSGI_MASTER=1 \
+	UWSGI_WORKERS=2 \
+	UWSGI_THREADS=8 \
+    # Disable worker memory sharing optimizations.  They can cause memory leaks
+    # and issues with packages like Sentry.
+    # See https://discuss.newrelic.com/t/newrelic-agent-produces-system-error/43446
+	UWSGI_LAZY_APPS=1 \
+	UWSGI_WSGI_ENV_BEHAVIOR=holy \
+    # Make uWSGI die instead of reload when it gets SIGTERM (fixed in uWSGI 2.1)
+    UWSGI_DIE_ON_TERM=1 \
+    # Check that the options we gave uWSGI are sane
+    UWSGI_STRICT=1 \
+    # Die if the application threw an exception on startup
+    UWSGI_NEED_APP=1
 
 RUN apk --update --no-cache add \
 	sqlite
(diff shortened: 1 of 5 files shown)


docker/Dockerfile-prod
```diff
@@ -4,22 +4,59 @@
...
+			| sort -u \
+	)" \
+	&& apk add --virtual .python-rundeps $runDeps \
+    && apk del .build-deps
 
 COPY . /app
 RUN pip install --no-cache /app
 
-# run as non priviledged user
+# Run as a non-privileged user
```
purelogiq: Some of our invoke commands (specifically the ones that create the database) stopped working with the addition of these changes. I tracked this down to the `Dockerfile-prod` (since when editing the docker-compose.overrides.yml` file that it what it uses as the context for all invoke commands). The removal of 
```
WORKDIR /app
RUN cd / && pip install --no-cache /app
```
is what causes these changes. 

We either need to re-add them, or we need to 1) confirm that the production environment doesn't need them and 2) confirm that the our circle.yml (which runs our invoke commands for production, doesn't use the production image as the context). 

I tested the app after re-adding those two lines and confirmed that it worked. I recommend that we re-add those two lines.

docker/Dockerfile-prod
```diff
@@ -4,22 +4,59 @@
...
 RUN apk --update --no-cache add \
 	sqlite
 RUN mkdir /db
 RUN chown app:app /db
 
 COPY requirements.txt /requirements.txt
-RUN pip install --no-cache -r /requirements.txt
+
+
+RUN set -ex \
```
purelogiq: This seems excessively complicated to save a couple Docker cache hits. At the very least can you add a comment describing the purpose of this / what it does (besides installing the deps and removing them afterwards). Thoughts?

docker/Dockerfile-prod
```diff
@@ -4,22 +4,59 @@
...
+EXPOSE 9000
+CMD ["/usr/local/bin/uwsgi"]
+
+RUN addgroup -g 10001 app && adduser -D -u 10001 -G app -h /app app
+
+# uWSGI configuration
+ENV UWSGI_MODULE=landoapi.wsgi:app \
+	UWSGI_SOCKET=:9000 \
+	UWSGI_MASTER=1 \
+	UWSGI_WORKERS=2 \
```
purelogiq: Is there a way to let @ckolos and the ops team configure these values? I'm not sure what the best values are for the hardware they will deploy this on.

docker/Dockerfile-prod
mars-f: Reproduced.  I added the `WORKDIR /app` line back in, `invoke upgrade` appears to work again.

docker/Dockerfile-prod
mars-f: They can set these values by editing the Dockerfile, setting them in the Docker environment, or by passing in a uWSGI config file.  I'll let @ckolos decide which is best.  We can iterate on the Dockerfile to find the approach that works best for them.

docker/Dockerfile-prod
mars-f: This stanza is standard in the official Docker image's Dockerfiles.  I've added a comment to the code explaining what it does.

The big thing it does is remove the need to explicitly state a dependency on both package XYZ and XYZ-dev in the Dockerfile, since the 'scanelf' stanza automagically keeps XYZ and removes XYZ-dev.  Otherwise I would have to keep 2 separate lists of packages in the Dockerfile, and keep them in sync, too.  Keeping one list should make the housekeeping easier.  It's fewer lines to read, understand, and maintain.

docker/Dockerfile-prod
ckolos: Env vars are fine; we can override those

docker/Dockerfile-prod
```diff
@@ -15,13 +37,32 @@ RUN chown app:app /db
...
+           pcre-dev \
+    && pip install --no-cache -r /requirements.txt \
+	&& runDeps="$( \
+		scanelf --needed --nobanner --recursive /usr/local \
+			| awk '{ gsub(/,/, "\nso:", $2); print "so:" $2 }' \
+			| sort -u \
+			| xargs -r apk info --installed \
+			| sort -u \
+	)" \
+	&& apk add --virtual .python-rundeps $runDeps \
```
ckolos: *nit: Do you anticipate this list changing often? if not, why not just add these packages as part of the container build?

ckolos: +1 for originality!
;)

mars-f: The list of packages changes every time we add a new compiled dependency.  So far that's been for database adapters, uWSGI, and crypto/SSL handling.

This is the second time a reviewer has said they would be less surprised by an explicit list of runtime packages, and I'm sensing a trend, so I'll look into changing it.  :)  But we might not want to block on the rewrite and make that a follow-up commit.

PR #15, opened 2017-06-23
Add Alembic migrations using flask-alembic

zalun:


docker/Dockerfile-prod
```diff
@@ -12,6 +12,8 @@ RUN apk --update --no-cache add \
 RUN mkdir /db
 RUN chown app:app /db
 
+COPY migrations /migrations
```
mars-f: Does this also need to be applied to Dockerfile-dev?

tasks.py
```diff
@@ -89,12 +89,21 @@ def imageid(ctx):
     )
 
 
-@task(name='create_db')
-def create_db(ctx):
-    """Call `create_all` on a SQLAlchemy database."""
+@task
+def revision(ctx, msg):
```
mars-f: Can we rename this to something more intuitive, like `add-migration`?  Running `invoke add-migration` to start a new migration from a template sounds more intuitive than `invoke revision`, especially if the developer isn't familiar with Alembic terminology.

docker/Dockerfile-prod
zalun: It is linked in docker-compose

PR #16, opened 2017-06-27
Add pingback support (bug 1377305)

zalun:
Added POST API endpoint `/landings/{landing_id}/update`.
`landing_id` (from path) and `request_id` (from data) have to correspond
with database entry, otherwise 404 is raised.
Modified the schema by adding error and result.


landoapi/models/landing.py
```diff
@@ -42,35 +51,45 @@ class Landing(db.Model):
     request_id = db.Column(db.Integer, unique=True)
     revision_id = db.Column(db.String(30))
     status = db.Column(db.Integer)
+    auth = db.Column(db.String(30))
```
mars-f: I don't think the auth mechanism should be stored in the database.  The [security design doc](https://docs.google.com/document/d/1cRIdpVqYcn0Mle7vyjJePNw5TcFoZ20N42dLVy5xVEk/edit#) says we should consider a shared API key for validating the client doing the update, and that means we'll end up adding a second migration at a later to remove the auth column.  I recommend we split this user story into two parts: one for this patch to perform status updates with no auth, and a second story to have a special API key for a pingback client to use to update statuses.

docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
       - PHABRICATOR_UNPRIVILEGED_API_KEY=api-123456789
       - TRANSPLANT_URL=https://stub.transplant.example.com
       - DATABASE_URL=sqlite:////db/sqlite.db
+      - HOST_URL=https://landoapi.mozilla.org
```
purelogiq: Agree that we will need this variable. Lets set to `http://lando-api.test` and in ckolos change it for dev, stage, and prod.

landoapi/spec/swagger.yml
```diff
@@ -121,6 +121,45 @@ paths:
...
+      operationId: landoapi.api.landings.update
+      description: |
+        Sends request to transplant service and responds just the status code
+      parameters:
+        - name: data
+          in: body
+          description: |
+            Retrieve status of the landing job
+          required: true
+          schema:
```
purelogiq: See https://docs.google.com/document/d/1cRIdpVqYcn0Mle7vyjJePNw5TcFoZ20N42dLVy5xVEk/edit# for the remaining data.

landoapi/api/landings.py
```diff
@@ -71,3 +71,18 @@ def get(landing_id):
...
+    if not landing:
+        return problem(
+            404,
+            'Landing not found',
+            'The requested Landing does not exist',
+            type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404'
+        )
+    landing.status = data['status']
+    landing.save()
+    return landing.serialize(), 202
```
purelogiq: Does transplant expect a 202, or 200? 
https://httpstatuses.com/202 vs https://httpstatuses.com/200

landoapi/models/landing.py
```diff
@@ -1,10 +1,13 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+import os
+
 from landoapi.models.storage import db
 from landoapi.phabricator_client import PhabricatorClient
 from landoapi.transplant_client import TransplantClient
 
+TRANSPLANT_JOB_STARTING = 'starting'
```
purelogiq: Just for me, what is the difference between 'starting' and 'started'?

zalun: It is set before request is made. No idea if it is started on the Transplant side.

landoapi/api/landings.py
zalun: I think it just checks for errors

landoapi/models/landing.py
mars-f: > It is set before request is made. No idea if it is started on the Transplant side.

'pending' might be a better word to communicate the status.

landoapi/api/landings.py
```diff
@@ -71,3 +69,25 @@ def get(landing_id):
...
+        return problem(
+            404,
+            'Landing not found',
+            'The requested Landing does not exist',
+            type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404'
+        )
+
+    landing.error = data.get('error_msg', '')
+    landing.result = data.get('result', '')
+    landing.status = TRANSPLANT_JOB_LANDED if data['landed'
```
mars-f: This line confuses me.  If the 'landed' field is present, regardless of value, everything is OK, and if an update is ever made with this field missing, it's a failure in the other system?  I think there should be a comment or two in here to explain what the possible combinations of incoming values are, and to explain what they represent in the other system and in our system.

landoapi/models/landing.py
```diff
@@ -40,39 +49,34 @@ def create(cls, revision_id, phabricator_api_key=None, save=True):
...
         repo = phab.get_revision_repo(revision)
 
+        # save landing to make sure we've got the callback
+        landing = cls(
+            revision_id=revision_id,
+        ).save()
+
         trans = TransplantClient()
+        callback = '%s/landings/%s/update' % (
+            os.getenv('HOST_URL'), landing.id
```
mars-f: This setting's name could be improved.  I can't tell from the name alone that it is used for pingbacks.  Maybe something like `PINGBACK_HOST_URL` or just `PINGBACK_URL`?

landoapi/spec/swagger.yml
```diff
@@ -111,8 +111,70 @@ paths:
...
+            properties:
+              id:
+                type: integer
+                description: |
+                  A newly created Landing id
         404:
-          description: Revision does not exist
+          description:
+            Revision does not exist or its request_id doesn't correspond to
+            the saved one
```
mars-f: Probably best not to mention internal implementation details here.  I don't see a `request_id` mentioned in the POST parameters, either?

landoapi/spec/swagger.yml
```diff
@@ -111,8 +111,70 @@ paths:
...
+        default:
+          description: Unexpected error
+          schema:
+            allOf:
+              - $ref: '#/definitions/Error'
+  /landings/{landing_id}/update:
+    post:
+      operationId: landoapi.api.landings.update
+      description: |
+        Sends request to transplant service and responds just the status code
```
mars-f: Typo, "responds with just the status code"

landoapi/models/landing.py
zalun: it's the lando-api host url - I thought we might use it for something else.

landoapi/spec/swagger.yml
zalun: right - it should stay as it was before
`request_id` is mentioned as "in body" parameter

landoapi/api/landings.py
zalun: I'm adding description of fields which are received from Transplant (from Google doc)

landoapi/models/landing.py
mars-f: Discussed on IRC, I'm fine with both keeping the variable as-is and with changing it.

purelogiq: I think HOST_URL works well, imo.

PR #17, opened 2017-06-30
Disable Alembic migrations during testing

mars-f:
A side-effect of initializing the Alembic database schema migration
system is that it creates directories on the file system.  This
breaks the test suite in some environments and bloats test startup
time.  Since we don't need Alembic to run our acceptance tests, we
replace Alembic with a stub.


mars-f:
:+1:  to merge from @zalun

PR #19, opened 2017-06-30
Use unbuffered stdout for containers

mars-f:
Use unbuffered stdout for all containers.  In dev this makes print()
statements work.  In prod using unbuffered output for daemons is just
plain good practice.

### DIFF
--- a/docker/Dockerfile-prod
+++ b/docker/Dockerfile-prod
@@ -11,6 +11,8 @@ CMD ["/usr/local/bin/uwsgi"]
 
 RUN addgroup -g 10001 app && adduser -D -u 10001 -G app -h /app app
 
+ENV PYTHONUNBUFFERED=1
+
 # uWSGI configuration
 ENV UWSGI_MODULE=landoapi.wsgi:app \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile-prod
```diff
@@ -11,6 +11,8 @@ CMD ["/usr/local/bin/uwsgi"]
 
 RUN addgroup -g 10001 app && adduser -D -u 10001 -G app -h /app app
 
+ENV PYTHONUNBUFFERED=1
```
purelogiq: we don't need this in production. In production we will be using mozlog which doesn't have this problem. (And I'm unsure what this will do in production, do you think it is fine to have in prod??)

purelogiq: Nevermind, I see your comment at the top!

mars-f: Yes, in prod it's even more important to not lose output because of a buffer.  You never know when there will be a message in a buffer somewhere that's critical to debugging a problem, and you'll spend days pulling your hair out trying to figure out why there is no error message before the crash. :smile: :cry:

PR #20, opened 2017-06-30
Return HTTP 502 from failed phabricator heartbeat checks

mars-f:
To test this change, add the following to docker-compose.override.yml:

```yaml
version: '2'
services:
  lando-api:
    environment:
      - PHABRICATOR_URL=https://nosuchhost.test
```

and hit http://<yourdockerhostname>:8888/__heartbeat__.

The container log should output a WARNING level message.  It will not output INFO or DEBUG messages because we don't have the loggers configured for that yet (we should implement mozlogging instead).

### DIFF
--- a/landoapi/dockerflow.py
+++ b/landoapi/dockerflow.py
@@ -7,8 +7,15 @@ specification.  See https://github.com/mozilla-services/Dockerflow for details.
 """
 
 import json
+import logging
 
+import requests
 from flask import Blueprint, current_app, jsonify
+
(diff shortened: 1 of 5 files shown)


ckolos:
I have some concern about monitoring the health of phabricator from this application because phabricator being down doesn't result in a hard-down condition for lando. Whenever we do a phabricator deploy that will involve downtime (should only be schema changes), we will need to expect to receive an alert from lando reflecting that. I'm r+ on this right now with the understanding that we may wish to remove this in the future.

PR #21, opened 2017-07-04
Use mozlog for application logging

mars-f:
Configure the application to use the mozlog structured log format for
 application-level log messages (log messages from libraries and
 frameworks still go through the Python Standard Library string logger).

To test, modify `docker-compose.override.yml` and set the `LOG_LEVEL` environment variable to the string value `debug`.  A message about logging initialization should appear in the container logs.  

Try setting the `LOG_LEVEL` value to `foobar` to see what happens if an unknown/typo'd log level is set.

### DIFF
--- a/landoapi/transplant_client.py
+++ b/landoapi/transplant_client.py
@@ -3,11 +3,15 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import os
+
+import logging
 import requests
 import requests_mock
 
 from sqlalchemy import text
 from landoapi.models.storage import db
+
+logger = logging.getLogger(__name__)
 
 
 class TransplantClient:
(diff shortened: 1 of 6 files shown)


landoapi/transplant_client.py
```diff
@@ -48,8 +52,29 @@ def land(self, ldap_username, hgpatch, tree, pingback, request):
...
+            # some sort.
+            logger.info(
+                {
+                    'service':
+                    'transplant',
+                    'username':
+                    ldap_username,
+                    'msg':
+                    'received an empty response from the transplant service',
+                }, 'transplant.failure'
```
purelogiq: at the end of this line you can add a comment like: `# yapf: disable` to prevent the linter from formating this hash. I don't think this is good style just because the linter says so, it's clearly a bug and we know better :)

mars-f: Will do

PR #23, opened 2017-07-06
Authenticate pingback source

zalun:
Uses API-Key header parameter to authenticate the request.
Adds TRANSPLANT_API_KEY environment variable to compare to the header.

### DIFF
--- a/landoapi/api/landings.py
+++ b/landoapi/api/landings.py
@@ -5,7 +5,11 @@
 Transplant API
 See the OpenAPI Specification for this API in the spec/swagger.yml file.
 """
+import hmac
+import json
 import logging
+import os
+
 from connexion import problem
 from flask import request
 from sqlalchemy.orm.exc import NoResultFound
@@ -15,6 +19,7 @@ from landoapi.models.landing import (
 )
 
 logger = logging.getLogger(__name__)
+TRANSPLANT_API_KEY = os.getenv('TRANSPLANT_API_KEY')
 
 
 def land(data, api_key=None):
@@ -96,6 +101,15 @@ def get(landing_id):
     return landing.serialize(), 200
 
 
+def _not_authorized_problem():
+    return problem(
+        403,
+        'Not Authorized',
+        'You\'re not authorized to proceed.',
+        type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403'
+    )
+
+
 def update(landing_id, data):
     """Update landing on pingback from Transplant.
 
@@ -118,7 +132,33 @@ def update(landing_id, data):
     result: string
         revision (sha) of push if landed == true
         empty string if landed == false
+
+    API-Key header is required to authenticate Transplant API
     """
+    if os.getenv('PINGBACK_ENABLED', 'n') != 'y':
+        logger.warning(
+            {
+                'request_id': data.get('request_id', None),
+                'landing_id': landing_id,
+                'remote_addr': request.remote_addr,
+                'msg': 'Attempt to access a disabled pingback',
+            }, 'pingback.warning'
+        )
+        return _not_authorized_problem()
+
+    if not hmac.compare_digest(
+        request.headers.get('API-Key', ''), TRANSPLANT_API_KEY
+    ):
+        logger.warning(
+            {
+                'request_id': data.get('request_id', None),
+                'landing_id': landing_id,
+                'remote_addr': request.remote_addr,
+                'msg': 'Wrong API Key',
+            }, 'pingback.error'
+        )
+        return _not_authorized_problem()
+
     try:
         landing = Landing.query.filter_by(
             id=landing_id, request_id=data['request_id']
--- a/landoapi/models/landing.py
+++ b/landoapi/models/landing.py
@@ -63,14 +63,18 @@ class Landing(db.Model):
         # save landing to make sure we've got the callback
         landing = cls(revision_id=revision_id, diff_id=diff_id).save()
 
-        trans = TransplantClient()
-        callback = '%s/landings/%s/update' % (
-            os.getenv('HOST_URL'), landing.id
+        # Define the pingback URL with the port
+        callback = '{host_url}:{pingback_port}/landings/{id}/update'.format(
+            host_url=os.getenv('HOST_URL'),
+            pingback_port=os.getenv('PINGBACK_PORT', 80),
+            id=landing.id
         )
+
+        trans = TransplantClient()
         # The LDAP username used here has to be the username of the patch
-        # pusher.
-        # FIXME: what value do we use here?
-        # FIXME: This client, or the person who pushed the 'Land it!' button?
+        # pusher (the person who pushed the 'Land it!' button).
+        # FIXME: change ldap_username@example.com to the real data retrieved
+        #        from Auth0 userinfo
         request_id = trans.land(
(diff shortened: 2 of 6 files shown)


landoapi/spec/swagger.yml
```diff
@@ -135,6 +135,11 @@ paths:
         Sends request to transplant service and responds with just the
         status code
       parameters:
+        - name: API-Key
```
purelogiq: Also what do you think of a more descriptive name like `Transplant-API-Key`?

zalun: For Transplant it's just its own API-Key. We certainly make it more descriptive, just wondering if it's needed.

landoapi/api/landings.py
```diff
@@ -118,6 +121,13 @@ def update(landing_id, data):
         revision (sha) of push if landed == true
         empty string if landed == false
     """
+    if request.headers['API-Key'] != TRANSPLANT_API_KEY:
```
globau: you should be doing constant time string comparison to protect against password timing attacks.
there also needs to be some form of brute force detection/protection.

landoapi/api/landings.py
```diff
@@ -118,6 +121,13 @@ def update(landing_id, data):
...
+        return problem(
+            401,
```
globau: this should be 403; 401 means "unauthenciated" and requires a www-authenticate header in the response.

landoapi/spec/swagger.yml
```diff
@@ -172,6 +177,11 @@ paths:
       responses:
         202:
           description: OK
+        401:
```
globau: 403 here too

landoapi/api/landings.py
purelogiq: Nice read I found with example code, didn't really know about this stuff before: https://codahale.com/a-lesson-in-timing-attacks/.

globau: yup; with python you'd normally use ``hmac.compare_digest``.

landoapi/decorators.py
```diff
@@ -0,0 +1,46 @@
...
+import logging
+
+from connexion import problem
+from flask import request
+from netaddr import IPAddress, IPNetwork
+
+logger = logging.getLogger(__name__)
+
+
+def limit_ips(
```
purelogiq: This decorator is defined but never used?

landoapi/app.py
```diff
@@ -55,6 +56,14 @@ def create_app(version_path):
     # Intialize the alembic extension
     alembic.init_app(app.app)
 
+    # Initialize rate limiter
+    # Most of the calls should come from other Conduit servers.
```
globau: i don't think it makes any sense to place rate limits on pingbacks from autoland-transplant.

landoapi/decorators.py
zalun: It shouldn't be commited (the file)

landoapi/decorators.py
```diff
@@ -0,0 +1,90 @@
...
+                        'msg': 'Wrong proxy. %s' % log_msg,
+                    }, log_tag
+                )
+                return problem(
+                    403,
+                    err_title,
+                    err_msg,
+                    type=
+                    'https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403'
+                )
```
globau: this doesn't make much sense.

the whole deal with configuring for proxies is they are used to indicate which ip addresses can be trusted in `x-forwarded-for`. note the bugzilla code i provided as a reference on irc is part of it's util function that returns the client's ip address.

remove this as well as the logic for iterating over each ip in the `x-forwarded-for` header.  instead write a custom `get_remote_address` function that uses the configured list of trusted proxies to determine the remote endpoint's address.  use that function any place you need to get the client's ip address.

    trusted_proxies = {'127.0.0.1'}
    route = request.access_route + [request.remote_addr]
    remote_addr = next((addr for addr in reversed(route) 
                        if addr not in trusted_proxies), request.remote_addr)

landoapi/decorators.py
```diff
@@ -0,0 +1,90 @@
...
+            in_whitelist = False
+            for ip_mask in whitelist_ip:
+                if IPAddress(ip) in IPNetwork(ip_mask):
+                    in_whitelist = True
+                    break
+            if not in_whitelist:
+                logger.warning(
+                    {
+                        'remote_addr': ip,
+                        'msg': 'Wrong IP address. %s' % log_msg,
```
globau: "wrong ip address" isn't a useful error message.

try something along the lines of "the ip 0.0.0.0 is not allowed to connect"

tests/test_landings.py
```diff
@@ -188,7 +190,25 @@ def test_update_landing_bad_request_id(db, client):
...
+        data=json.dumps({
+            'request_id': 1,
+            'landed': True,
+            'result': 'sha123'
+        }),
+        headers=[('API-Key', 'wrongapikey')],
+        content_type='application/json'
+    )
+
+    assert response.status_code == 403
```
globau: whitelisting needs a test.

landoapi/api/landings.py
```diff
@@ -118,6 +127,21 @@ def update(landing_id, data):
         revision (sha) of push if landed == true
         empty string if landed == false
     """
+    if not hmac.compare_digest(request.headers['API-Key'], TRANSPLANT_API_KEY):
```
globau: this will throw a `KeyError` exception if the `API-Key` header is missing.  treat a missing `API-Key` header the same as an incorrect api-key.

tests/test_landings.py
globau: need a test to ensure requests without an api-key are rejected

globau: still needs a test to ensure requests without an api-key are rejected

globau:
can you please provide links to the issues/cards tracking the whitelisting and rate limiting work

zalun:
Just created a card about it and "trellobug'd" it.
https://trello.com/c/1soDarfj/474-restrict-lando-api-pingback-endpoint-to-transplant-only
https://bugzilla.mozilla.org/show_bug.cgi?id=1385857

landoapi/models/landing.py
```diff
@@ -63,14 +63,18 @@ def create(cls, revision_id, phabricator_api_key=None, diff_id=None):
         # save landing to make sure we've got the callback
         landing = cls(revision_id=revision_id, diff_id=diff_id).save()
 
-        trans = TransplantClient()
-        callback = '%s/landings/%s/update' % (
-            os.getenv('HOST_URL'), landing.id
+        # Define the pingback URL with the port
+        callback = '{host_url}:{pingback_port}/landings/{id}/update'.format(
```
ckolos: host_url here will be something different than the public facing url. Since this is taken from the env, no code needs to be changed, but noting here so you are all aware.

smacleod:
@zalun You still have open review requests from glob and he had requested changes on this pull request. Please do not merge without getting proper final review, *especially* on security sensitive code.

landoapi/models/landing.py
globau: this issue should have blocked this from merging - the callback url as implemented won't work.  please ensure this is tracked and resolved before release.

globau:
i can't provide an proper r+ as this has already merged.  aside from the callback url issue this lgtm.

PR #25, opened 2017-07-12
Store Diff ID in Landing model (bug 1380283)

zalun:
Storing diff id and providing it in revision payload is enough info for lando-ui to decide if it should display the warning about an old diff


tests/test_landings.py
```diff
@@ -47,7 +45,7 @@ def test_landing_revision_saves_data_in_db(db, client, phabfactory):
 
     # test saved data
     landing = Landing.query.get(1)
-    assert landing.serialize() == CANNED_LANDING_1
+    assert landing.serialize() == CANNED_LANDING_FACTORY_1
```
mars-f: This is for a follow-up commit, but I don't understand what this test is verifying.  It saves two POSTs to the database, then checks the database contents.  If it's testing some sort of sequence generation across POSTs, then the test name and content should focus only on verifying that.  If it's testing POST data serialization to the database, then only that bit should be verified (and the second POST should be dropped).

purelogiq:
@mars-f We defiantly need to return and store diff ids, not phids. There is no problem querying diff information with a diff ID, as you can see here [https://mozphab.dev.mozaws.net/conduit/method/differential.querydiffs/](https://mozphab.dev.mozaws.net/conduit/method/differential.querydiffs/). In fact, that URL requires a normal diff ID, not a PHID

tests/test_landings.py
zalun: you're right - it was to check if request id increments in stubbed transplant service

migrations/363fc0301235_diff_phid_added.py
```diff
@@ -0,0 +1,30 @@
+"""Diff PHID added
```
purelogiq: Does this migration need to be changed to be diff_id?

tests/test_landings.py
```diff
@@ -145,7 +149,7 @@ def test_get_jobs(db, client):
 
 
 def test_update_landing(db, client):
-    Landing(1, 'D1', 'started').save()
+    Landing(1, 'D1', 'diff-phid-1', 'started').save()
```
purelogiq: phid?

tests/test_landings.py
```diff
@@ -179,7 +183,7 @@ def test_update_landing_bad_id(db, client):
 
 
 def test_update_landing_bad_request_id(db, client):
-    Landing(1, 'D1', 'started').save()
+    Landing(1, 'D1', 'diff-phid-1', 'started').save()
```
purelogiq: phid?

migrations/363fc0301235_diff_phid_added.py
zalun: I think I will simply remove the phid step, but if someone had it already in sqlite than it would raise an error...

PR #26, opened 2017-07-14
Use Sentry to report exceptions

mars-f:
Does what is says on the tin.  Sentry is shut off for local development.  To use it you have to give Sentry a real DSN.

I have tested this change using the lando-api Sentry project.  It reports errors correctly when they occur in a local development container.  I think we can ship this.

### DIFF
--- a/landoapi/app.py
+++ b/landoapi/app.py
@@ -1,6 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+import json
 import os
 
 import click
@@ -8,11 +9,15 @@ import connexion
 import logging
 
 from connexion.resolver import RestyResolver
+from raven.contrib.flask import Sentry
+
 from landoapi.dockerflow import dockerflow
 from landoapi.models.storage import alembic, db
 from mozlogging import MozLogFormatter
 
 logger = logging.getLogger(__name__)
+
+sentry = Sentry()
 
 
 def create_app(version_path):
@@ -24,8 +29,15 @@ def create_app(version_path):
 
     # Get the Flask app being wrapped by the Connexion app.
     flask_app = app.app
+
     flask_app.config['VERSION_PATH'] = version_path
     log_config_change('VERSION_PATH', version_path)
+
+    version_info = json.load(open(version_path))
+    logger.info(version_info, 'app.version')
+
+    this_app_version = version_info['version']
+    initialize_sentry(flask_app, this_app_version)
 
     db_uri = flask_app.config.setdefault(
         'SQLALCHEMY_DATABASE_URI', os.environ.get('DATABASE_URL', 'sqlite://')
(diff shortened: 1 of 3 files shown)


landoapi/app.py
```diff
@@ -1,19 +1,24 @@
...
 from connexion.resolver import RestyResolver
+from raven.contrib.flask import Sentry
+
 from landoapi.dockerflow import dockerflow
 from landoapi.models.storage import alembic, db
 from mozlogging import MozLogFormatter
 
 logger = logging.getLogger(__name__)
 
+sentry = Sentry()
```
darkwing: n00blyfe™ but why not initialize this in `initialize_sentry` since it's the only place it's used?

landoapi/app.py
```diff
@@ -24,9 +29,17 @@ def create_app(version_path):
...
     # Get the Flask app being wrapped by the Connexion app.
     flask_app = app.app
+
     flask_app.config['VERSION_PATH'] = version_path
     log_config_change('VERSION_PATH', version_path)
 
+    version_info = json.load(open(version_path))
+    logger.info(version_info, 'app.version')
+
+    this_app_version = version_info['commit']
```
purelogiq: Considering that there is a field specifically named 'version' I think we should use that instead of 'commit' even if they are equal right now, they could differ in the future.

landoapi/app.py
```diff
@@ -24,9 +29,17 @@ def create_app(version_path):
...
+    environment = version_info['build']
```
purelogiq: The version.json 'build' doesn't refer to the environment, the refers to the specific cirlce ci build that created this image (see https://circleci.com/docs/1.0/environment-variables/). With that in mind it cannot be used as the environment. 

We could create another environment variable ('ENVIRONMENT' or something) and have @ckolos  set it properly in dev, staging, and prod. Or we can switch based on the HOST_URL that is already available.

landoapi/app.py
```diff
@@ -46,6 +59,40 @@ def create_app(version_path):
...
+        flask_app: A Flask() instance.
+        release: A string representing this application release number (such as
+            a git sha).  Will be used as the Sentry "release" identifier. See
+            the Sentry client configuration docs for details.
+        environment: A string representing the application environment, such
+            as 'dev', 'stage', or 'prod'.  Will be used as the Sentry
+            "environment" identifier.  See the Sentry client configuration docs
+            for details.
+    """
+    sentry_dsn = os.environ.get('SENTRY_DSN', None)
```
purelogiq: The environment variable 'SENTRY_DSN' should be present in our docker-compose file even if it is empty or just a comment because our docker-compose defines all the environment variables needed to run the application. It is not used in production of course, it is just a good way to remind us what will be in the environment.

landoapi/app.py
purelogiq: It seems like `sentry` can be used in other files to manually handle exceptions, it is a global like `logger`.

purelogiq: It is a global, like logger, that can be used in other files to manually handle exceptions 😃

mars-f: Good point, will do.  I was copying the docs, but you are totally right. :)

purelogiq: Wait, it is not a bad thing that it is global. That will let us use it whereever we want.

landoapi/app.py
mars-f: Cool, I'll change it.

landoapi/app.py
mars-f: As I understand it, the environment is one way to partition the set of exceptions thrown by Sentry.  Hostname is presumably another?  But if you have multiple hosts per environment, like in prod, then partitioning the pool of exceptions coming from all the different environments by hostname might be a pain.

I like the `ENVIRONMENT` suggestion.  I'd also like to know if Ops already has a way they do this.  @ckolos, are you aware of any standard-ish way to get the environment label a host lives under into the container data so Sentry can report on it?

landoapi/app.py
mars-f: Nice catch, I'll add it.

landoapi/app.py
ckolos: "environment label a host lives under"... ?

You mean like "this is running in dev/stage/prod"? If so, that's available from the env as the ENV var.

mars-f: @ckolos perfect, thanks!

landoapi/app.py
mars-f: We can always extract it again if we need it.

PR #32, opened 2017-08-01
Remove mocked transplant service

zalun:
Transplant service is running via requests_mock.
This commit is removing this dependency so Lando API will be contacting
a real Transplant service.

### DIFF
--- a/tests/test_landings.py
+++ b/tests/test_landings.py
@@ -29,23 +29,38 @@ def db(app):
         _db.drop_all()
 
 
-def test_landing_revision_saves_data_in_db(db, client, phabfactory):
+def test_landing_revision_saves_data_in_db(
+    db, client, phabfactory, transfactory
+):
+    # Id of the landing in Autoland is created as a result of a POST request to
+    # /autoland endpoint. It is provided by Transplant API
+    land_request_id = 3
+    # Id of a Landing object is created as a result of a POST request to
+    # /landings endpoint of Lando API
+    landing_id = 1
+    # Id of the diff existing in Phabricator
+    diff_id = 2
+
     phabfactory.user()
     phabfactory.revision()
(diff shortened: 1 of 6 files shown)


tests/test_landings.py
```diff
@@ -29,9 +29,13 @@ def db(app):
         _db.drop_all()
 
 
-def test_landing_revision_saves_data_in_db(db, client, phabfactory):
+def test_landing_revision_saves_data_in_db(
+    db, client, phabfactory, transfactory
+):
     phabfactory.user()
     phabfactory.revision()
+    transfactory.land()
```
mars-f: I think this test is missing some data that would make the condition under test clearer.  The user is not important to what's being verified, but I think seeing that a revision with the same ID is being created and being landed is important to seeing what is verified.

Could we add a "revision" parameter to `transfactory` calls so that they aren't using invisible default data to do their work?  And change the language of method so that is implies creating a response rather than executing a process.  Something like this:

```python
phabfactory.user()
revision = phabfactory.revision()
transfactory.add_response(revision, 'pending')
```

zalun: transfactory does not depend on revision - it's just returning the request_id under which it saved the landing request, I might make the number to be set with this command. The status of the Landing request is set using pingback, which is tested in another test.

tests/test_landings.py
```diff
@@ -29,23 +29,38 @@ def db(app):
...
             'revision_id': 'D1',
-            'diff_id': 1
+            'diff_id': diff_id
         }),
         content_type='application/json'
     )
     assert response.status_code == 202
     assert response.content_type == 'application/json'
-    assert response.json == {'id': 1}
+    # id of the Landing object in Lando API
```
mars-f: Typo: Comments on lines above start with "# Id", this line starts with "# id".

PR #33, opened 2017-08-01
Add diff and commit author information in GET /revisions/{id}

purelogiq:
Clients sometimes need to know about the most recent diff for a revision. This change returns some metadata about the diff as well as the name and email address of the commit authors.

For example, lando-ui will use this information to allow the user to preview the final commit message allowing the user to make any final edits ahead of time.

https://trello.com/c/V3EMcEQC/445-2-return-commit-authors-in-revision-data

### DIFF
--- a/landoapi/api/revisions.py
+++ b/landoapi/api/revisions.py
@@ -27,11 +27,18 @@ def get(revision_id, api_key=None):
             type='https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404'
         )
 
-    return _format_revision(phab, revision, include_parents=True), 200
+    return _format_revision(
+        phab, revision, include_diff=True, include_parents=True
+    ), 200
 
 
 def _format_revision(
-    phab, revision, include_parents=False, last_author=None, last_repo=None
+    phab,
+    revision,
+    include_diff=False,
+    include_parents=False,
+    last_author=None,
+    last_repo=None
 ):
     """ Formats a revision given by Phabricator to match Lando's spec.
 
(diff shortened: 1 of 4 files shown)


landoapi/api/revisions.py
```diff
@@ -53,6 +62,29 @@ def _format_revision(
...
+        diff = {
+            'id': int(raw_diff['id']),
+            'revision_id': int(raw_diff['revisionID']),
+            'date_created': int(raw_diff['dateCreated']),
+            'date_modified': int(raw_diff['dateModified']),
+            'vcs_base_revision': raw_diff['sourceControlBaseRevision'],
+            'authors': None
+        }
+
+        if raw_diff['properties'] and raw_diff['properties']['local:commits']:
```
mars-f: These two dict access attempts appear fragile: they will blow up if the phabricator API's JSON schema changes, or if the 'local:commits' property is not guaranteed to be present in all responses.  Is that by design?  If that's not the intention, it might be better to use `dict.get()`.

landoapi/api/revisions.py
```diff
@@ -53,6 +62,29 @@ def _format_revision(
     Returns:
         A hash of the formatted revision information.
     """
+    # Load the revision's diff if requested.
```
mars-f: This function is getting pretty large, and collecting distinct blocks of functionality that have a single comment describing them - those are signs it might be time to break this function up into a series of well-named functions.

(something I quickly scratched out, could probably be made even cleaner)
```python
if include_diff:
    response.update(build_response_diff(...))
if include_parents:
    response.update(build_response_parents(...))
if last_author:
    response.update(build_response_last_author(...))
...
```

landoapi/api/revisions.py
purelogiq: This is by design, 'properties' is always present but can be an empty dict or None, and so for 'local:commits'.

mars-f:
:+1:

PR #37, opened 2017-08-14
Saving patch in S3

zalun:
Currently patch's content was sent as a parameter of a POST request to
Transplant's `autoland` endpoint. Since patch can and often is a large
block of text, it has been decided to save it in S3 bucket.

The name of the bucket is to be saved in environent variable.
One might set the access and secret keys in AWS_ACCESS_KEY and
AWS_SECRET_KEY environment variables.


landoapi/models/landing.py
```diff
@@ -128,3 +132,37 @@ class RevisionNotFoundException(Exception):
...
+        revision_id: String ID of the revision (ex. 'D123')
+        diff_id: The integer ID of the raw diff
+
+    Returns
+        String representing the patch's URL in S3
+        (ex. 's3://{bucket_name}/D123_1.patch')
+    """
+    s3 = boto3.resource(
+        's3',
+        aws_access_key_id=os.getenv('AWS_ACCESS_KEY', None),
```
globau: as per our irc discussion and what was discussed in the meeting, credentials don't need to be present on the server - instead the ec2 policies should be used to allow access to the s3 bucket.

zalun: Yes, if the environment variables aren't set, access and secret keys are set to None.
That will make it work in the way we discussed. I can add the comment that this variables are not set in production. Using `flask-boto3` would move this code to `app.py`. No other benefits.

landoapi/models/landing.py
```diff
@@ -128,3 +132,38 @@ class RevisionNotFoundException(Exception):
...
+    patch_name = '{revision_id}_{diff_id}.patch'.format(
+        revision_id=revision_id, diff_id=diff_id
+    )
+    patch_url = 's3://{bucket}/{patch_name}'.format(
+        bucket=bucket, patch_name=patch_name
+    )
+    with tempfile.TemporaryFile() as patchfile:
+        patchfile.write(patch.encode('utf-8'))
+        patchfile.seek(0)
+        s3.meta.client.upload_fileobj(patchfile, bucket, patch_name)
```
mars-f: Uploading this file appears to be a significant action - does it need an INFO level log message?

landoapi/transplant_client.py
```diff
@@ -20,23 +15,18 @@ class TransplantClient:
     def __init__(self):
         self.api_url = os.getenv('TRANSPLANT_URL')
 
-    def land(self, ldap_username, hgpatch, tree, pingback):
+    def land(self, ldap_username, patch_url, tree, pingback):
         """ Sends a push request to Transplant API to land a revision.
 
         Returns request_id received from Transplant API.
```
mars-f: This docstring needs updating.

tests/conftest.py
```diff
@@ -83,3 +90,26 @@ def app(versionfile, docker_env_vars, disable_migrations, disable_log_output):
...
+        _db.init_app(app)
+        _db.create_all()
+        # we just created.
+        yield _db
+        _db.session.remove()
+        _db.drop_all()
+
+
+@pytest.fixture
+def s3():
```
mars-f: I think this fixture also needs to depend on the `docker_env_vars` fixture because it uses the `PATCH_BUCKET_NAME` env var.

landoapi/models/landing.py
zalun: Definitely

PR #46, opened 2018-09-17
docker: unify development and production images.

smacleod:
Using a single docker image for both local development and
production deployment makes it less likely that we'll break something in
production and not know it until build/deploy time. It is now much
easier to test production like running of the container locally since
the full uwsgi setup is included.

As part of this work the circle-ci configuration has been simplified to
build and run tests more similarily to local development. While it would
be nice to split build / test / deploy into their own jobs again, we
don't have access to the circle-ci docker layer caching at this point.
We will now only build the image a single time for a CI run, rather than
building it in each step.


dylanwh:
This looks fine -- the only reasons I avoided this before were I wasn't entirely sure of the differences between dev/prod and the old setup was definitely doing something different for running the tests.

Be aware that every time I've tried to use volumes in circleci Weird Shit has happened, as I think they're running a custom volume driver.

PR #56, opened 2019-03-27
WIP api: report secure revisions (bug 1539289)

mars-f:
Return a list of secure revisions from API calls that produce a TransplantAssessment.


mars-f:
Superceded by #61

PR #61, opened 2019-04-23
[sec-approval/1] api: report secure revisions (bug 1539289)

mars-f:
Revision entities returned by the API publish a new "is_secure"
property.  Revisions for which this property is True contain
security-sensitive information and should be handled appropriately by
clients.

### DIFF
--- a/landoapi/api/stacks.py
+++ b/landoapi/api/stacks.py
@@ -81,6 +81,8 @@ def get(revision_id):
     users = user_search(phab, involved_phids)
     projects = project_search(phab, involved_phids)
 
+    secure_project_phid = get_secure_project_phid(phab)
+
     revisions_response = []
     for phid, revision in stack_data.revisions.items():
         revision_phid = PhabricatorClient.expect(revision, "phid")
@@ -104,6 +106,7 @@ def get(revision_id):
             title, bug_id, accepted_reviewers, summary, revision_url
         )
         author_response = serialize_author(phab.expect(fields, "authorPHID"), users)
+
         revisions_response.append(
             {
                 "id": human_revision_id,
@@ -126,9 +129,7 @@ def get(revision_id):
                 "diff": serialize_diff(diff),
                 "author": author_response,
                 "reviewers": serialize_reviewers(reviewers, users, projects, diff_phid),
-                "is_secure": revision_is_secure(
-                    revision, get_secure_project_phid(phab)
-                ),
+                "is_secure": revision_is_secure(revision, secure_project_phid),
             }
         )
 
--- a/landoapi/api/stacks.py
+++ b/landoapi/api/stacks.py
@@ -10,7 +10,7 @@ from flask import current_app, g
 from landoapi.commit_message import format_commit_message
 from landoapi.decorators import require_phabricator_api_key
 from landoapi.phabricator import PhabricatorClient, ReviewerStatus
-from landoapi.projects import project_search
+from landoapi.projects import project_search, get_secure_project_phid
 from landoapi.repos import get_repos_for_env
 from landoapi.reviews import (
     get_collated_reviewers,
@@ -104,7 +104,6 @@ def get(revision_id):
             title, bug_id, accepted_reviewers, summary, revision_url
         )
         author_response = serialize_author(phab.expect(fields, "authorPHID"), users)
-
         revisions_response.append(
             {
                 "id": human_revision_id,
@@ -127,7 +126,9 @@ def get(revision_id):
                 "diff": serialize_diff(diff),
                 "author": author_response,
                 "reviewers": serialize_reviewers(reviewers, users, projects, diff_phid),
-                "is_secure": revision_is_secure(revision, phab),
+                "is_secure": revision_is_secure(
+                    revision, get_secure_project_phid(phab)
+                ),
             }
         )
 
--- a/landoapi/api/stacks.py
+++ b/landoapi/api/stacks.py
@@ -10,7 +10,7 @@ from flask import current_app, g
 from landoapi.commit_message import format_commit_message
 from landoapi.decorators import require_phabricator_api_key
 from landoapi.phabricator import PhabricatorClient, ReviewerStatus
-from landoapi.projects import project_search, get_secure_project_phid
+from landoapi.projects import get_secure_project_phid, project_search
 from landoapi.repos import get_repos_for_env
 from landoapi.reviews import (
     get_collated_reviewers,
@@ -20,10 +20,10 @@ from landoapi.reviews import (
 from landoapi.revisions import (
     gather_involved_phids,
     get_bugzilla_bug,
+    revision_is_secure,
     serialize_author,
     serialize_diff,
     serialize_status,
-    revision_is_secure,
 )
 from landoapi.stacks import (
     build_stack_graph,
--- a/landoapi/api/stacks.py
+++ b/landoapi/api/stacks.py
@@ -23,6 +23,7 @@ from landoapi.revisions import (
     serialize_author,
     serialize_diff,
     serialize_status,
+    revision_is_secure,
 )
 from landoapi.stacks import (
     build_stack_graph,
@@ -126,6 +127,7 @@ def get(revision_id):
                 "diff": serialize_diff(diff),
                 "author": author_response,
                 "reviewers": serialize_reviewers(reviewers, users, projects, diff_phid),
+                "is_secure": revision_is_secure(revision, phab),
             }
         )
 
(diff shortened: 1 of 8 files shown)


landoapi/api/transplants.py
```diff
@@ -5,7 +5,7 @@
 import urllib.parse
 
 import kombu
-from connexion import problem, ProblemException
+from connexion import ProblemException, problem
```
smacleod: Convention throughout lando-api is to ignore case when alphabetizing, please revert the changes to imports here and elsewhere.

landoapi/api/transplants.py
```diff
@@ -171,7 +171,9 @@ def _assess_transplant_request(phab, landing_path):
     reviewers = {
         revision["phid"]: get_collated_reviewers(revision) for revision, _ in to_land
     }
+    is_secure = revision_is_secure(revision, phab)
 
+    # fmt: off
```
smacleod: Why? This code seems pretty basic, I don't see why it warrants special formatting?

landoapi/revisions.py
```diff
@@ -118,3 +119,26 @@ def check_author_planned_changes(*, revision, **kwargs):
...
+
+    Such revisions should be handled according to the Security Bug Approval Process.
+    See https://wiki.mozilla.org/Security/Bug_Approval_Process.
+
+    Args:
+        revision: The 'data' element from a Phabricator API response containing
+            revision data.
+        phabclient: A PhabricatorClient instance.
+    """
+    secure_project_phid = get_secure_project_phid(phabclient)
```
smacleod: Every time this is called we go across the network to the redis cache at minimum, I'd much rather we leave things where the phid is grabbed once as part of the request and then passed around. Making this take the `phid` rather than a full phabclient means it's a pure function with only basic types going in and out as well.

Particularly I like to avoid sending phabclients down into functions unless there's a ton of value, to avoid the creep we had before where everything was layered super deep and calls were being made to phabricator all over the place. The general way things are laid out now is to call to external systems in the top level request most of the time and then pass just basic python types around.

The help functions on phabclient, such as `expect` are static methods, which again are pure functions, so you can make use of them directly off of the class `PhabricatorClient.expect(...)` without needing the instantiated client.

landoapi/api/transplants.py
```diff
@@ -180,8 +182,10 @@ def _assess_transplant_request(phab, landing_path):
         reviewers,
         users,
         projects,
-        get_secure_project_phid(phab),
+        is_secure,
```
smacleod: You've changed the semantics here. Previously we were checking the security of each revision, but now this would check only a single revision and apply it's security status to all the revisions.

Please revert things to just pass the `phid` and let the warning system check each revision. Having the new `revision_is_secure()` function will still simplify the warning body a lot.

landoapi/revisions.py
```diff
@@ -118,3 +119,26 @@ def check_author_planned_changes(*, revision, **kwargs):
...
+
+def revision_is_secure(revision, phabclient):
+    """Does the given revision contain security-sensitive data?
+
+    Such revisions should be handled according to the Security Bug Approval Process.
+    See https://wiki.mozilla.org/Security/Bug_Approval_Process.
+
+    Args:
+        revision: The 'data' element from a Phabricator API response containing
+            revision data.
```
smacleod: This isn't completely correct, it specifically needs to be a `differential.revision.search` response for a revision, which included the `projects` attachment. We have a convention for pulling all the data for the stack in the one place, so that it has all we need and is standard, you could refer to the stack data stuff if you want this comment to require less updating.

tests/test_revisions.py
```diff
@@ -65,3 +69,46 @@ def test_check_author_planned_changes_changes_planned(phabdouble):
         attachments={"reviewers": True, "reviewers-extra": True, "projects": True},
     )["data"][0]
     assert check_author_planned_changes(revision=revision) is not None
+
+
+def test_secure_api_flag_on_public_revision_is_false(client, phabdouble):
```
smacleod: You should probably update one of the integrated stack tests to make something secure, as that would have caught the issue around assuming security based on a single revision's status.

landoapi/api/transplants.py
mars-f: This change was reverted with the old warning behaviour.

landoapi/revisions.py
mars-f: Changed.

landoapi/api/transplants.py
mars-f: I added an integration test to catch the bug and reverted the call chain to the old behaviour.

landoapi/revisions.py
mars-f: I used the docstring from another function with a "revision" data argument as a template to clarify this docstring.

tests/test_revisions.py
mars-f: Agreed.  I added an integrated transplant test to ensure the warnings are computed correctly and an integrated stack test to make sure the stack values are correct.

landoapi/api/stacks.py
```diff
@@ -103,7 +104,6 @@ def get(revision_id):
             title, bug_id, accepted_reviewers, summary, revision_url
         )
         author_response = serialize_author(phab.expect(fields, "authorPHID"), users)
-
```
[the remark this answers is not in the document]
mars-f: No idea what that's about.  Reverted.

landoapi/api/stacks.py
```diff
@@ -126,6 +126,9 @@ def get(revision_id):
                 "diff": serialize_diff(diff),
                 "author": author_response,
                 "reviewers": serialize_reviewers(reviewers, users, projects, diff_phid),
+                "is_secure": revision_is_secure(
+                    revision, get_secure_project_phid(phab)
```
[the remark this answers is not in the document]
mars-f: Fixed.

landoapi/revisions.py
```diff
@@ -118,3 +118,26 @@ def check_author_planned_changes(*, revision, **kwargs):
...
+        secure_project_phid: The PHID of the Phabricator project used to tag
+            secure revisions.
+    """
+    revision_project_tags = PhabricatorClient.expect(
+        revision, "attachments", "projects", "projectPHIDs"
+    )
+    is_secure = secure_project_phid in revision_project_tags
+    logger.debug(
+        "revision is security-sensitive?",
+        extra={"value": is_secure, "revision": revision["id"]},
```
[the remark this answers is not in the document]
mars-f: Agreed, removed.

landoapi/transplants.py
```diff
@@ -242,17 +246,14 @@ def warning_revision_secure(*, revision, secure_project_phid, **kwargs):
     if secure_project_phid is None:
         return None
 
-    revision_project_tags = PhabricatorClient.expect(
-        revision, "attachments", "projects", "projectPHIDs"
-    )
-    if secure_project_phid not in revision_project_tags:
+    if revision_is_secure(revision, secure_project_phid):
```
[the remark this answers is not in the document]
mars-f: Done.

tests/test_revisions.py
```diff
@@ -65,3 +69,47 @@ def test_check_author_planned_changes_changes_planned(phabdouble):
...
+    revision = phabdouble.revision(projects=[secure_project])
+
+    response = client.get("/stacks/D{}".format(revision["id"]))
+
+    assert response.status_code == 200
+    response_revision = response.json["revisions"].pop()
+    assert response_revision["is_secure"]
+
+
+def test_public_revision_is_not_secure(app, phabdouble, secure_project):
```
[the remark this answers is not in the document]
mars-f: It's leftover from an earlier version of this test.  Removed.

tests/test_revisions.py
```diff
@@ -65,3 +69,47 @@ def test_check_author_planned_changes_changes_planned(phabdouble):
...
+    r = phabdouble.revision(projects=[public_project])
+    revision = phab.call_conduit(
+        "differential.revision.search",
+        constraints={"phids": [r["phid"]]},
+        attachments={"reviewers": True, "reviewers-extra": True, "projects": True},
+    )["data"].pop()
+    assert not revision_is_secure(revision, secure_project["phid"])
+
+
+def test_secure_revision_is_secure(app, phabdouble, secure_project):
```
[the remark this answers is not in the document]
mars-f: It's leftover from an earlier version of this test.  Removed.

tests/test_transplants.py
```diff
@@ -142,6 +142,46 @@ def test_dryrun_reviewers_warns(client, db, phabdouble, auth0_mock):
     assert response.json["confirmation_token"] is not None
 
 
+def test_dryrun_secure_revision_in_stack_warns(
```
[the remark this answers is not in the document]
mars-f: I agree that this test can be removed, I think this test is narrow and unlikely to catch future errors.  An earlier version of this PR broke how warnings were calculated, however no tests failed.  The test was written to expose the bug and verify the fix.

PR #64, opened 2019-05-02
app: only wrap API exceptions in production environments

mars-f:
The default error handlers eat the PhabricatorAPIException root cause
and traceback. This makes debugging API failures difficult and breaks
PyCharm unittest integration.

We can make the error handler optional in the same way that the Flask
default error handler is optional: in production it wraps all errors,
but in the test and debug environments exceptions are allowed to
propagate through the stack to make debugging easier.

### DIFF
--- a/landoapi/hooks.py
+++ b/landoapi/hooks.py
@@ -85,6 +85,8 @@ def initialize_hooks(flask_app):
     flask_app.before_request(request_logging_before_request)
     flask_app.after_request(request_logging_after_request)
 
-    flask_app.register_error_handler(
-        PhabricatorAPIException, handle_phabricator_api_exception
-    )
+    # Only wrap/mask exceptions if we are in a production-like environment.
+    if not flask_app.propagate_exceptions:
+        flask_app.register_error_handler(
+            PhabricatorAPIException, handle_phabricator_api_exception
+        )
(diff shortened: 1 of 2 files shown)


landoapi/hooks.py
```diff
@@ -85,6 +85,8 @@ def initialize_hooks(flask_app):
     flask_app.before_request(request_logging_before_request)
     flask_app.after_request(request_logging_after_request)
 
-    flask_app.register_error_handler(
-        PhabricatorAPIException, handle_phabricator_api_exception
-    )
+    # Only wrap/mask exceptions if we are in a production-like environment.
+    if not flask_app.propagate_exceptions:
```
smacleod: It feels strange to me not to register the handler when we're running in development/testing. What about adding the check to the handler itself like flask does for its exceptions: https://github.com/pallets/flask/blob/1dda032b00cb4f7a783e018a7c370ee88f15e797/flask/app.py#L1807 

Would make the unit testing setup simpler since you could just flip the propgate_exceptions config and not have to re-initialize everything

mars-f: Yeah it's better to switch.  The FlaskApp code is complicated in a way I didn't want to reproduce, but on second look it turns out it is 95% Python 2 compat code.  The Python 3 re-raise code is trivial and my testing shows it correctly re-raises the original exception site with full traceback.

PR #65, opened 2019-05-07
transplants: remove #check-in_needed when landing (Bug 1535157).

smacleod:
Because Check-in Needed is being migrated to Phabricator and will be
handled by Lando, we now remove the tag from revisions when landing.
This is done asynchronously after triggering the landing so that any
errors will not block a proper response. It could have been done before
triggering the landing but any transient failure talking to transplant
would have prevented the landing even though the tag was removed.

Additionally Lando now has a `PHABRICATOR_ADMIN_API_KEY` environment
variable which is used to specify an API key for privileged actions.
This key must have access to all revisions, including those marked as
secure. Care should be taken when using this API key.

Removal of the #check-in_needed tag uses the admin api key to ensure
that it is successful for all revisions and that it is clear Lando is
the bot user causing the removal.

This change was tested by updating the automated tests, and running the
removal code manually against phabricator-dev.allizom.org.

### DIFF
--- a/landoapi/phabricator.py
+++ b/landoapi/phabricator.py
@@ -14,6 +14,8 @@ from landoapi.systems import Subsystem
 
 logger = logging.getLogger(__name__)
 
+PHAB_API_KEY_RE = re.compile(r"^api-.{28}$")
+
 
 @unique
 class RevisionStatus(Enum):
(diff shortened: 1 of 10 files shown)


landoapi/phabricator.py
```diff
@@ -348,19 +350,21 @@ class PhabricatorSubsystem(Subsystem):
...
-            and re.match(
-                r"^api-.{28}$",
-                self.flask_app.config["PHABRICATOR_UNPRIVILEGED_API_KEY"],
-            )
-            is None
-        ):
+        unpriv_key = self.flask_app.config["PHABRICATOR_UNPRIVILEGED_API_KEY"]
+        priv_key = self.flask_app.config["PHABRICATOR_ADMIN_API_KEY"]
+
+        if unpriv_key and PHAB_API_KEY_RE.match(unpriv_key) is None:
```
globau: you probably want `re.search()` not `re.match()` - your regex has a start of string anchor making `match` redundant, and imho `search` is clearer

tests/mocks.py
```diff
@@ -748,6 +746,137 @@ def to_response(i):
...
+            if "title" not in transaction_types or "update" not in transaction_types:
+                error_info = "Validation errors:"
+                if "title" not in transaction_types:
+                    error_info = error_info + "\n  - Revisions must have a title."
+                if "update" not in transaction_types:
+                    error_info = (
+                        error_info
+                        + "\n  - "
+                        + "You must specify an initial diff when creating a revision."
+                    )
```
globau: elsewhere you let the line run long and used `# noqa`.

the approach to long lines should be consistent - my vote is to concatenate instead of using `# noqa`.

tests/mocks.py
```diff
@@ -748,6 +746,137 @@ def to_response(i):
...
+            for r in self._revisions:
+                if r["phid"] == i or r["id"] == i or "D{}".format(r["id"]) == i:
+                    return r
+            return None
+
+        revision = identifier_to_revision(objectIdentifier)
+        if objectIdentifier is not None and revision is None:
+            error_info = (
+                f'Monogram "{objectIdentifier}" does not identify a valid object.'
+            )  # noqa
```
globau: i suspect `noqa` isn't needed here

PR #67, opened 2019-05-07
logging: log conduit API calls

mars-f:
Add DEBUG logging to conduit API calls to aid local development.

### DIFF
--- a/landoapi/phabricator.py
+++ b/landoapi/phabricator.py
@@ -198,6 +198,14 @@ class PhabricatorClient:
 
         data = {"output": "json", "params": json.dumps(kwargs)}
 
+        extra_data = {
+            "params": kwargs.copy(),
+            "method": method,
+            "api_url": self.api_url,
+        }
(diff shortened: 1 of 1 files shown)


landoapi/phabricator.py
```diff
@@ -198,6 +198,14 @@ def call_conduit(self, method, **kwargs):
 
         data = {"output": "json", "params": json.dumps(kwargs)}
 
+        extra_data = {
+            "params": kwargs.copy(),
+            "method": method,
+            "api_url": self.api_url,
+        }
+        del extra_data["params"]["__conduit__"]  # Sanitize the api token.
+        logger.debug("call to conduit", extra=extra_data)
```
smacleod: If you want to keep all the logging call objects temporary in a single call you could go with something like:
```
logger.debug(
    "call to conduit",
    extra={
        "params": {k: kwargs[k] for k in kwargs.keys() - {"__conduit__"}},
        "method": method,
        "api_url": self.api_url,
    }
)
```

PR #69, opened 2019-05-10
[sec-approval/2] api: accept sanitized commit messages (bug 1548610)

mars-f:
Add an API operation that accepts sanitized commit messages for
revisions.  Executing the operation kicks off the sec-approval workflow.


.flake8
```diff
@@ -11,3 +11,5 @@ exclude =
     docker,
     # Ignore test data, which is annoying to properly format.
     tests/canned_responses/,
+per-file-ignores =
+    landoapi/secapproval.py:B950,W293
```
smacleod: It'd be a real shame to turn these off for the whole file, especially `W293`. See my other comment below, I'd prefer we work around it a different way.

landoapi/projects.py
```diff
@@ -32,35 +37,53 @@ def project_search(phabricator, project_phids):
     return result_list_to_phid_dict(phabricator.expect(projects, "data"))
 
 
-@cache.cached(key_prefix=SEC_PROJ_CACHE_KEY, timeout=SEC_PROJ_CACHE_TIMEOUT)
-def get_secure_project_phid(phabricator):
-    """Return a phid for the project indicating revision security.
+def get_project_phid(project_slug, phabricator, allow_empty_result=True):
+    """Looks up a project's PHID.
+
+    The result is cached so further lookups don't hit the network.
```
smacleod: I think you missed updating some things.

landoapi/secapproval.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Args:
+        revision_phid: The PHID of the revision to edit.
+        message: The sanitized commit message string we want to be reviewed.
+        phabclient: A PhabClient instance.
+    """
+    comment = format_sanitized_message_comment_for_review(message)
+    sec_approval_phid = get_sec_approval_project_phid(phabclient)
+    phabclient.call_conduit(
+        "differential.revision.edit",
```
smacleod: This should return a list of transactions that you applied back to you. I thought the plan was to save the phid in the database, to make sure we're using the correct one? It'd also make the searching for the comment much quicker in your followup PRs.

landoapi/api/revisions.py
```diff
@@ -0,0 +1,73 @@
...
+from landoapi import auth
+from landoapi.decorators import require_phabricator_api_key
+from landoapi.projects import get_secure_project_phid
+from landoapi.revisions import revision_is_secure
+from landoapi.secapproval import send_sanitized_commit_message_for_review
+
+logger = logging.getLogger(__name__)
+
+
+@auth.require_auth0(())
```
smacleod: I feel like we should at least be requiring the `lando` scope.

landoapi/api/revisions.py
```diff
@@ -0,0 +1,73 @@
...
+        return problem(
+            400,
+            "Operation only allowed for secure revisions",
+            "Only security-sensitive revisions can be given sanitized commit messages",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
+
+    send_sanitized_commit_message_for_review(revision["phid"], alt_message, phab)
+
+    return "OK", 200
```
smacleod: It'd be cool if we could return some information that would allow the UI to either link to, or redirect to, the actual comment on Phabricator that resulted.

landoapi/spec/swagger.yml
```diff
@@ -186,6 +186,41 @@ paths:
           schema:
             allOf:
               - $ref: '#/definitions/Error'
+  /submitSanitizedCommitMessage:
```
smacleod: This also kicks off the actual request for `sec-approval`. I wonder if something like `/requestSecApproval` would be more appropriate (also, shorter, which would be nice :D)

landoapi/projects.py
mars-f: I kept the refactoring isolated to this module to keep from complicating the larger PR.  If the refactoring is good I can apply it in other places as a follow-up PR.

.flake8
mars-f: Done.

landoapi/api/revisions.py
mars-f: I don't think that's possible with the current approach.  Phab gives us back 2 transaction IDs when we add a custom commit message and we don't know which one is the comment.

landoapi/secapproval.py
mars-f: Changed.

landoapi/secapproval.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Args:
+        message: The commit message to be reviewed.
+    """
+    # The message is written in first-person form because it is being authored by the
+    # user in Lando and posted under their username.
+    #
+    # The message is formatted as Remarkup.
+    # See https://phabricator.services.mozilla.com/book/phabricator/article/remarkup/
+    return dedent(
```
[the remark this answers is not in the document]
mars-f: Changed.

landoapi/secapproval.py
```diff
@@ -0,0 +1,83 @@
...
+        f"""
+        I have written a sanitized comment message for this revision. It should
+        follow the [Security Bug Approval Guidelines](https://wiki.mozilla.org/Security/Bug_Approval_Process).
+        
+        ````
+        {message}
+        ````
+        
+        Could a member of the `sec-approval` team please review this message?
```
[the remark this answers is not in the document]
mars-f: Changed.

landoapi/api/revisions.py
```diff
@@ -0,0 +1,86 @@
...
+        sanitized_message: The sanitized commit message.
+    """
+    phab = g.phabricator
+
+    revision_phid = data["revision_phid"]
+    alt_message = data["sanitized_message"]
+
+    logger.debug(
+        "Got request to update revision with a sanitized message",
+        extra=dict(revision_phid=revision_phid, sanitized_message=alt_message),
```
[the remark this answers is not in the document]
mars-f: I can remove the message itself from the extra data and only record the fact that a request for sec-approval review was made.  I can then bump the log level up to INFO (which is a stand-in for AUDIT).

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,47 @@
...
+See See https://wiki.mozilla.org/Security/Bug_Approval_Process.
+"""
+
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.orm import relationship
+
+from landoapi.storage import db
+
+
+class SecApprovalRevision(db.Model):
```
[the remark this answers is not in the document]
mars-f: I thought we denormalized just the transaction data table, not the revisions table.  I normalized it  because we have multiple request events per revision.  However the request events table should be small, I don't see a huge performance hit for pulling the revision PHID back into the event table as a column.

landoapi/api/revisions.py
```diff
@@ -0,0 +1,86 @@
...
+    save_sec_approval_request_event,
+    send_sanitized_commit_message_for_review,
+)
+
+logger = logging.getLogger(__name__)
+
+
+@auth.require_auth0(scopes=("lando",))
+@require_phabricator_api_key(optional=False)
+def submit_sanitized_commit_message(data=None):
```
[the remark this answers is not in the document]
mars-f: Changed.

landoapi/spec/swagger.yml
```diff
@@ -186,6 +186,41 @@ paths:
...
+          description: The data we want to submit
+          required: true
+          in: body
+          schema:
+            type: object
+            required:
+              - revision_phid
+              - sanitized_message
+            properties:
+              revision_phid:
```
smacleod: We accept revision IDs everywhere else, converting to PHIDs as needed behind the scenes. Probably best to keep this consistent. Sorry I didn't catch this on previous passes.

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,32 @@
...
+
+See See https://wiki.mozilla.org/Security/Bug_Approval_Process.
+"""
+
+from sqlalchemy.dialects.postgresql import JSONB
+
+from landoapi.storage import db
+
+
+class SecApprovalRequestEvent(db.Model):
```
smacleod: "RequestEvent" is jumping out as a little redundant to me, `SecApprovalRequest` would be just as descriptive, no?

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,32 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Models related to the sec-approval process.
+
+See See https://wiki.mozilla.org/Security/Bug_Approval_Process.
+"""
+
+from sqlalchemy.dialects.postgresql import JSONB
```
smacleod:
```suggestion
from sqlalchemy.dialects.postgresql.json import JSONB
```
just to stay consistent with our other imports of this.

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,32 @@
...
+
+class SecApprovalRequestEvent(db.Model):
+    """Represents an event that added a sec-approval comment to a revision."""
+
+    __tablename__ = "secapproval_request_events"
+
+    id = db.Column(db.Integer, primary_key=True)
+
+    # The revision PHID that this event applies to.
+    revision_phid = db.Column(db.String(128), nullable=False)
```
smacleod: Just as with the API endpoints, we are already storing things in the DB as the id, rather than phid. Probably best to keep it all the same.

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # The active diff PHID when sec-approval was requested.
+    diff_phid = db.Column(db.String(128), nullable=False)
```
smacleod: Same here, id vs phid.

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # Timestamp used to find the latest event in the series.
+    created_at = db.Column(
+        db.DateTime(timezone=True), nullable=False, default=db.func.now()
+    )
+
+    comment_candidates = db.Column(JSONB, nullable=False)
```
smacleod: This should probably have a decent comment explaining what structure is expected. See our use in `Transplant` https://github.com/mozilla-conduit/lando-api/blob/8b323be8ac9a8b0d62124b2ec0202c6cf5fe11c4/landoapi/models/transplant.py#L53-L72

landoapi/secapproval.py
```diff
@@ -0,0 +1,100 @@
...
+        phid = PhabricatorClient.expect(transaction, "phid")
+        possible_comment_phids.append(phid)
+
+    event = SecApprovalRequestEvent(
+        revision_phid=revision_phid,
+        diff_phid=diff_phid,
+        comment_candidates=possible_comment_phids,
+    )
+    db.session.add(event)
+    db.session.commit()
```
smacleod: It's best to keep use of the db session separate from the code creating and manipulating database objects. There were some race conditions in the past that required a bunch of refactoring to fix because of session use similar to this. In general, maybe a class method where you pass the information and it constructs the object would be better, and then the session management can happen right in the actual api endpoint function. For a decent explanation of the benefits of structuring things this way take a look at https://docs.sqlalchemy.org/en/13/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it

landoapi/models/secapproval.py
mars-f: I'm going to keep this one as-is.  The `differential.revision.search` API only gives us the diff PHID.  It would take a second API call to retrieve the diff ID.

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,61 @@
...
+
+    __tablename__ = "secapproval_requests"
+
+    id = db.Column(db.Integer, primary_key=True)
+
+    # The revision ID that this event applies to.
+    revision_id = db.Column(db.Integer, nullable=False)
+
+    # The active diff PHID when sec-approval was requested.
+    diff_phid = db.Column(db.String(128), nullable=False)
```
smacleod: Where are you getting the `128` from? From what I can tell Phabricator stores all phids in `VARBINARY(64)` mysql columns, meaning at most 64 characters in any character encoding that would be sanely used.

If we're not too concerned about throwing errors if some limit is passed, we should probably just use `TEXT` here. Take a read of https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_varchar.28n.29_by_default

and from https://www.postgresql.org/docs/9.1/datatype-character.html

> Tip: There is no performance difference among these three types, apart from increased storage space when using the blank-padded type, and a few extra CPU cycles to check the length when storing into a length-constrained column. While character(n) has performance advantages in some other database systems, there is no such advantage in PostgreSQL; in fact character(n) is usually the slowest of the three because of its additional storage costs. In most situations text or character varying should be used instead.

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,61 @@
...
+
+class SecApprovalRequest(db.Model):
+    """Represents an event that added a sec-approval comment to a revision."""
+
+    __tablename__ = "secapproval_requests"
+
+    id = db.Column(db.Integer, primary_key=True)
+
+    # The revision ID that this event applies to.
+    revision_id = db.Column(db.Integer, nullable=False)
```
smacleod: Probably worth sticking an index on this column since it'll be queried to find specific rows. a `HASH` index is probably the most efficient here, but the default index type would also be fine.

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # The active diff PHID when sec-approval was requested.
+    diff_phid = db.Column(db.String(128), nullable=False)
+
+    # Timestamp used to find the latest event in the series.
+    created_at = db.Column(
+        db.DateTime(timezone=True), nullable=False, default=db.func.now()
+    )
+
+    # A JSON list of transaction PHIDs that may be sec-approval request comments.
```
smacleod: pedantic nit:
```suggestion
    # A JSON array of string transaction PHIDs that may be sec-approval request comments.
```

landoapi/models/secapproval.py
```diff
@@ -0,0 +1,61 @@
...
+            A `SecApprovalRequest` that is ready to be added to the session.
+        """
+        possible_comment_phids = []
+        for transaction in transactions:
+            phid = PhabricatorClient.expect(transaction, "phid")
+            possible_comment_phids.append(phid)
+
+        return cls(
+            revision_id=revision["id"],
+            diff_phid=revision["fields"]["diffPHID"],
```
smacleod: These should probably use `PhabricatorClient.expect`

landoapi/api/revisions.py
```diff
@@ -0,0 +1,92 @@
...
+    # quickly fetch the comment from Phabricator later in the process.
+    #
+    # NOTE: Each call to Phabricator returns two transactions: one for adding the
+    # comment and one for adding the reviewer.  We don't know which transaction is
+    # which at this point so we record both of them.
+    sa_request = SecApprovalRequest.build(revision, resulting_transactions)
+    db.session.add(sa_request)
+    db.session.commit()
+
+    return "OK", 200
```
smacleod: gah, missed this, will apply myself and land. All responses should be JSON.
```suggestion
    return {}, 200
```

PR #72, opened 2019-05-29
[sec-approval/3] client: wrap the transaction.search endpoint (bug 1553715)

mars-f:
Wrap the Phabricator API's 'transaction.search' endpoint with some
useful functions.

### DIFF
--- a/landoapi/transactions.py
+++ b/landoapi/transactions.py
@@ -5,7 +5,9 @@
 from landoapi.phabricator import PhabricatorClient
 
 
-def transaction_search(phabricator, object_identifier, limit=100):
+def transaction_search(
+    phabricator, object_identifier, transaction_phids=None, limit=100
+):
     """Yield the Phabricator transactions related to an object.
 
     See https://phabricator.services.mozilla.com/conduit/method/transaction.search/.
@@ -17,6 +19,8 @@ def transaction_search(phabricator, object_identifier, limit=100):
         phabricator: A PhabricatorClient instance.
         object_identifier: An object identifier (PHID or monogram) whose transactions
             we want to fetch.
+        transaction_phids: An optional list of specific transactions PHIDs we want to
+            find for the given object_identifier.
         limit: Integer keyword, limit the number of records retrieved per API call.
             Default is 100 records.
 
@@ -25,17 +29,23 @@ def transaction_search(phabricator, object_identifier, limit=100):
     """
     next_page_start = None
 
+    if transaction_phids:
+        constraints = {"phids": transaction_phids}
+    else:
+        constraints = {}
+
     while True:
         transactions = phabricator.call_conduit(
             "transaction.search",
             objectIdentifier=object_identifier,
+            constraints=constraints,
             limit=limit,
             after=next_page_start,
         )
 
         yield from PhabricatorClient.expect(transactions, "data")
 
-        next_page_start = transactions["cursor"]["after"]
+        next_page_start = PhabricatorClient.expect(transactions, "cursor", "after")
 
         if next_page_start is None:
             # This was the last page of results.
@@ -50,4 +60,7 @@ def get_raw_comments(transaction):
 
     See https://phabricator.services.mozilla.com/conduit/method/transaction.search/.
     """
-    return [comment["content"]["raw"] for comment in transaction["comments"]]
+    return [
+        PhabricatorClient.expect(comment, "content", "raw")
+        for comment in PhabricatorClient.expect(transaction, "comments")
+    ]
--- a/landoapi/transactions.py
+++ b/landoapi/transactions.py
@@ -1,0 +1,53 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Functions for working with Phabricator transactions."""
+from landoapi.phabricator import PhabricatorClient
+
+
+def transaction_search(phabricator, object_identifier, limit=100):
+    """Yield the Phabricator transactions related to an object.
+
+    See https://phabricator.services.mozilla.com/conduit/method/transaction.search/.
+
+    If the transaction list is larger that one page of API results then the generator
+    will call the Phabricator API successive times to fetch the full transaction list.
+
+    Args:
+        phabricator: A PhabricatorClient instance.
+        object_identifier: An object identifier (PHID or monogram) whose transactions
+            we want to fetch.
+        limit: Integer keyword, limit the number of records retrieved per API call.
+            Default is 100 records.
+
+    Returns:
+        Yields individual transactions.
+    """
+    next_page_start = None
+
+    while True:
+        transactions = phabricator.call_conduit(
+            "transaction.search",
+            objectIdentifier=object_identifier,
+            limit=limit,
+            after=next_page_start,
+        )
+
+        yield from PhabricatorClient.expect(transactions, "data")
+
+        next_page_start = transactions["cursor"]["after"]
+
+        if next_page_start is None:
+            # This was the last page of results.
+            return
+
+
+def get_raw_comments(transaction):
+    """Return a list of 'raw' comment bodies in a Phabricator transaction.
+
+    A single transaction can have multiple comment bodies: e.g. a top-level comment
+    and a couple of inline comments along with it.
+
+    See https://phabricator.services.mozilla.com/conduit/method/transaction.search/.
+    """
+    return [comment["content"]["raw"] for comment in transaction["comments"]]
--- a/tests/mocks.py
+++ b/tests/mocks.py
@@ -39,6 +39,7 @@ class PhabricatorDouble:
     """
 
     def __init__(self, monkeypatch):
+        self._transactions = []
         self._users = []
         self._projects = []
         self._revisions = []
@@ -79,6 +80,7 @@ class PhabricatorDouble:
         depends_on=[],
         bug_id=None,
         projects=[],
+        comments=[],
     ):
         revision_id = self._new_id(self._revisions)
         phid = self._new_phid("DREV-")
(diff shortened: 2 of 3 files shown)


tests/mocks.py
```diff
@@ -145,6 +147,19 @@ def revision(
             }
         )
 
+        next_free_txn_id = len(self._transactions)
```
smacleod:
```suggestion
```

tests/mocks.py
```diff
@@ -145,6 +147,19 @@ def revision(
...
+        for txn_id, comment in enumerate(comments, start=next_free_txn_id):
```
smacleod:
```suggestion
        for comment in comments:
```

tests/mocks.py
```diff
@@ -145,6 +147,19 @@ def revision(
...
+            # See https://phabricator.services.mozilla.com/conduit/method/transaction.search/  # noqa
+            # for the Conduit API result object structure.
+            self._transactions.append(
+                {
+                    "id": txn_id,
```
smacleod: This should use the id generation helper method and transactions (really, every object) needs a phid:
```suggestion
                    "id": self._new_id(self._transactions),
                    "phid": self._new_phid("PHID-XACT-DREV"),
```

tests/mocks.py
```diff
@@ -145,6 +147,19 @@ def revision(
             }
         )
 
+        next_free_txn_id = len(self._transactions)
+        for txn_id, comment in enumerate(comments, start=next_free_txn_id):
+            # See https://phabricator.services.mozilla.com/conduit/method/transaction.search/  # noqa
+            # for the Conduit API result object structure.
+            self._transactions.append(
```
smacleod: I wonder if it's worth sticking all the transaction generation logic into a `def transaction` on `PhabricatorDouble`? As we add more uses of transactions (there are a number of features coming that need it), we'll have to represent the other types. Would make this `PhabricatorDouble.revision` method simpler by having it just do something like `self.transaction(on=phid, comments=comments)`

tests/mocks.py
```diff
@@ -145,6 +147,19 @@ def revision(
...
+                {
+                    "id": txn_id,
+                    "type": "comment",
+                    "objectPHID": phid,
+                    "comments": [{"content": {"raw": comment}}],
+                }
```
smacleod: This is missing a bunch of fields that transactions have, which should be there even if filled with bogus data:
```json
{
      "id": 1003798,
      "phid": "PHID-XACT-DREV-3qlwutlsdcfwrge",
      "type": "subscribers",
      "authorPHID": "PHID-APPS-PhabricatorHeraldApplication",
      "objectPHID": "PHID-DREV-pm6r32ocou6ocjik4i5n",
      "dateCreated": 1559779750,
      "dateModified": 1559779750,
      "groupID": "zmgrbvzbcclaubtg4yt2ihwwotxewc4h",
      "comments": [],
      "fields": {
        "operations": [
          {
            "operation": "add",
            "phid": "PHID-USER-zp4gy3jjzfehipsktf6x"
          }
        ]
      }
```
It'd be good to run some `transaction.search`'s on production Phabricator and figure out what exactly is there.

You can take a look athow some of the other methods handle things like `dateCreated` etc.

tests/mocks.py
```diff
@@ -1037,6 +1052,47 @@ def to_response(i):
...
+        before=None,
+        after=None,
+        limit=100,
+    ):
+        matching_transactions = [
+            t for t in self._transactions if t["objectPHID"] == objectIdentifier
+        ]
+        # We don't want tests that work with our returned result to accidentally
+        # modify a transaction object.
+        matching_transactions = deepcopy(matching_transactions)
```
smacleod: Convention is to give every method a `to_response(i)` which deals with serializing an item. Fields are all explicitly copied so if anything needs to be added in the future (such as linking to other objects, this happened with edges in the past) you don't start spilling extra data into responses. It'd be good to explicitly fail in the case of a transaction type which we haven't added support for yet.
```suggestion
        def to_response(i):
            if i["type"] == "comment":
                return deepcopy(
                    {
                        "id": i["id"],
                        "type": i["type"],
                        "objectPHID": i["objectPHID"],
                        "comments": i["comments"],
                        # ... TODO: rest of required items here.
                    }
                )
            
            raise ValueError(
                "PhabricatorDouble transactions do not have support"
                'for the "{}" transaction type. '
                "If you have added use of a new transaction type please "
                "update PhabricatorDouble to support it.".format(i["type"])
            )
```

tests/mocks.py
```diff
@@ -1037,6 +1052,47 @@ def to_response(i):
...
+        self,
+        *,
+        objectIdentifier=None,
+        constraints={},
+        order=None,
+        before=None,
+        after=None,
+        limit=100,
+    ):
+        matching_transactions = [
```
smacleod: Convention is to call this `items`

tests/mocks.py
mars-f: I agree, I've added the new method and the code for both adding transactions and creating revisions is now easier to read.

tests/mocks.py
mars-f: I've added a bunch.  See the new `transaction()` method.

tests/mocks.py
mars-f: I added the `to_response()` function but kept the type-checking code out of the method.  Transactions are so varied in structure that I think it makes sense for the method caller to have the freedom to put whatever they want in there.  I did this in the test code and it was nice to use.  Also it would take a long time to reverse engineer every transaction type, codify, then enforce its structure.  I don't see the benefit to writing that code at this early point with so few uses of the transaction code in the codebase.

tests/mocks.py
```diff
@@ -1165,6 +1294,19 @@ def phid_query(self, *, phids=None):
         return {i["phid"]: deepcopy(i) for i in self._phids if i["phid"] in phids}
 
     def _new_phid(self, prefix):
+        """Generate a unique PHID of the given type, e.g. 'PHID-DREV-123'.
+
+        For example, given the prefix 'DREV-', the function will generate a
+        PHID of 'PHID-DREV-0'. Given the prefix of 'DIFF-' it will return
+        'PHID-DIFF-abc123'.
```
smacleod:
```suggestion
        'PHID-DIFF-0'.
```

tests/mocks.py
```diff
@@ -1165,6 +1294,19 @@ def phid_query(self, *, phids=None):
...
+
+        PHIDs are number starting at zero and proceed sequentially for each type. For
```
smacleod:
```suggestion
        Generated PHIDs start at zero and proceed sequentially for each type. For
```

tests/mocks.py
smacleod: Sorry, to be clear I'm not suggesting you should type check against all the real transactions. What I mean is type checking for only the transactions that this mock can properly represent. The reason for this is so that if someone adds a new use of a transaction type to Lando the `PhabricatorDouble` won't silently represent it the wrong way, tricking the author into thinking things are working when they're using the wrong structure for that transaction type. If it turns out "TransactionA" is the same structure as "comment" lets say, then the dev implementing things will explicitly mark that and move on.

tests/mocks.py
```diff
@@ -1037,6 +1105,67 @@ def to_response(i):
...
+                    "dateModified": i["dateModified"],
+                    "groupID": i["groupID"],
+                    "comments": i["comments"],
+                    "fields": i["fields"],
+                }
+            )
+
+        items = list(self._transactions)
+
+        if objectIdentifier:
```
smacleod: `objectIdentifier` is required by transaction search:
```json
{"result":null,"error_code":"ERR-CONDUIT-CORE","error_info":"When calling \"transaction.search\", you must provide an object to retrieve transactions for."}
```

See other examples in this file where we raise a `PhabricatorAPIException`

tests/mocks.py
```diff
@@ -409,6 +416,67 @@ def project(self, name, *, no_slug=False):
...
+            "comments": comments,
+            "fields": fields,
+        }
+        self._transactions.append(transaction)
+        self._phids.append(
+            {
+                "phid": phid,
+                # Unlike other objects transactions don't have a URI. This field
+                # is just the hostname with no path component.
+                "uri": "https://phabricator.services.mozilla.com",
```
smacleod: http://phabricator.test/ is used everywhere else in this file.
```suggestion
                "uri": "http://phabricator.test/",
```

landoapi/transactions.py
```diff
@@ -0,0 +1,53 @@
...
+
+def get_raw_comments(transaction):
+    """Return a list of 'raw' comment bodies in a Phabricator transaction.
+
+    A single transaction can have multiple comment bodies: e.g. a top-level comment
+    and a couple of inline comments along with it.
+
+    See https://phabricator.services.mozilla.com/conduit/method/transaction.search/.
+    """
+    return [comment["content"]["raw"] for comment in transaction["comments"]]
```
smacleod: `PhabricatorClient.expect` should be used.

landoapi/transactions.py
```diff
@@ -0,0 +1,53 @@
...
+        transactions = phabricator.call_conduit(
+            "transaction.search",
+            objectIdentifier=object_identifier,
+            limit=limit,
+            after=next_page_start,
+        )
+
+        yield from PhabricatorClient.expect(transactions, "data")
+
+        next_page_start = transactions["cursor"]["after"]
```
smacleod: `PhabricatorClient.expect` should be used.

landoapi/transactions.py
```diff
@@ -0,0 +1,53 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""Functions for working with Phabricator transactions."""
+from landoapi.phabricator import PhabricatorClient
+
+
+def transaction_search(phabricator, object_identifier, limit=100):
```
smacleod: We'll want a way to specify the constraints. At least the phid one. Maybe that's coming in a future change though when you make use of this function?

mars-f: The missing PHID constraint is a hold-over from the previous implementation of revision comment retrieval.  I've added a new parameter to the method to constrain searches to specific PHIDs.

PR #73, opened 2019-05-29
WIP [sec-approval/4] api: publish secure commit messages (bug 1553715)

mars-f:
If a revision has a secure alternate commit message available in
Phabricator then substitute it for the normal commit message when a
revision's data is retrieved via the API.

Also refactor the new bulk of sec-approval-related functionality into
its own module.


mars-f:
Superceded by #84.

PR #75, opened 2019-07-31
stacks: provide a more specific error message when no repo is specifed for a revision (Bug 1569721) r?smacleod

cgsheeh:
In `calculate_landable_subgraphs` we iterate over the repository
values attached to the revisions in our stack and check if the
repo is valid using `repo not in landable_repos`. This blocks
the revision from landing and displays a sensible error message
about an unsupported repo.

However, in the case a repo has not been associated with a
revision, the condition will still be true as the empty value is
not in the set of landable repos. This causes confusion to end
users regarding what to do to fix the problem. In this commit we
add an extra check to assert the value of `repo` is not a
False-y value. If the value is False-y, we provide a more instructive
error message and block the revision from landing. Otherwise,
we continue with the existing block logic.

### DIFF
--- a/landoapi/stacks.py
+++ b/landoapi/stacks.py
@@ -169,10 +169,16 @@ def calculate_landable_subgraphs(
     # We won't land anything that has a repository we don't support, so make
     # a pass over all the revisions and block these.
     for phid, revision in revision_data.revisions.items():
-        if (
-            PhabricatorClient.expect(revision, "fields", "repositoryPHID")
-            not in landable_repos
-        ):
+        repo = PhabricatorClient.expect(revision, "fields", "repositoryPHID")
+        if not repo:
+            block(
+                phid,
+                "Revision's repository unset. Specify a target using"
+                '"Edit revision" in Phabricator',
+            )
+            continue
+
+        if repo not in landable_repos:
             block(phid, "Repository is not supported by Lando.")
 
     # We only want to consider paths starting from the open revisions
--- a/tests/test_stacks.py
+++ b/tests/test_stacks.py
@@ -550,6 +550,31 @@ def test_calculate_landable_subgraphs_extra_check(phabdouble):
     assert blocked[r3["phid"]] == REASON
 
 
+def test_calculate_landable_subgraphs_missing_repo(phabdouble):
+    """Test to assert a missing repository for a revision is
+    blocked with an appropriate error
+    """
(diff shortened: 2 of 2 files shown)


tests/test_stacks.py
```diff
@@ -576,31 +576,41 @@ def test_integrated_stack_endpoint_simple(client, phabdouble, mocked_repo_config
...
     assert r2["phid"] in revisions
     assert r3["phid"] in revisions
     assert r4["phid"] in revisions
+    assert r5["phid"] in revisions
 
     assert revisions[r4["phid"]]["blocked_reason"] == (
         "Repository is not supported by Lando."
     )
 
+    assert revisions[r5["phid"]]["blocked_reason"] == (
```
smacleod: This change is testable from a unit test, I think I'd prefer if we had something mimicking one of the `test_calculate_landable_subgraphs_*` tests, specifically for this case, rather than making this test more complicated. Because they only rely on the `phabdouble` fixture, they're much much cheaper than anything relying on `client` etc. (which requires things like the database).

landoapi/stacks.py
```diff
@@ -169,10 +169,18 @@ def block(node, reason):
...
-            not in landable_repos
-        ):
+        repo = PhabricatorClient.expect(revision, "fields", "repositoryPHID")
+        if not repo:
+            block(
+                phid,
+                "No repository configured for this revision. Please "
+                "specify a repository for this revision using the "
+                "'Edit revision' option in the top-right of the "
+                "Phabricator UI.",
```
smacleod: These messages end up in a small tooltip, so keeping it a little less verbose would probably be preferrable. How about something like: `'Revision\'s repository unset. Specify a target using "Edit revision" in Phabricator'`

tests/test_stacks.py
```diff
@@ -550,6 +550,31 @@ def custom_check(*, revision, diff, repo):
...
+    landable, blocked = calculate_landable_subgraphs(
+        revision_data, edges, {repo1["phid"]}
+    )
+
+    repo_unset_warning = (
+        "Revision's repository unset. Specify a target using"
+        '"Edit revision" in Phabricator'
+    )
+
+    assert not landable
```
cgsheeh: I don't know enough about Lando-API to know if we need a more complicated stack for the test, so I decided to write a minimal test (single revision, single repo).

PR #81, opened 2019-10-08
tests: mock Phabricator comments with PhabDouble

mars-f:
Allow the mocking of Phabricator comment objects, like you would have
from adding a comment to a Revision. Having comments properly mocked
makes the Transaction mocks match real Phabricator Transaction objects
more closely.

### DIFF
--- a/tests/mocks.py
+++ b/tests/mocks.py
@@ -48,6 +48,7 @@ class PhabricatorDouble:
         self._repos = []
         self._diffs = []
         self._diff_refs = []
+        self._comments = []
         self._phids = []
         self._phid_counters = {}
         self._edges = []
@@ -149,9 +150,7 @@ class PhabricatorDouble:
         )
 
         for comment in comments:
-            self.transaction(
-                "comment", revision, comments=[{"content": {"raw": comment}}]
-            )
+            self.transaction("comment", revision, comments=[comment])
 
         return revision
 
@@ -433,7 +432,7 @@ class PhabricatorDouble:
                 e.g. When adding a subscriber to a revision the transaction type would
                 be "subscriber" and the operations argument would be
                 [{ "operation": "add", "phid": "PHID-USER-abc123"}]
-            comments: Optional list of comments modified by this transaction. Only
+            comments: Optional list of comment objects this transaction created. Only
                 applies when the transaction type is "comment" or "inline". Structure
                 varies greatly depending on the comment type.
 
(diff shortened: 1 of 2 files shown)


tests/mocks.py
```diff
@@ -1146,7 +1174,11 @@ def to_response(i):
         items = list(self._transactions)
 
         if objectIdentifier:
-            items = [i for i in items if i["objectPHID"] == objectIdentifier]
+            items = [
+                i
+                for i in items
+                if i["objectPHID"] == objectIdentifier or i["id"] == objectIdentifier
+            ]
```
smacleod: This is not how the Phabricator API behaves, why are you making this change?

tests/mocks.py
```diff
@@ -478,6 +477,35 @@ def transaction(
...
+            "dateCreated": 1570502242,
+            "dateModified": 1570502242,
+            "removed": False,
+            "content": {"raw": content},
+        }
+
+        self._comments.append(comment)
+        # NOTE: Comments are an internal type. They cannot be introspected. We don't
+        # know the value of a number of the fields that other API objects have,
+        # such as 'type' and 'uri'.
```
smacleod: This comment placement confuses me, what is it supposed to apply to? I'm not exactly sure what this is trying to say or what makes comments unique compared to all the other internal representations we have?

tests/mocks.py
mars-f: That's a bug.  Searching by `objectIdentifier` does a fuzzy match on PHID or monogram.  There is no search by ID.

tests/mocks.py
mars-f: I removed the comment and the PHID registration since calling `phid.lookup` on a comment doesn't return any result with a real Phabricator instance.

mars-f:
> What is the overall context of this change/is there a larger purpose that will build on this? The actual mocking of the comment data structure was already encapsulated (inside `revision()`), so adding a `comment()` function at this point just added extra steps, is there something else you're planning to use it for?

As discussed via chat, this change is part of the sec-approval request feature.  sec-approval works directly with the comment objects in transactions so having better transactions mocks gives me a higher confidence that the code is correct.

tests/mocks.py
```diff
@@ -478,6 +477,31 @@ def transaction(
...
 
+    def comment(self, content, author=None):
+        """Return a Phabricator Comment object.
+
+        Args:
+            content: The raw (Remarkup) comment contents.
+            author: Optional Phabricator User that authored the comment. If None
+                then a new User will be generated.
+        """
+        phid = self._new_phid("XCMT-")
```
smacleod: It's very strange that comments have PHIDs, but can't be found with a `phid.query`... wonder if that's considered an upstream bug or is intentional?

PR #85, opened 2019-10-16
tests: add detail to failed API responses in tests

mars-f:
Add a pytest custom assertion handler to decorate failed API responses.
In addition to reporting the failed HTTP status code the failed response 
JSON, including the error message within it, are reported in the test 
output.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -338,7 +338,7 @@ def pytest_assertrepr_compare(op, left, right):
         #   response = client.get()
         #   assert response == 200  # Fails if response is HTTP 401, triggers this hook
         return [
-            f"Comparing API response:",
-            f"    status_code: {left.status_code}",
-            f"    problem: {left.json}",
+            f"Mismatch in status code for response: {left.status_code} != {right}",
+            f"",
+            f"    Response JSON: {left.json}",
         ]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,7 @@ import pytest
 import requests_mock
 from flask import current_app
 from moto import mock_s3
+from pytest_flask.plugin import JSONResponse
 
 from landoapi.app import construct_app, load_config, SUBSYSTEMS
 from landoapi.cache import cache, cache_subsystem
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -325,3 +326,19 @@ def celery_app(app):
...
+
+def pytest_assertrepr_compare(op, left, right):
+    if isinstance(left, JSONResponse) and isinstance(right, int) and op == "==":
+        # Hook failures when comparing JSONResponse objects so we get the detailed
+        # failure description from inside the response object contents.
+        #
+        # The following example code would trigger this hook:
+        #
+        #   response = client.get()
+        #   assert response == 200  # Fails if response is HTTP 401, triggers this hook
```
smacleod: Why are you comparing the response object itself to the status code? I'm not sure what implementing this hook is providing to us here?

tests/conftest.py
```diff
@@ -325,3 +326,19 @@ def celery_app(app):
...
+        return [
+            f"Comparing API response:",
+            f"    status_code: {left.status_code}",
+            f"    problem: {left.json}",
+        ]
```
smacleod: IIUC, won't this mean the right hand side is completely missing from the output of the test? I'd consider that a serious regression in test output quality.

mars-f: Here's an example of the test failure output (edited for clarity):

```
====================================================== FAILURES =======================================================
_______________________________________ test_stack_has_alternate_commit_message _______________________________________

    def test_stack_has_alternate_commit_message(
        db, client, phabdouble, mock_repo_config, secure_project, authed_headers
    ):
       ...

        response = client.post(....)

>       assert response == 200
E       AssertionError: assert Comparing API response:
E             status_code: 401
E             problem: {'detail': 'Authorization header is required and was not provided', 'status': 401, 'title': 'Authorization Header Required', 'type': 'https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401'}

tests/test_sanitized_commit_messages.py:102: AssertionError
```

The test output in the console is colourized, the failing line with the right-hand-side is highlighted, and the status code we expected to see is by convention an integer literal.

You can also shut off the decorated assertion by changing the comparison to `int == int`, like so:

```
assert response.status_code == expected_status_code
```

mars-f: I've updated the output text to resemble the original message in pytest-flask and to include the right-hand side data.  Here's a sample:

```
>       assert response == 202
E       AssertionError: assert Mismatch in status code for response: 400 != 202
E         
E             Response JSON: {'blocker': None, 'confirmation_token': ...<bunch of JSON>
E         

tests/test_sanitized_commit_messages.py:242: AssertionError

```

tests/conftest.py
mars-f: There is also a custom assertion reporter in pytest-flask that closely resembles what I wrote here.  My version includes the JSON response body where the pytest-flask reporter provides just the status code and status name.

PR #86, opened 2019-10-16
tests: transactions can be searched by object name

mars-f:
Transaction objects can now be searched for by both PHID and by
object name, such as "D123".  This is necessary for situations where
only the object name is locally available without a network
request. It also matches the Phabricator API, which allows searches
by both PHID and object name.

### DIFF
--- a/tests/mocks.py
+++ b/tests/mocks.py
@@ -1145,8 +1145,29 @@ class PhabricatorDouble:
 
         items = list(self._transactions)
 
+        def filter_by_phid(iterable, phid):
+            return [obj for obj in iterable if obj["objectPHID"] == phid]
+
         if objectIdentifier:
-            items = [i for i in items if i["objectPHID"] == objectIdentifier]
+            if objectIdentifier.startswith("PHID-"):
+                items = filter_by_phid(items, objectIdentifier)
+            elif objectIdentifier.startswith("D"):
+                # Search for the matching object by name.
+                matches = [
+                    obj for obj in self._phids if obj.get("name") == objectIdentifier
+                ]
+                if matches:
+                    obj = matches[0]
+                    items = filter_by_phid(items, obj["phid"])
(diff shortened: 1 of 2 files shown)


tests/mocks.py
```diff
@@ -1145,8 +1145,29 @@ def to_response(i):
...
+            else:
+                raise ValueError(
+                    f"PhabricatorDouble transaction search does not support the "
+                    f'objectIdentifier "{objectIdentifier}". If you require '
+                    f"transaction searches of this type please consider adding "
+                    f"support for them."
+                )
         else:
             error_info = 'When calling "transaction.search", you must provide an object to retrieve transactions for.'  # noqa
             raise PhabricatorAPIException(
```
smacleod: This doesn't quite match the real behaviour and is really nested. How about something like:
```Python
        if not objectIdentifier:
            error_info = 'When calling "transaction.search", you must provide an object to retrieve transactions for.'  # noqa
            raise PhabricatorAPIException(
                error_info, error_code="ERR-CONDUIT-CORE", error_info=error_info
            )


        if not objectIdentifier.startswith("PHID-"):
            matches = [
                obj for obj in self._phids if obj.get("name") == objectIdentifier
            ]
            if not matches:
                error_info = f'No object "{objectIdentifier}" exists.'
                raise PhabricatorAPIException(
                    error_info, error_code="ERR-CONDUIT-CORE", error_info=error_info
                )
            
            objectIdentifier = matches[0]["phid"]

        # TODO: Check that transactions are supported on the object type of
        # the phid in objectIdentifier.
        # TODO: Error out if PhabDouble does not support transactions for the
        # type of phid in objectIdentifier.
        items = [i for i in items if i["objectPHID"] == objectIdentifier]
```

mars-f: I like it. Fixed.

PR #87, opened 2019-10-20
api: land commits using sec-approval commit messages (bug 1569209)

mars-f:
When a user requests landing of a commit that has an alternate commit 
message going through the sec-approval process use that alternate sanitized
commit title and commit summary as the commit's title and summary.


mars-f:
rebase

PR #89, opened 2019-11-28
docs: fix the name of the revision ID field in swagger.json

mars-f:


mars-f:
@smacleod I'm not entirely sure if the bug is the field name in the docs, `revision_id`, or the field name returned by the code, `id`.  Other endpoints use `revision_id`.  However it's harder to change our deployed code and API clients that use the field name `id`, so I went with fixing the docs.

PR #91, opened 2019-12-17
uplift: Use bugzilla id only when available

La0:
I encountered a bug while playing with the UI: revisions without a bugzilla id crash the uplift creation process.


zzzeid:
Closing for the time being as stale, can reopen/revisit when we look at uplift support.

PR #92, opened 2020-01-07
commit message: allow single period in nickname (bug 1515217)

zzzeid:
Fix that prevents Lando from mangling the reviewer names when there is a period. See https://bugzilla.mozilla.org/show_bug.cgi?id=1515217 for more info.

### DIFF
--- a/landoapi/commit_message.py
+++ b/landoapi/commit_message.py
@@ -39,7 +39,7 @@ LIST = r"[;,\/\\]\s*"
 
 # Note that we only allows a subset of legal IRC-nick characters.
 # Specifically, we do not allow [ \ ] ^ ` { | }
-IRC_NICK = r"[a-zA-Z0-9\-\_]+"
+IRC_NICK = r"[a-zA-Z0-9\-\_.]*[a-zA-Z0-9\-\_]+"
 
 # fmt: off
 REVIEWERS_RE = re.compile(  # noqa: E131
--- a/tests/test_commit_message.py
+++ b/tests/test_commit_message.py
@@ -6,32 +6,20 @@ import pytest
 from landoapi.commit_message import format_commit_message, split_title_and_summary
 
 COMMIT_MESSAGE = """
-Bug 1 - A title. r=reviewer_one,reviewer_two
+Bug 1 - A title. r=reviewer_one,reviewer.two
 
 A summary.
 
 Differential Revision: http://phabricator.test/D123
 """.strip()
 
-FIRST_LINE = "Bug 1 - A title. r=reviewer_one,reviewer_two"
+FIRST_LINE = COMMIT_MESSAGE.split("\n")[0]
 
 
 def test_commit_message_for_multiple_reviewers():
-    reviewers = ["reviewer_one", "reviewer_two"]
+    reviewers = ["reviewer_one", "reviewer.two"]
     commit_message = format_commit_message(
         "A title.", 1, reviewers, "A summary.", "http://phabricator.test/D123"
-    )
-    assert commit_message == (FIRST_LINE, COMMIT_MESSAGE)
-
-
-def test_commit_message_reviewers_replaced():
-    reviewers = ["reviewer_one", "reviewer_two"]
-    commit_message = format_commit_message(
-        "A title. r=not_reviewer r?bogus",
-        1,
-        reviewers,
-        "A summary.",
-        "http://phabricator.test/D123",
     )
     assert commit_message == (FIRST_LINE, COMMIT_MESSAGE)
 
@@ -82,18 +70,25 @@ def test_commit_message_blocking_reviewers_requested(reviewer_text):
 @pytest.mark.parametrize(
     "reviewer_text",
     [
+        "r?bogus",
         "r?#group1",
         "r?#group1, #group2",
         "r?reviewer_one,#group1",
-        "r?#group1 r?reviewer_two",
+        "r?#group1 r?reviewer.two",
         "r?#group1! r?group2",
         "r?#group1 r?group2!",
         "r?#group1! r?group2!",
-        "r?#group1, reviewer_two!",
+        "r?#group1, reviewer.two!",
+        "r=.a",
+        "r=..a",
+        "r=a...a",
+        "r=a.b",
+        "r=a.b.c",
+        "r=aa,.a,..a,a...a,a.b,a.b.c",
     ],
 )
-def test_group_reviewers_replaced(reviewer_text):
-    reviewers = ["reviewer_one", "reviewer_two"]
+def test_commit_message_reviewers_replaced(reviewer_text):
+    reviewers = ["reviewer_one", "reviewer.two"]
     commit_message = format_commit_message(
         "A title. {}".format(reviewer_text),
         1,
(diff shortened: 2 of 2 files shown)


purelogiq:
steven/mars would be good for this review since they work on lando-api more often. glob would also be good since he reviewed the related change in mozphab. Thanks for the fix!

landoapi/commit_message.py
```diff
@@ -39,7 +39,7 @@
 
 # Note that we only allows a subset of legal IRC-nick characters.
 # Specifically, we do not allow [ \ ] ^ ` { | }
-IRC_NICK = r"[a-zA-Z0-9\-\_]+"
+IRC_NICK = r"[a-zA-Z0-9\-\_.]+[a-zA-Z0-9\-\_]+"
```
mars-f: The PR title says the patch enables author names with a single period.  However I got some pretty funky names with multiple periods to match the regex, too.  Is that behaviour OK?

```
aa
a.b
a.b.c
.a
...a
a...a
-.-.-
```

tests/test_commit_message.py
```diff
@@ -85,15 +85,15 @@ def test_commit_message_blocking_reviewers_requested(reviewer_text):
         "r?#group1",
         "r?#group1, #group2",
         "r?reviewer_one,#group1",
-        "r?#group1 r?reviewer_two",
+        "r?#group1 r?reviewer.two",
```
mars-f: The bug report mentions how a single period messed up the parsing of the text between two names.  This test data covers the single period case, but you may want to consider some test data that parses multiple names with periods, or a single name with multiple periods, depending on how the underlying parser goes about its job.

landoapi/commit_message.py
globau: Yes - we need to support multiple periods in names.

tests/test_commit_message.py
```diff
@@ -85,15 +85,15 @@ def test_commit_message_blocking_reviewers_requested(reviewer_text):
...
         "r?#group1! r?group2",
         "r?#group1 r?group2!",
         "r?#group1! r?group2!",
-        "r?#group1, reviewer_two!",
+        "r?#group1, reviewer.two!",
```
globau: I'd also like to see a test to ensure trailing periods aren't matched.

zzzeid: Noted.

landoapi/commit_message.py
zzzeid: Indeed supporting multiple periods would be good. I originally had this in my approach but simplified the regex to match a single period however we might as well support an arbitrary number of periods from the get go. Will update.

tests/test_commit_message.py
zzzeid: `format_commit_message` actually uses the RE differently from other repos when it comes to parsing reviewers  -- it's doing the reverse, i.e. creating a reviewer list from a predetermined list of reviewers fetched via the Phabricator client, and thus the behaviour here is quite different. What it does do though is look for the first occurrence of `r=` using the RE and go from there. The behaviour here is actually more desirable, however this is broken in other repos.

tests/test_commit_message.py
```diff
@@ -36,6 +36,58 @@ def test_commit_message_reviewers_replaced():
...
+    commit_message = format_commit_message(
+        "A title!", 1, reviewers, "", "http://phabricator.test/D123"
+    )
+
+    assert commit_message[0] == "Bug 1 - A title! r={}".format(",".join(reviewers))
+
+
+def test_period_at_end_of_reviewer_name():
+    """Tests that a period at the end of a reviewer's name does not get
+    parsed.
```
smacleod: This test doesn't actually end up parsing a reviewer which has a period at the end of the name. When you're calling...
```Python
    commit_message = format_commit_message(
        "A title! r=SDfsdfsd", 1, ["test.", "test2"], "", "http://phabricator.test/D123"
    )
```
...the third argument where you've added the the `.` is part of the list of reviewers to be written in, and isn't parsed at all. If you'd like to test the behaviour of parsing a reviewer with a trailing period it would have to be part of the first argument, which is the current title to be rewritten.

Currently, only the updates you made to `test_group_reviewers_replaced` parse reviewers that contain a period at all, the other changes just cause the formatter to write out reviewer names with periods.

tests/test_commit_message.py
```diff
@@ -36,6 +36,58 @@ def test_commit_message_reviewers_replaced():
...
+        ["a...a"],
+        ["a.b"],
+        ["a.b.c"],
+        ["-.-.-"],
+        ["a", "aa", ".a", "..a", "...a", "a...a", "a.b", "a.b.c", "-.-.-"],
+    ],
+)
+def test_various_reviewer_name_combinations(reviewers):
+    """Tests single and multiple characters separated by various combinations
+    of periods for single and multiple reviewers.
```
smacleod: nit: we follow [PEP 257](https://www.python.org/dev/peps/pep-0257) for our docstrings. Specifically your new tests violate...
> The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

...from [PEP 257 #multi-line-docstrings](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)

tests/test_commit_message.py
```diff
@@ -36,6 +36,58 @@ def test_commit_message_reviewers_replaced():
...
+        ["..a"],
+        ["...a"],
+        ["a...a"],
+        ["a.b"],
+        ["a.b.c"],
+        ["-.-.-"],
+        ["a", "aa", ".a", "..a", "...a", "a...a", "a.b", "a.b.c", "-.-.-"],
+    ],
+)
+def test_various_reviewer_name_combinations(reviewers):
```
smacleod: Since we don't actually parse any of these, and just write them out, this doesn't add much to our test coverage. Maybe you could re-work this test to mimic `test_group_reviewers_replaced`, but focused on our period parsing?

tests/test_commit_message.py
zzzeid: I removed this test and updated the other existing tests to be inline with this.

tests/test_commit_message.py
zzzeid: Noted and moved a subset of these to `test_group_reviewers_replaced`.

tests/test_commit_message.py
```diff
@@ -104,6 +110,34 @@ def test_group_reviewers_replaced(reviewer_text):
...
+
+    # This is the current behaviour
+    assert commit_message == (
+        "Bug 1 - A title. r=reviewer_one,reviewer.two.,b",
+        "Bug 1 - A title. r=reviewer_one,reviewer.two.,b\n\n"
+        "A summary.\n\nDifferential Revision: http://phabricator.test/D123",
+    )
+
+    # This is the desired future behaviour
+    assert commit_message != (
```
smacleod: For things like this you could always write the test as if it would pass but put an `@pytest.mark.xfail(strict=True)` so that we update the test when we fix things. See https://docs.pytest.org/en/latest/skipping.html#xfail

tests/test_commit_message.py
```diff
@@ -85,15 +85,21 @@ def test_commit_message_blocking_reviewers_requested(reviewer_text):
...
+        "r?#group1, reviewer.two!",
+        "r=.a",
+        "r=..a",
+        "r=a...a",
+        "r=a.b",
+        "r=a.b.c",
+        "r=aa,.a,..a,a...a,a.b,a.b.c",
     ],
 )
 def test_group_reviewers_replaced(reviewer_text):
```
smacleod: Nit: This test could maybe use a rename now that it's not just focused on groups.

zzzeid: Done. Also consolidated another test into this one.

tests/test_commit_message.py
zzzeid: I considered using that but wanted to include the passing case as well, but now that I think about it, I can still include the passing case for reference anyway! Changed.

PR #93, opened 2020-01-08
[sec-approval#1] api: rename landoapi.api.revisions

mars-f:
Clarify the name of the landoapi.api.revisions module.  It only contains
secapproval code so we'll rename it to match.  This also adds symmetry
with the uplift API module.

### DIFF
--- a/landoapi/api/secapproval.py
+++ b/landoapi/api/secapproval.py
@@ -6,7 +6,6 @@ import logging
 
 from connexion import problem
 from flask import g
-
 from landoapi import auth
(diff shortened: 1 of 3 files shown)


landoapi/api/secapproval.py
```diff
@@ -6,7 +6,6 @@
 
 from connexion import problem
 from flask import g
-
```
smacleod: This blank line should not be removed.

zzzeid:
Closing for the time being as this PR is stale. Will revisit/reopen if needed when we revisit sec-approval workflow.

PR #94, opened 2020-01-08
WIP [sec-approval#2] api: rename requestSecApproval

mars-f:
Rename the requestSecApproval method to the more specific
requestSecApprovalForCommitMessage.  Moving the old name aside allows us
 to use the original endpoint name for sec-approval form submission.


mars-f:
Taking a different approach in #96 where we use a single API endpoint for both sec-approval creation and sec-approval updates.

PR #97, opened 2020-01-23
Integrated transplant functionality (1/2)

smacleod:
WIP dump of the final pieces of the transplant functionality in Lando while I finish up the last few TODOs.


landoapi/models/landing_job.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @property
+    def landing_path(self):
+        return [(int(r), self.revision_to_diff_id[r]) for r in self.revision_order]
+
+    @property
+    def head_revision(self):
+        """Human-readable representation of the branch head's Phabricator revision ID.
+        """
+        assert (
```
zzzeid: Not sure if the use of `assert` here is ideal -- if we want to enforce this usage (outside of debugging), a specific exception should be raised instead. As an example, asserts are dropped in the bytecode if `-o` is used (see https://docs.python.org/3/using/cmdline.html#cmdoption-o)

landoapi/models/landing_job.py
```diff
@@ -0,0 +1,209 @@
...
+    attempts = db.Column(db.Integer, nullable=False, default=0)
+
+    # Priority of the job. Higher values are processed first.
+    priority = db.Column(db.Integer, nullable=False, default=0)
+
+    def __repr__(self):
+        return "<LandingJob: %s>" % self.id
+
+    @property
+    def landing_path(self):
```
zzzeid: Docstring missing frmo this.

landoapi/models/landing_job.py
```diff
@@ -0,0 +1,209 @@
...
+    def head_revision(self):
+        """Human-readable representation of the branch head's Phabricator revision ID.
+        """
+        assert (
+            self.revision_order
+        ), "head_revision should never be called without setting self.revision_order!"
+        return "D" + self.revision_order[-1]
+
+    @classmethod
+    def revisions_query(cls, revisions):
```
zzzeid: Docstring missing from this.

landoapi/models/landing_job.py
```diff
@@ -0,0 +1,209 @@
...
+        """Return a query which selects the next job and locks the row."""
+        q = cls.job_queue_query(repositories=repositories)
+
+        # Returned rows should be locked for updating, this ensures the next
+        # job can be claimed.
+        q = q.with_for_update()
+
+        return q
+
+    def landed(self, commit_id):
```
zzzeid: nit: Should this be renamed to `land`? I.e. when using it, `landing_job.land(<commit_id>)` would make a little more sense than this. Same with `failed_transient` and `failed_permanent` below.

landoapi/models/landing_job.py
```diff
@@ -0,0 +1,209 @@
...
+        self.status = LandingJobStatus.IN_PROGRESS
+        self.error = msg
+
+    def failed_permanent(self, msg):
+        """Mark the job as permanently failed."""
+        self.status = LandingJobStatus.FAILED
+        self.error = msg
+
+    def serialize(self):
+        """Return a JSON compatible dictionary."""
```
zzzeid: Can you specify more details as to what this dictionary is for, and what the different fields that are being renamed or added are for? (e.g. "tree", and "details" respectively)

landoapi/models/landing_job.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+
+    # URL of the repository revisions are to land to.
+    repository_url = db.Column(db.Text(), default="")
+
+    # Identifier for the most descendent commit created by this landing.
+    landed_commit_id = db.Column(db.Text(), default="")
+
+    # JSON list of bug ids tied to the revisions at creation time.
+    bug_ids = db.Column(JSONB, nullable=False)
```
zzzeid: An `ARRAY` of strings type would be more appropriate here.

landoapi/models/landing_job.py
```diff
@@ -0,0 +1,209 @@
...
+    #     }
+    revision_to_diff_id = db.Column(JSONB, nullable=False)
+
+    # JSON array of string revision ids of the form "<int>" (used to match
+    # the string type of revision_to_diff_id keys) listing the order
+    # of the revisions in the request from most ancestral to most
+    # descendant.
+    # e.g.
+    #     ["1001", "1002", "1003"]
+    revision_order = db.Column(JSONB, nullable=False)
```
zzzeid: An `ARRAY` of strings or integers type would be more appropriate here.

zzzeid:
Merged into and replaced by https://github.com/mozilla-conduit/lando-api/pull/102/commits

PR #102, opened 2020-03-18
Integrated transplant functionality (2/2)

zzzeid:
This pull request implements transplants as part of Lando API as a replacement for Autoland-Transplant. The functionality in this revision largely implements feature parity with that in Autoland-Transplant, with some additional improvements. Below is the summary of all the features that are included in this PR:

- Updated repos.py to support different local repos (for local development)
- Temporarily removed app user from Dockerfile to allow app user access to mounted /repos filesystem with docker-compose (fixed/reverted)
- Updated transplant endpoint to support legacy + new landings
- Added Dockerfile-dev for local development (uses flask's built-in dev server)
- Added class-based landing worker to support new landing jobs, with graceful shutdown, tree closure, and deferred jobs support
- Minor refactor to how config defaults are loaded
- Added _lock_table_for helper method
- General clean up, exception handling


landoapi/landings.py
```diff
@@ -0,0 +1,201 @@
...
+                "commits for you automatically. Please manually "
+                "rebase your commits and try again.\n\n"
+                f"{str(exc)}"
+            )
+            job.transition_status("fail", message=message, commit=True, db=db)
+            return True
+        except TreeClosed:
+            # TODO when fetching jobs to run, we should fetch ones that are
+            # "in progress" but transiently failed, though we should put them
+            # at the bottom of the stack.
```
globau: yes!  that's much closer to ensuring we land things in the order they were requested.

landoapi/models/landing_job.py
```diff
@@ -0,0 +1,225 @@
...
+
+        Args:
+            action (str): the action to take, e.g. "land" or "fail"
+            commit (bool): whether to commit the changes to the database or not
+            db (SQLAlchemy.db): the database to commit to
+            **kwargs:
+                Additional arguments required by each action, e.g. `message` or
+                `commit_id`.
+        """
+        # TODO: fix the action keys so that they are not free form strings.
```
globau: _whispers_ another enum

tests/test_hg.py
```diff
@@ -102,3 +63,92 @@ def test_integrated_hgrepo_can_log(hg_clone):
...
+# Diff Start Line 7
+add another file.
+diff --git a/test.txt b/test.txt
+--- a/test.txt
++++ b/test.txt
+@@ -1,1 +1,2 @@
+ TEST
++adding another line
+""".strip()
+
```
globau: there's some tests missing from https://github.com/mozilla-conduit/autoland-transplant/blob/master/testing/tests/test-post-autoland-job-from-patch.t

unicode, falling back to patch, and patching a file that doesn't exist

landoapi/landings.py
zzzeid: I implemented this but forgot to remove the TODO -- jobs with `DEFERRED` status will be fetched and stacked in order. Will remove this TODO!

landoapi/hg.py
```diff
@@ -4,27 +4,101 @@
...
+        self.hg_args = hg_args
+        self.out = out
+        self.err = err
+        super().__init__(msg)
+
+
+class TreeClosed(HgException):
+    """Exception when pushing failed due to a closed tree."""
+
+    SNIPPETS = (b"is CLOSED!",)
```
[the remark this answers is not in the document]
zzzeid: Yes, and agreed! I missed this when I merged the latest commit from the old PR. Will update.

landoapi/hg.py
```diff
@@ -150,6 +204,79 @@ def clean_repo(self, strip_non_public_commits=True):
...
+            # in the local repo.
+            import_cmd = ["import", "--no-commit"]
+            # Apply the patch, with file rename detection (similarity).
+            # Using 95 as the similarity to match automv's default.
+            import_cmd += ["-s", "95"]
+
+            try:
+                self.run_hg(import_cmd + [f_diff.name])
+            except hglib.error.CommandError as exc:
+                # Try again using 'patch' instead of hg's internal patch utility.
```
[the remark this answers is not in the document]
zzzeid: Noted.

landoapi/hg.py
```diff
@@ -150,6 +204,79 @@ def clean_repo(self, strip_non_public_commits=True):
...
+                ["commit"]
+                + ["--date", helper.header("Date")]
+                + ["--user", helper.header("User")]
+                + ["--landing_system", "lando"]
+                + ["--logfile", f_msg.name]
+            )
+
+    def push(self, target, bookmark=None):
+        if bookmark is None:
+            self.run_hg_cmds([["push", "-r", "tip", target]])
```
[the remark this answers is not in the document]
zzzeid: Noted.

tests/test_hg.py
zzzeid: Noted.

landoapi/hg.py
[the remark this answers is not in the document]
zzzeid: This wording is coming in from `hglib.error.CommandError` (and `SNIPPETS` is a way of detecting the hardcoded text in those errors and raising specific exceptions instead.) So we'd have to change the text in VCT which I think is out of scope for this PR. There could be a cleaner way of interpreting these errors though.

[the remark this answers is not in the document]
globau: Ah, I see.

landoapi/hg.py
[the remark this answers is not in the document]
zzzeid: This should be resolved now.

landoapi/hg.py
```diff
@@ -150,6 +204,79 @@ def clean_repo(self, strip_non_public_commits=True):
...
+                # might have partially applied the patch.
+                logger.info("import failed, retrying with 'patch'", exc_info=exc)
+                import_cmd += ["--config", "ui.patch=patch"]
+                self.clean_repo(strip_non_public_commits=False)
+
+                try:
+                    self.run_hg(import_cmd + [f_diff.name])
+                    # When using an external patch util mercurial won't
+                    # automatically handle add/remove/renames.
+                    self.run_hg(["addremove", "-s", "95"])
```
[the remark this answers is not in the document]
zzzeid: Fixed.

landoapi/landings.py
```diff
@@ -0,0 +1,201 @@
...
+from landoapi.models.landing_job import LandingJob, LandingJobStatus
+from landoapi.repos import repo_clone_subsystem
+from landoapi.storage import db
+from landoapi.treestatus import treestatus_subsystem
+
+logger = logging.getLogger(__name__)
+
+# TODO: Ensure there are no other landing workers before processing
+# jobs. A job may be executed twice in parralel if a second worker
+# is accidentally started.
```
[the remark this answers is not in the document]
zzzeid: For the time being, this will be handled by Kubernetes (by means of limiting the number of workers to 1) -- in the future this can be reimplemented to have a more scalable solution (or at least a more robust enforcement of single workers, e.g. by means of a registry or lock system). I think this is out of scope for this PR and can be done in a follow up.

landoapi/landings.py
```diff
@@ -0,0 +1,201 @@
...
+                logger.info(
+                    "Landing job queue empty, sleeping for {} seconds".format(
+                        self.seconds_between_jobs
+                    )
+                )
+                time.sleep(self.seconds_between_jobs)
+                continue
+
+            # TODO: use a context manager here instead of job_processing boolean
+            self.job_processing = True
```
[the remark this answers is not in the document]
zzzeid: Using context manager now.

tests/test_hg.py
zzzeid: Added unicode test, the other two are actually part of the `PATCH_WITH_CONFLICT` test. Note that we're only testing `apply_patch` for all of theses types of tests. We should probably add more unit tests for the rest of the functionality (e.g. `build_patch_for_revision` and `push`, etc). Or some integration tests that do not involve S3, but I think the latter, at least, can be done as a follow up.

globau: >` PATCH_WITH_CONFLICT` only passes through the "falling back to patch" but ultimately still fails and re-raises the exception.

Ah, testing that the patch fallback works is important.  

> Any thoughts on what type of merge conflict could fail hg's default patching but pass with patch?

https://bugzilla.mozilla.org/show_bug.cgi?id=1541181 has an example.

In transplant I took the route of passing in a header which triggers a false `hg import` failure.
This ensures the code path is tested even after hg fixes the particular bug in its internal patch code.
https://github.com/mozilla-conduit/autoland-transplant/blob/master/autoland/transplant.py#L404-L418

landoapi/hg.py
```diff
@@ -150,6 +202,82 @@ def clean_repo(self, strip_non_public_commits=True):
...
+            # --landing_system is provided by the set_landing_system hgext.
+            self.run_hg(
+                ["commit"]
+                + ["--date", patch_helper.header("Date")]
+                + ["--user", patch_helper.header("User")]
+                + ["--landing_system", "lando"]
+                + ["--logfile", f_msg.name]
+            )
+
+            # TODO: do we need to ever add "--encoding utf-8" to th commit command?
```
globau, line 261: Shouldn't be needed - we're setting the default encoding when starting up the hg command server.

landoapi/hg.py
```diff
@@ -175,10 +303,10 @@ def get_remote_head(self):
...
         return cset
 
-    def update_from_upstream(self, remote_rev):
-        # Pull "upstream" and update to remote tip.
+    def update_from_upstream(self, source, remote_rev):
+        # Pull and update to remote tip.
         cmds = [
-            ["pull", "upstream"],
+            ["pull", source],
             # TODO: why there is a -r?
```
globau, line 310: Only the specific revision in question needs to be rebased.

landoapi/treestatus.py
```diff
@@ -28,7 +28,7 @@ def is_open(self, tree):
         try:
             resp = self.get_trees(tree=tree)
         except TreeStatusError as exc:
-            if exc.status_code != 404:
+            if exc.status_code != 404 and exc.status_code != 400:
```
globau, line 31: nit: `if exc.status_code not in (400, 404):`

Only caught my eye because I'm used to using pylint, where this would trigger `consider-using-in (R1714)`.

zzzeid: lol -- I missed this, will fix (here's a deja vu https://github.com/mozilla-conduit/lando-api/pull/97#discussion_r370308524)

tests/test_hg.py
zzzeid: Thanks, I will look into this!

zzzeid: This is fixed now. The test that existed already wasn't actually doing anything -- turns out, in order to reproduce this bug the file has to already be on the repo (and not added via a patch). I updated the test so that the new line + trailing `r` are manually committed, and then the patch to remove the file is applied; also confirmed that the outgoing changes actually contain what is expected.

globau: I don't think the test for falling back to patch should rely on any specific hg bug existing.  It would be too easy to upgrade the version of hg and have this test generate a false positive as the test doesn't actually check if the fallback code path was taken.

zzzeid: Good point - I added a new test that forces this change, as well as modified the "newline bug" test to be skipped if the mercurial version is changed.

PR #105, opened 2020-05-20
Landings: create base repos directory in docker image

zzzeid:
Create base /repos directory for mounting, also fix get_subsystems
method to use subsystem instances for comparison.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,9 @@ RUN cd / && pip install --no-cache /app
 ENV PYTHONPATH /app
 RUN chown -R app:app /app
 
+# Create repos directory for transplanting in landing-worker
+RUN mkdir /repos
+
 # Run as a non-privileged user
 USER app
 
--- a/landoapi/cli.py
+++ b/landoapi/cli.py
@@ -16,14 +16,14 @@ def get_subsystems(exclude=None):
     """Get subsystems from the app, excluding those specified in the given parameter.
 
     Args:
-        exclude (list of str): names of subsystems to exclude
+        exclude (list of Subsystem): Subsystems to exclude.
 
     Returns: list of Subsystem
     """
     from landoapi.app import SUBSYSTEMS
 
     exclusions = exclude or []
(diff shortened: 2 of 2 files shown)


landoapi/cli.py
```diff
@@ -57,7 +57,9 @@ def init():
 @click.argument("celery_arguments", nargs=-1, type=click.UNPROCESSED)
 def worker(celery_arguments):
     """Initialize a Celery worker for this app."""
-    for system in get_subsystems(exclude=["repo_clone_subsystem"]):
+    from landoapi.app import repo_clone_subsystem
```
cgsheeh, line 60: I still think it's backwards to import all the things you *don't* need.

I notice there's a pattern in `cli.py` where we check the subsystems are ready and then call out to some other `landoapi` module. Have you considered moving the `ensure_ready` calls into these other modules? For example in `landing_worker` you could ensure the subsystems are ready in the `LandingWorker()` constructor.

I don't know if this is a big architecture change, so it may not be worth doing. Just seems like an anti-pattern to me. :)

Dockerfile
```diff
@@ -52,6 +52,9 @@ RUN cd / && pip install --no-cache /app
 ENV PYTHONPATH /app
 RUN chown -R app:app /app
 
+# Create repos directory for transplanting in landing-worker
+RUN mkdir /repos
```
cgsheeh, line 56: Do we need `VOLUME /repos` too?

landoapi/cli.py
zzzeid: I agree 100% that this could benefit from a refactor -- in terms of `ready` and `ensure_ready`, and it's worth doing though it is a bit of a bigger change, so I'll have to do it in a different PR.

> I still think it's backwards to import all the things you don't need.

I also agree here. There are some alternatives that would clean this up, let me see what I can add in a different PR to help with this.

Dockerfile
zzzeid: We just need the directory so that it can be mounted here. There is no data to be persisted in the container as is.

PR #107, opened 2020-06-02
landing jobs: add ability to cancel landing jobs (bug 1503000)

zzzeid:
- add REST endpoint (update)
- update landing job actions enum
- add grace period functionality

See https://github.com/mozilla-conduit/lando-ui/pull/108 for UI implementation.
See https://bugzilla.mozilla.org/show_bug.cgi?id=1503000.

### DIFF
--- a/landoapi/api/landing_jobs.py
+++ b/landoapi/api/landing_jobs.py
@@ -17,9 +17,9 @@ logger = logging.getLogger(__name__)
 def put(landing_job_id, data):
     """Update a landing job.
 
-    This method checks whether the logged in user is allowed to modify the landing job
-    that is passed, does some basic validation on the data passed, and updates the
-    landing job instance accordingly.
+    Checks whether the logged in user is allowed to modify the landing job that is
+    passed, does some basic validation on the data passed, and updates the landing job
+    instance accordingly.
 
     Args:
         landing_job_id (int): The unique ID of the LandingJob object.
@@ -60,14 +60,15 @@ def put(landing_job_id, data):
             type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
         )
 
-    if landing_job.status == LandingJobStatus.SUBMITTED:
-        landing_job.transition_status(LandingJobAction.CANCEL)
-        db.session.commit()
-        return {"id": landing_job.id}, 200
-    else:
-        raise ProblemException(
-            400,
-            "Landing job could not be cancelled.",
-            f"Landing job status ({landing_job.status}) does not allow cancelling.",
-            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
-        )
+    if data["action"] == LandingJobAction.CANCEL.value:
+        if landing_job.status == LandingJobStatus.SUBMITTED:
+            landing_job.transition_status(LandingJobAction.CANCEL)
+            db.session.commit()
+            return {"id": landing_job.id}, 200
+        else:
+            raise ProblemException(
+                400,
+                "Landing job could not be cancelled.",
+                f"Landing job status ({landing_job.status}) does not allow cancelling.",
+                type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+            )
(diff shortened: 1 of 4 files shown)


landoapi/api/landing_jobs.py
```diff
@@ -0,0 +1,73 @@
...
+logger = logging.getLogger(__name__)
+
+
+@auth.require_auth0(scopes=("lando", "profile", "email"), userinfo=True)
+def put(landing_job_id, data):
+    """Update a landing job.
+
+    This method checks whether the logged in user is allowed to modify the landing job
+    that is passed, does some basic validation on the data passed, and updates the
+    landing job instance accordingly.
```
zzzeid, line 22:
```suggestion
    Checks whether the logged in user is allowed to modify the landing job that is
    passed, does some basic validation on the data passed, and updates the landing job
    instance accordingly.
```

landoapi/models/landing_job.py
```diff
@@ -13,6 +13,8 @@
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_GRACE_SECONDS = 1  # 60 * 15
```
zzzeid, line 16:
```suggestion
DEFAULT_GRACE_SECONDS = 60 * 2
```

landoapi/api/landing_jobs.py
```diff
@@ -0,0 +1,73 @@
...
+        landing_job.transition_status(LandingJobAction.CANCEL)
+        db.session.commit()
+        return {"id": landing_job.id}, 200
+    else:
+        raise ProblemException(
+            400,
+            "Landing job could not be cancelled.",
+            f"Landing job status ({landing_job.status}) does not allow cancelling.",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
```
zzzeid, line 73:
```suggestion
    if data["action"] == LandingJobAction.CANCEL.value:
        if landing_job.status == LandingJobStatus.SUBMITTED:
            landing_job.transition_status(LandingJobAction.CANCEL)
            db.session.commit()
            return {"id": landing_job.id}, 200
        else:
            raise ProblemException(
                400,
                "Landing job could not be cancelled.",
                f"Landing job status ({landing_job.status}) does not allow cancelling.",
                type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
            )```

**Comment by zzzeid** [comment]
Closing, replaced by https://github.com/mozilla-conduit/lando-api/pull/116.

**Comment by zzzeid** [comment]
> Should we have a test for cancel job via REST API?

Good call, will add that in.

**Comment by zzzeid** [comment]
> The grace period essentially means each landing triggered through Lando will wait 2 minutes before having the job started, right? Or the job can start, but the landing won't happen until 2 minutes has passed? I just want to make sure the grace period won't compound between jobs and cause landings to take a considerable amount of extra time.

The landing worker will only pick up jobs that have been submitted before the grace period -- until then, the jobs will remain in a `SUBMITTED` state (their initial state). Once a job is picked up by the landing worker, it can not be stopped. Before it is picked up, it is as though it was not created and can be deleted/cancelled without issue.

Otherwise everything behaves the same. In other words, instead of picking up the jobs that were submitted 0 seconds ago and earlier (no grace period), it is picking up jobs that were submitted 120 seconds ago and earlier.


> Maybe some explanation of the grace period functionality in the bug/commit message/here would be nice to have.

Will do.

**Comment by zzzeid** [comment]
To clarify the need for this feature, it is to add a minimum buffer period. Even though some jobs may not be picked up anyway by the landing worker for 120 seconds or more (e.g. if it is busy), it's possible that a landing will go through immediately if the landing worker is free. If a job is submitted 120 seconds before the landing worker picks it up, and the landing worker was busy for 200 seconds with another job, the new job will be picked up right away once the landing worker is able to.

File: landoapi/api/landing_jobs.py [hunk 4]
```diff
@@ -0,0 +1,72 @@
...
+    ldap_username = g.auth0_user.email
+    if landing_job.requester_email != ldap_username:
+        raise ProblemException(
+            403,
+            "Unauthorized",
+            f"User not authorized to update landing job {landing_job_id}",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403",
+        )
+
+    if data["status"] == LandingJobStatus.CANCELLED.value:
```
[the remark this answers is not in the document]
zzzeid: Done.

PR #108, opened 2020-06-04
landing worker: add support for ssh

zzzeid:
- Load ssh keys from the environment variable
- Setup ssh agent for use in container

### DIFF
--- a/landoapi/landing_worker.py
+++ b/landoapi/landing_worker.py
@@ -44,12 +44,11 @@ def job_processing(worker):
 
 class LandingWorker:
     def __init__(self, sleep_seconds=5):
+        SSH_PRIVATE_KEY_ENV_KEY = "SSH_PRIVATE_KEY"
+
         self.sleep_seconds = sleep_seconds
         config_keys = ["AWS_SECRET_KEY", "AWS_ACCESS_KEY", "PATCH_BUCKET_NAME"]
         self.config = {k: current_app.config[k] for k in config_keys}
-
-        # Create a list to keep track of any environment variables being manipulated.
-        self.config["SSH_ENV_KEYS"] = []
 
         # The list of all repos that are enabled for this worker
         self.applicable_repos = (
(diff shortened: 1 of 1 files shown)


landoapi/landing_worker.py
```diff
@@ -69,6 +72,53 @@ def __init__(self, sleep_seconds=5):
...
+        # This pattern will match keys and values, and ignore everything after the
+        # semicolon. For example, the output of `agent_process` is of the form:
+        #     SSH_AUTH_SOCK=/tmp/ssh-c850kLXXOS5e/agent.120801; export SSH_AUTH_SOCK;
+        #     SSH_AGENT_PID=120802; export SSH_AGENT_PID;
+        #     echo Agent pid 120802;
+
+        pattern = re.compile("(.+)=([^;]*)")
+        for key, value in pattern.findall(agent_process.stdout):
+            self.config["SSH_ENV_KEYS"].append(key)
+            os.environ[key] = value
```
cgsheeh, line 102: Nit: It seems like we really want to find the values of two specific variables here, `SSH_AUTH_SOCK` and `SSH_AGENT_PID`. Maybe we should be more explicit about finding the values of those keys instead of using a funky regex. This solution seems to work fine, so it's not a huge problem.

landoapi/landing_worker.py
```diff
@@ -69,6 +72,53 @@ def __init__(self, sleep_seconds=5):
...
+        )
+        if add_process.returncode != 0:
+            raise Exception(add_process.stderr)
+        logger.info("Added private SSH key from environment.")
+
+    def _teardown_ssh(self):
+        """Clean up relevant environment variables we've set up for ssh-agent.
+        """
+        for key in self.config["SSH_ENV_KEYS"]:
+            del os.environ[key]
```
cgsheeh, line 120: Can you explain why we need this?

landoapi/landing_worker.py
```diff
@@ -5,6 +5,8 @@
 
 import logging
 import os
+import subprocess
+import re
```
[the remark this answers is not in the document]
zzzeid: Will fix.

landoapi/landing_worker.py
```diff
@@ -43,10 +45,11 @@ def job_processing(worker):
 class LandingWorker:
```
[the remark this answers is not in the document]
zzzeid: What would be the purpose of explicitly extending `object`? This project uses Python 3.8. If you look at the docs (e.g. [here](https://docs.python.org/3/tutorial/classes.html#class-objects)) all the examples omit the explicit `object` reference. As far as I know this was a Python 2 compatibility relic, or possibly needed in some edge cases but I haven't come across one yet.

landoapi/landing_worker.py
```diff
@@ -69,6 +72,53 @@ def __init__(self, sleep_seconds=5):
...
+        """
+        SSH_PRIVATE_KEY_ENV_KEY = "SSH_PRIVATE_KEY"
+
+        # Fetch ssh private key from the environment. Note that this key should be
+        # stored in standard format including all new lines and new line at the end
+        # of the file.
+        ssh_private_key = os.environ.get(SSH_PRIVATE_KEY_ENV_KEY)
+        if not ssh_private_key:
+            logger.warning(f"No {SSH_PRIVATE_KEY_ENV_KEY} present in environment.")
+            return
```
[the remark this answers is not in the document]
zzzeid: The worker does not care about the protocol. If provided an HTTPS path, even on production, it will still try to push (though obviously we don't have any repos like this right now).

Also, when running the worker locally or on the dev environments, not having an SSH key won't necessarily result in an error.

Thoughts?

landoapi/landing_worker.py
zzzeid: I pondered this, but ended up using regex because the nature of `ssh-agent -s` is to generate both the keys and values that we need, and I figured this is future-proof, in case those keys change.

The regex also helps us extract those keys and values more easily. I could add explicit checks for those keys, but parsing the output manually without the regex is not necessarily cleaner. For example, we would have to handle the `echo` statement at the end, new lines, and doing splits to ignore text after the semicolon for the relevant lines.

For example, for the first two lines we'd have to do `key, value = line.split(";")[0].split("=")` which is not any cleaner than `re.compile("(.+)=([^;]*)")`. And we would have to do this only for lines 1 and 2 (i.e. we would have to either do `lines = agent_process.stdout.split("\n")[0:2]`, or check for presence of `=` or `;` on a line, neither which are pretty.)

Thoughts?

landoapi/landing_worker.py
zzzeid: Good point. I'm not really sure we need this, it doesn't really do anything. If the worker is killed or dies, then these variables will be reset anyway on the next start. Will remove.

landoapi/landing_worker.py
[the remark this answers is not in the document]
cgsheeh: Ah you're right, the semantics are the same in Python 3.

landoapi/landing_worker.py
cgsheeh: Okay, I think you're right about future proofing and having this be cleaner that `split` + index + `split`, etc. I still think my concern about missing variables is warranted though. Could we log which variables this loop finds and adds to the environment? Then if we do hit that problem (`ssh-agent` changes or something is wonky with the regex) we can tell from the logs.

I also noticed that we create the `self.config["SSH_ENV_KEYS"]` list, but don't seem to do anything with it.

zzzeid: Forgot to remove `self.config["SSH_ENV_KEYS"]` when I removed the tear down code, will do.

Agreed RE: env keys, will add proper logs to show which keys/values have been set.

PR #110, opened 2020-06-18
landing worker: add support for fetching ssh params from environment

zzzeid:


landoapi/repos.py
```diff
@@ -113,13 +127,11 @@
         "third-repo": Repo(
             tree="third-repo",
             access_group=SCM_LEVEL_1,
-            push_path="ssh://autoland.hg-ssh//repos/third-repo",
+            push_path="ssh://autoland.hg:8022//repos/third-repo",
```
ckolos, line 130: This should be a var

PR #114, opened 2020-07-06
landing worker: send notification of failed transplants (bug 1649465)

zzzeid:
Add support for email notification when a landing job fails.

### DIFF
--- a/tests/test_landings.py
+++ b/tests/test_landings.py
@@ -2,12 +2,32 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import mock
+import pytest
+
 from landoapi import patches
 from landoapi.hg import HgRepo
 from landoapi.landing_worker import LandingWorker
 from landoapi.models.landing_job import LandingJob, LandingJobStatus
 from landoapi.models.transplant import Transplant, TransplantStatus
 from landoapi.repos import Repo, SCM_LEVEL_3
+
+
+@pytest.fixture
+def upload_patch():
+    """A fixture that fake uploads a patch"""
+
+    def _upload_patch(number):
+        patches.upload(
+            number,
+            number,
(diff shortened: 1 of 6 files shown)


tests/test_landings.py
```diff
@@ -185,3 +191,51 @@ def test_integrated_execute_job(
...
+    db,
+    s3,
+    mock_repo_config,
+    hg_server,
+    hg_clone,
+    treestatusdouble,
+    monkeypatch,
+    upload_patch,
+):
+    """Ensure that a failed landings triggers a user notification.
```
zzzeid, line 207:
```suggestion
    """Ensure that a failed landing triggers a user notification.
```

tests/test_landings.py
```diff
@@ -10,6 +13,23 @@
 from landoapi.repos import Repo, SCM_LEVEL_3
 
 
+@pytest.fixture
+def upload_patch():
+    """A fixture that fake uploads a patch"""
```
zzzeid, line 18:
```suggestion
    """A fixture that fake uploads a patch."""
```

tests/test_notifications.py
```diff
@@ -95,7 +95,13 @@ def test_notify_user_of_landing_failure(check_celery, app, celery_worker, smtp):
...
     # celery_worker fixture causes the test to cross threads.  We only ensure the
     # happy-path runs cleanly.
-    notify_user_of_landing_failure(Transplant(revision_order=["1"]))
+    transplant = Transplant(revision_order=["1"])
+    notify_user_of_landing_failure(
+        transplant.requester_email,
+        transplant.head_revision,
+        transplant.error,
+        transplant.request_id,
+    )
```
zalun, line 104: Is creating the `Transplant` object is required?

zzzeid: It is not required at all, indeed. I will update this in a follow up PR. This test needs to be rewritten though, as it's not really testing or doing much, as per the original comment above.

PR #120, opened 2020-07-21
stacks: add support for DONTBUILD flag (bug 1642939)

zzzeid:
The API should accept a `flags` parameter when submitting a transplant, and allowed flags should be appended to commit messages.


zzzeid:
Closing this PR and re-opening from a non-fork.

PR #126, opened 2020-07-28
transplants: remove check-in needed flag with new landings (bug 1655833)

zzzeid:


landoapi/api/transplants.py
```diff
@@ -367,77 +368,76 @@ def post(data):
 
         db.session.commit()
         logger.info("New landing job {job.id} created for {landing_repo.tree} repo")
+        job_id = job.id
```
cgsheeh, line 371: So basically instead of returning here we set the job ID, put the transplanting code behind the `else` clause, and then we always hit the code path which removes the checkin-needed flag.

zzzeid: Yes correct. It's a little cleaner than copying/pasting. The `else` statement will be removed in https://bugzilla.mozilla.org/show_bug.cgi?id=1654646.

PR #127, opened 2020-07-29
landing worker: defer jobs when push race is lost (bug 1655826)

zzzeid:


landoapi/hg.py
```diff
@@ -281,6 +287,16 @@ def apply_patch(self, patch_io_buf):
...
+        # For testing, force a LostPushRace exception if this header is
+        # defined.
+        if (
+            self.patch_header
+            and self.patch_header("Fail HG Import") == b"LOSE_PUSH_RACE"
+        ):
+            exc = hglib.error.CommandError(
+                (), 1, b"", b"forced fail: abort: push creates new remote head"
+            )
+            raise HgException.from_hglib_error(exc)
```
zalun, line 299: Why we just don't raise the `LostPushRace` directly?

zzzeid: This will also test the `HgException.from_hglib_error` code flow for this exception as well to ensure that we're raising the right exception given this error message. I can move that code flow to a separate test.

cgsheeh:
Oops, clicked approve twice. :)

zzzeid:
> Oops, clicked approve twice. :)

All good.

PR #130, opened 2020-07-30
stacks: add support for DONTBUILD flag (bug 1642939)

zzzeid:
The API should accept a flags parameter when submitting a transplant, and allowed flags should be appended to commit messages.

### DIFF
--- a/landoapi/api/stacks.py
+++ b/landoapi/api/stacks.py
@@ -159,26 +159,20 @@ def get(revision_id):
         short_name = PhabricatorClient.expect(
             stack_data.repositories[phid], "fields", "shortName"
         )
+
         repo = supported_repos.get(short_name)
-        if repo is None:
-            landing_supported, approval_required = False, None
-        else:
-            landing_supported, approval_required = True, repo.approval_required
-        url = (
-            "{phabricator_url}/source/{short_name}".format(
-                phabricator_url=current_app.config["PHABRICATOR_URL"],
-                short_name=short_name,
-            )
-            if not landing_supported
-            else supported_repos[short_name].url
-        )
+        landing_supported = repo is not None
+
         repositories.append(
             {
-                "phid": phid,
-                "short_name": short_name,
-                "url": url,
                 "landing_supported": landing_supported,
-                "approval_required": approval_required,
+                "approval_required": landing_supported and repo.approval_required,
+                "url": repo.url
+                if landing_supported
+                else (f"{current_app.config['PHABRICATOR_URL']}/source/{short_name}"),
+                "phid": phid,
+                "shortname": short_name,
+                "commit_flags": repo.commit_flags if repo else [],
             }
         )
 
--- a/landoapi/api/stacks.py
+++ b/landoapi/api/stacks.py
@@ -162,14 +162,17 @@ def get(revision_id):
 
         repo = supported_repos.get(short_name)
         landing_supported = repo is not None
+        url = (
+            repo.url
+            if landing_supported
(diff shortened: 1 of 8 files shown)


landoapi/commit_message.py
```diff
@@ -65,35 +65,36 @@
...
-        bug: The bug number to use or None.
-        reviewers: A list of reviewer usernames.
-        summary: A string containing the revision's summary
-        revision_url: The revision's url in Phabricator
+        title (str): The first line of the original commit message.
+        bug (str): The bug number to use or None.
+        reviewers (list of str): A list of reviewer usernames.
+        summary (str): A string containing the revision's summary.
+        revision_url (str): The revision's url in Phabricator.
+        flags (list of str): A list of flags to append to the title.
```
cgsheeh, line 85: Maybe these can be type annotations?

landoapi/api/stacks.py
```diff
@@ -159,26 +159,23 @@ def get(revision_id):
...
             {
-                "phid": phid,
-                "short_name": short_name,
-                "url": url,
                 "landing_supported": landing_supported,
-                "approval_required": approval_required,
+                "approval_required": landing_supported and repo.approval_required,
+                "url": url,
+                "phid": phid,
+                "shortname": short_name,
```
cgsheeh, line 177: Is the change to `shortname` from `short_name` expected? Seems to break the convention.

landoapi/api/transplants.py
```diff
@@ -251,6 +271,14 @@ def post(data):
             f"{to_land}, {landing_repo}, {stack_data}"
         )
 
+    if set(flags) - set(landing_repo.commit_flags):
+        raise ProblemException(
+            400,
+            "Invalid flags specified",
+            f"Flags must be one or more of {landing_repo.commit_flags}; {flags} provided.",
```
cgsheeh, line 278: Nit: if we save the result of `set(flags) - set(landing_repo.commit_flags)` we could explicitly return the bad flag instead of returning all the passed flags.

landoapi/spec/swagger.yml
```diff
@@ -88,6 +88,10 @@ paths:
                       with the current warnings and the matching token. If
                       the warnings have changed between requesting a dryrun
                       and requesting a landing, the landing will fail.
+                  flags:
```
cgsheeh, line 91: Missing a description of this parameter.

landoapi/api/stacks.py
zzzeid: No, this was not expected. Nice catch! Doesn't seem to be used anywhere interestingly enough.

landoapi/api/transplants.py
zzzeid: Done.

landoapi/commit_message.py
zzzeid: Done.

landoapi/spec/swagger.yml
zzzeid: Done.

landoapi/commit_message.py
```diff
@@ -65,12 +65,19 @@
...
 
-def format_commit_message(title, bug, reviewers, summary, revision_url):
+def format_commit_message(
+    title: str,
+    bug: str,
+    reviewers: List[str],
+    summary: str,
+    revision_url: str,
+    flags: List[str] = None,
+) -> str:
```
cgsheeh, line 75: `Tuple[str]`

zzzeid: oops, 🤦. Fixed & updated docstring.

PR #135, opened 2020-09-21
transplants: add support for testing policy warnings (bug 1666309)

zzzeid:
Adds support for testing policy tags by:
- Checking if a repository has a `testing-policy` project associated with it, thus implying that said repo should enforce testing policy tags on all revisions
- Checking if a particular revision has a testing policy tag, and if not, adds a warning to the revision

### DIFF
--- a/landoapi/projects.py
+++ b/landoapi/projects.py
@@ -3,8 +3,10 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import logging
 
+from typing import Optional, List
+
 from landoapi.cache import cache
-from landoapi.phabricator import result_list_to_phid_dict
+from landoapi.phabricator import result_list_to_phid_dict, PhabricatorClient
 
 logger = logging.getLogger(__name__)
 
@@ -89,27 +91,15 @@ def get_project_phid(project_slug, phabricator, allow_empty_result=True):
 
 
 @cache.cached(key_prefix=SEC_PROJ_CACHE_KEY, timeout=DEFAULT_CACHE_KEY_TIMEOUT)
-def get_secure_project_phid(phabricator):
+def get_secure_project_phid(phabricator: PhabricatorClient) -> Optional[str]:
     """Return a phid for the project indicating revision security.
-
-    Args:
-        phabricator: A PhabricatorClient instance.
-
-    Returns:
-        A string phid if the project is found, otherwise None.
     """
     return get_project_phid(SEC_PROJ_SLUG, phabricator)
 
 
 @cache.cached(key_prefix=CHECKIN_PROJ_CACHE_KEY, timeout=DEFAULT_CACHE_KEY_TIMEOUT)
-def get_checkin_project_phid(phabricator):
+def get_checkin_project_phid(phabricator: PhabricatorClient) -> Optional[str]:
     """Return a phid for the project indicating check-in is needed.
-
-    Args:
-        phabricator: A PhabricatorClient instance.
-
-    Returns:
-        A string phid if the project is found, otherwise None.
     """
     return get_project_phid(CHECKIN_PROJ_SLUG, phabricator)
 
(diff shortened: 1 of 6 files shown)


landoapi/projects.py
```diff
@@ -14,6 +14,13 @@
 SEC_PROJ_CACHE_KEY = "secure-project-phid"
 CHECKIN_PROJ_SLUG = "check-in_needed"
 CHECKIN_PROJ_CACHE_KEY = "checkin-project-phid"
+
+# needs-testing-tag is used to mark those revisions that are approved but are
+# missing a testing tag. This is used to trigger a warning when attempting to
+# land something.
+NEEDSTESTINGTAG_PROJ_SLUG = "needs-testing-tag"
```
bgrins, line 21: If we check on needs-testing-tag this will also rely on the Herald rules being set up. I'm not necessarily opposed to that, but something that might be more straightforward is checking for the _lack of_ one of the testing tags (`testing-approved`, `testing-exception-elsewhere`, `testing-exception-other`, `testing-exception-ui`, `testing-exception-unchanged`).

zzzeid: Was about to reach out RE: preference. Either works, and I do prefer the approach of checking for the lack of any of the tags (i.e. what you suggested), this was easier to test while I was coding things. Is the list of testing tags finalized? (note that changing any of the tags means we will need up update the hardcoded values).

bgrins: The list of testing tags is stable. I don't think "finalized" is the right word, since I could definitely see adding/removing/renaming tags as needs evolve. But the bar to change is going to be fairly high and if it does, we'll have other tooling to update as well so I'm fine with documenting this and landing with the hardcoded list.

landoapi/projects.py
```diff
@@ -14,6 +14,22 @@
...
+
+# Testing tag slugs. Revisions need one of these tags to remove the respective warnings.
+TESTING_TAG_PROJ_SLUGS = (
+    "testing-approved",
+    "testing-exception-elsewhere",
+    "testing-exception-other",
+    "testing-exception-ui",
+    "testing-exception-unchanged",
+)
+
```
zzzeid, line 26:
```suggestion

```

tests/test_revisions.py
```diff
@@ -3,14 +3,17 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import pytest
+from unittest.mock import MagicMock
 
 from landoapi.phabricator import RevisionStatus, ReviewerStatus
 from landoapi.repos import get_repos_for_env
+
```
zzzeid, line 10:
```suggestion

```

landoapi/projects.py
```diff
@@ -98,6 +114,35 @@ def get_checkin_project_phid(phabricator):
     return get_project_phid(CHECKIN_PROJ_SLUG, phabricator)
 
 
+@cache.cached(
+    key_prefix=TESTING_POLICY_PROJ_CACHE_KEY, timeout=DEFAULT_CACHE_KEY_TIMEOUT
+)
+def get_testing_policy_phid(phabricator):
+    """Return a phid for the project indicating testing policy.
+
+    Args:
```
cgsheeh, line 123: These `Args`/`Returns` comments should be type annotations.

landoapi/revisions.py
```diff
@@ -181,6 +183,53 @@ def revision_is_secure(revision, secure_project_phid):
...
+            associated with a repo.
+
+    Returns:
+        bool: True if the revision needs a testing policy tag, else False.
+    """
+    repo_phid = revision["fields"]["repositoryPHID"]
+    repo = get_phabricator_repo(repo_phid)
+
+    # Check if the repo has a testing-policy tag.
+    if testing_policy_phid in repo["attachments"]["projects"]["projectPHIDs"]:
```
cgsheeh, line 226: Is it possible for these fields to be absent? Will raise a `KeyError` if they are ever missing.

cgsheeh (changes requested):
Just needs type annotations and a quick question re: guarantees from the Phab APIs, otherwise it looks good.

landoapi/projects.py
zzzeid: Fixed.

cgsheeh (approved):
You missed annotations on some of your other changes, but it seems we don't have annotations enforced/actually completed globally, so I won't bother with an `r-`. Sorry if I'm being nit-picky. :)

zzzeid:
> You missed annotations on some of your other changes, but it seems we don't have annotations enforced/actually completed globally, so I won't bother with an `r-`. Sorry if I'm being nit-picky. :)

Ah -- indeed we do not enforce it, but I add argument descriptions and type hinting in the `Args` section -- which is practically equivalent since we don't do anything with these type hints programatically. Most of the code does not have type hinting, so I just changed the one that you asked for and all the ones similar to it in that file.

Good to be nit-picky, so apology not accepted!

PR #139, opened 2020-10-29
requirements: add requirements.in file (bug 1674127)

zzzeid:


cgsheeh (approved):

cgsheeh:
Is there some motivation for this or just picking off some low hanging fruit? :)

zzzeid:
> Is there some motivation for this or just picking off some low hanging fruit? :)

I'd rather do this before we update the cryptography version in place (i.e. in #138) to keep things more consistent (Lando UI already uses a `requirements.in`)

zzzeid:
This is a minimal change to support `requirements.in`. We should upgrade `pytest` and `pytest-flask` and this would resolve the issue with `setuptools`.

PR #140, opened 2020-10-30
repos: update repo config to reflect new repos (bug 1673626)

zzzeid:


landoapi/repos.py
```diff
@@ -179,41 +179,21 @@ def __post_init__(self):
...
+            access_group=SCM_LEVEL_1,
         ),
         "m-c": Repo(
             tree="m-c",
-            url="https://autolandhg.devsvcdev.mozaws.net/local/m-c/",
-            access_group=SCM_VERSIONCONTROL,
-            push_path="ssh://autolandhg.devsvcdev.mozaws.net//local_repos/m-c",
-            pull_path="https://hg.mozilla.org/integration/autoland",
+            url="https://hg.mozilla.org/conduit-test/m-c",
+            access_group=SCM_LEVEL_1,
```
cgsheeh, line 190: Can we make this `SCM_VERSIONCONTROL` and add an `@` bookmark to more closely mirror hgmo? Shouldn't make a huge difference but worth mirroring since it's easy.

cgsheeh (changes requested):

cgsheeh (approved):

PR #141, opened 2020-10-30
Bug 1548114: implement autoformatting in Lando

cgsheeh:
Implements autoformatting in Lando via `hg fix`.

Updates Mercurial to 5.5.1 and adds a few minor changes for the update.
We need to update Mercurial to a newer version to get a minor bugfix in
`hg fix`. We remove the `-r` flag from the call to `rebase --abort`
since the flags are mutually exclusive on newer versions of Mercurial.


landoapi/hg.py
```diff
@@ -345,7 +365,7 @@ def update_from_upstream(self, source, remote_rev):
         # Pull and update to remote tip.
         cmds = [
             ["pull", source],
-            ["rebase", "--abort", "-r", remote_rev],
+            ["rebase", "--abort"],
```
cgsheeh, line 368: Can't use `--abort` and `-r` at the same time in newer versions of `hg`.

landoapi/landing_worker.py
```diff
@@ -171,9 +172,20 @@ def start(self):
...
+
+                # Set up config override
+                config_override = {}
+                if repo.config_override:
+                    config_override.update(repo.config_override)
+
+                # Set `hg fix` config if `formatting` kwarg is set
+                if repo.formatting:
+                    for k, v in repo.formatting.items():
+                        config_override[f"fix.{k}"] = v
```
cgsheeh, line 184: This is a little ugly and I'm sure there's a better way.

landoapi/repos.py
```diff
@@ -215,6 +219,17 @@ def __post_init__(self):
...
         ),
+        # A repo to test autoformatting
+        "formatting-repo": Repo(
+            tree="autoformatting",
+            url="https://autolandhg.devsvcdev.mozaws.net/local/autoformatting/",
+            access_group=SCM_VERSIONCONTROL,
+            pull_path="https://autolandhg.devsvcdev.mozaws.net/local/autoformatting/",
+            formatting={
+                "black:command": "black -t py36 -- -",
+                "black:pattern": "set:**.py",
```
cgsheeh, line 230: Just an example of how the config would look in production.

tests/conftest.py
```diff
@@ -382,7 +382,8 @@ def hg_server(hg_test_bundle, tmpdir):
     hg_url = "http://localhost:" + port
 
     repo_dir = tmpdir.mkdir("hg_server")
-    subprocess.run(["hg", "clone", hg_test_bundle, repo_dir], check=True)
+    subprocess.run(["hg", "clone", hg_test_bundle, repo_dir], 
+                   check=True, cwd="/")
```
cgsheeh, line 386: Without this I get an error on my local machine due to an extension I have installed. So setting to `/` will put the cwd outside of the mounted docker volume and avoid that problem, and since it's a clone the `cwd` doesn't really matter.

landoapi/hg.py
```diff
@@ -109,6 +114,10 @@ class HgRepo:
         "extensions.strip": "",
         "extensions.rebase": "",
         "extensions.set_landing_system": "/app/hgext/set_landing_system.py",
+
+        # Turn on fix extension for autoformatting, set to abort on failure
+        "extensions.fix": "",
+        "fix.failure": "abort",
```
zzzeid, line 120: We should probably put this in `config_overrides` per-repo, since most repos will not have this extension enabled.

landoapi/landing_worker.py
zzzeid: I think it would be a bit cleaner to merge this config within `HgRepo.__init__` rather than doing it here.

landoapi/repos.py
zzzeid, line 230: I think these would be cleaner placed as normal config overrides, as
```suggestion
            formatting={
                "fix.black:command": "black -t py36 -- -",
                "fix.black:pattern": "set:**.py",
```

We would also include:
```
config_overrides = {
    "extensions.fix": "",
    "fix.failure": "abort",
},
```

And merge them within `HgRepo` into the same eventual config dict.

landoapi/landing_worker.py
```diff
@@ -236,6 +248,10 @@ def run_job(self, job, repo, hgrepo, treestatus, patch_bucket):
                     )
                     hgrepo.apply_patch(patch_buf)
 
+                # Run `hg fix` configured formatters if enabled
+                if repo.formatting:
+                    hgrepo.format()
```
zzzeid, line 253: I think it would be helpful here to set a `job.is_autoformatted` flag so we can use it to indicate this in the UI. We could also make it more general as a list of flags.

zzzeid (commented):
A few initial comments.

landoapi/hg.py
cgsheeh: Enabling this extension doesn't do anything except turn on the `hg fix` command, which is the same behaviour as enabling `purge`/`rebase`/`strip`, and we enable those globally. I think there are more reasons to turn this on globally than to leave it off (violates DRY, prone to error if someone forgets to set error to abort, makes formatting configs more verbose than they need to be, etc).

landoapi/repos.py
cgsheeh: > I think it would be a bit cleaner to merge this config within HgRepo.__init__ rather than doing it here.

Thought of that but since we don't have a `repo` object when calling the constructor, we can't set up the config override there. :/

landoapi/landing_worker.py
cgsheeh: > I think it would be helpful here to set a `job.is_autoformatted` flag so we can use it to indicate this in the UI. We could also make it more general as a list of flags.

Right now this code just returns if the autoformatting doesn't fail. We'll need to pass information indicating if the changesets were actually changed by the formatters. I'll add that in the next iteration.

Would adding `job.is_autoformatted` require a change to the DB schema?

zzzeid: > Right now this code just returns if the autoformatting doesn't fail. We'll need to pass information indicating if the changesets were actually changed by the formatters. I'll add that in the next iteration.

Sounds good

> Would adding `job.is_autoformatted` require a change to the DB schema?

Yes -- found in `landoapi.models.landing_job`.

hgext/postfix_hook.py
```diff
@@ -0,0 +1,35 @@
...
+    if wdirwritten:
+        ui.warn(b"Working directory was written; this should not happen\n")
+        return FAIL
+
+    #if replacements:
+    #    for k, v in replacements.items():
+    #        ui.write(b"key=%s, val=%s\n" % 
+    #                 (pycompat.bytestr(k.decode('utf-8', 'surrogatepass')), 
+    #                  pycompat.bytestr(v.decode('utf-8', 'surrogatepass'))))
+
```
cgsheeh, line 33: My plan is to put information about re-written commits here and parse them from the output of `run_hg(["fix...`

hgext/set_landing_system.py
```diff
@@ -3,7 +3,7 @@
```
cgsheeh, line 3: Wondering if it makes sense to expand this `hgext` to be `landoext` and put the hooks in there? Not sure we'll add any more Mercurial customizations or not.

landoapi/landing_worker.py
cgsheeh: Still need to address this.

landoapi/repos.py
```diff
@@ -56,6 +56,9 @@ class Repo:
         config_override (dict): Parameters to override when loading the Mercurial
             configuration. The keys and values map directly to configuration keys and
             values. Defaults to `None`.
+        formatting (dict): `hg fix` config for automatic formatters on the repo. Only
+            the `[fix]` section values should be passed here. Omitting this parameter
+            assumes formatting is not enabled for a repo. Defaults to `None`.
```
cgsheeh, line 61: Forgot to remove this. :)

tests/test_landings.py
```diff
@@ -181,6 +181,56 @@ def _create_transplant(
...
+ 
+ adding another line
++add one more line
+""".strip()
+
+TESTTXT_FORMATTED_1 = b"""
+TeSt
+
+
+aDdInG AnOtHeR LiNe
```
cgsheeh, line 223: I implemented this in an attempt to show `hg fix`'s behaviour of not touching unrelated changes in a patch. Turns out that behaviour is behind the `linerange` config option. After looking around it turns out that feature is supported by clang-format and prettier, but not by black. So black changes will always modify the entire file, including unrelated changes to the patch. This might end up being fine but worth pointing out now.

landoapi/hg.py
```diff
@@ -286,6 +297,28 @@ def apply_patch(self, patch_io_buf):
...
+            # Update the working directory to the latest change
+            logger.info("updating to most recent revision")
+            self.run_hg(["update", "-C", "-r", "tip"])
+
+            for line in fix_output:
+                if not line.startswith(b"replacements"):
+                    continue
+
+                logger.info("replacement revisions found")
+                return json.loads(line[len(b"replacements::\t") :].decode("utf-8"))
```
zzzeid, line 315: Looks like we are trying to extract exactly one line from `fix_output`. This section of the code could be refactored to improve readability.

Also, `len(b"replacements::\t")` could be replaced by an actual integer (plus a comment), or put into a variable, to improve readability.

landoapi/repos.py
```diff
@@ -88,6 +88,14 @@ def __post_init__(self):
         if not self.short_name:
             self.short_name = self.tree
 
+    @property
+    def formatting(self) -> bool:
+        """Return `True` if formatting is enabled for the repo."""
+        if not self.config_override:
+            return False
+
+        return any(config.startswith("fix") for config in self.config_override.keys())
```
zzzeid, line 97:
```suggestion
        return self.config_override and any(
            config.startswith("fix") for config in self.config_override.keys())
```

tests/test_landings.py
```diff
@@ -290,3 +340,143 @@ def test_failed_landing_job_notification(
...
+    job = LandingJob(
+        status=LandingJobStatus.IN_PROGRESS,
+        requester_email="test@example.com",
+        repository_name="mozilla-central",
+        revision_to_diff_id={"1": 1, "2": 2},
+        revision_order=["1", "2"],
+        attempts=1,
+    )
+
+    worker = LandingWorker(sleep_seconds=0.01)
```
zzzeid, line 478: See `test_failed_landing_job_notification` for a workaround, in order to prevent a notification from triggering. `        "landoapi.landing_worker.notify_user_of_landing_failure` would need to be mocked. That or we would have to refactor `test_notifications`.

zzzeid (commented):
Few minor comments on first pass.

landoapi/hg.py
```diff
@@ -91,6 +99,21 @@ class PatchConflict(PatchApplicationFailure):
     )
 
 
+def check_fix_output_for_replacements(fix_output: List[bytes]) -> Optional[List[str]]:
+    """Parses `hg fix` and returns a list of changeset hashes, or None
+    if no changesets are changed.
```
zzzeid, line 104: The summary line should be a single line (see pep 257)

landoapi/hg.py
```diff
@@ -91,6 +99,21 @@ class PatchConflict(PatchApplicationFailure):
...
+    """
+    for line in fix_output:
+        if not line.startswith(b"REPLACEMENTS: "):
+            continue
+
+        replacements_list = line.split(b"REPLACEMENTS: ", maxsplit=1)[-1].split(b",")
+
+        return [element.decode("latin-1") for element in replacements_list]
+
+    return None
```
zzzeid, line 114:
```suggestion
    # Get first occurrence of matching line
    replacements_line = next(
        (l for l in fix_output if l.startswith(b"REPLACEMENTS")), None)
    if replacements_line:
        # Parse output into changeset IDs
        replacements_list = line.split(b"REPLACEMENTS: ", maxsplit=1)[-1].split(b",")
        return [element.decode("latin-1") for element in replacements_list]
```

landoapi/hg.py
```diff
@@ -286,6 +313,28 @@ def apply_patch(self, patch_io_buf):
                 + ["--logfile", f_msg.name]
             )
 
+    def format(self) -> Optional[List[str]]:
+        """Run `hg fix` to format the currently checked-out stack."""
+        try:
+            logger.info("running `hg fix`")
```
zzzeid, line 319: `self.run_hg` already logs the command being run (same applies to `update` below)

landoapi/hg.py
```diff
@@ -286,6 +313,28 @@ def apply_patch(self, patch_io_buf):
...
+            fix_output = self.run_hg(["fix", "-r", "stack()"]).splitlines()
+
+            # Update the working directory to the latest change
+            logger.info("updating to most recent revision")
+            self.run_hg(["update", "-C", "-r", "tip"])
+
+            formatted_replacements = check_fix_output_for_replacements(fix_output)
+            logger.info(
+                f"replacement revisions found: {formatted_replacements}"
```
zzzeid, line 328:
```suggestion
                f"replacement revisions found: {", ".join(formatted_replacements)}"
```

landoapi/landing_worker.py
```diff
@@ -236,6 +237,15 @@ def run_job(self, job, repo, hgrepo, treestatus, patch_bucket):
...
                     hgrepo.apply_patch(patch_buf)
 
+                # Run `hg fix` configured formatters if enabled
+                if repo.formatting:
+                    replacements = hgrepo.format()
+
+                    # If autoformatting changed any changesets, note those in the job.
+                    if replacements:
+                        job.formatted_replacements = replacements
+                        logger.info(f"replacements: {replacements}")
```
zzzeid, line 247: Is this not already logged [here](https://github.com/mozilla-conduit/lando-api/pull/141/files#diff-6cb69a0bc46f891a3d3fdcb4f2ff92b15f0e3471fcf4008f9a2e630e4935c585R327)?

landoapi/repos.py
```diff
@@ -88,6 +88,13 @@ def __post_init__(self):
         if not self.short_name:
             self.short_name = self.tree
 
+    @property
+    def formatting(self) -> bool:
```
zzzeid, line 92: For consistency, I would suggest renaming this to `has_autoformat_enabled`, or something similar.

zzzeid (changes requested):

landoapi/hg.py
cgsheeh: Should flake8 and/or black have caught this?

zzzeid: Looks like there's at least a [discussion](https://github.com/psf/black/issues/144) about it. Not sure if anything was decided.

zzzeid (changes requested):
Minor changes noted.

landoapi/hg.py
zzzeid: nit:
```suggestion
    """Parse `hg fix` output.
    
    Returns:
        A list of changeset hashes, or None if no changesets are changed.
```

requirements.txt
```diff
@@ -295,9 +295,14 @@ mccabe==0.6.1 \
...
-    --hash=sha256:6222d92d860e411e422b7dc58062132bc512fa52ec48a71143b40cab4be7c829
-    # via -r requirements.in
+mercurial==5.5.1 \
+    --hash=sha256:0f861955143710e0ad711e6fd9b207b415a381efd23516080277400a4d7c0a9a \
+    --hash=sha256:1e65e522dfb79886214da3e2cad97a30e0ba33e981c7257ff12b312709e57a7e \
+    --hash=sha256:28c7ff77c48520903bc6b3ca5a3da045d90ff494fd777177773ebef3d2fef6f1 \
+    --hash=sha256:38173ad4b6e92adf855571fa80ad39354179746272425c0f80c41a05213ef2b4 \
+    --hash=sha256:3dd711c915762c579ce1a7d1ea09e2dca82ed8bcce9d0e78b8f8993123257037 \
+    --hash=sha256:4f95ad8d575941835d67210e9f8700cc594a441e4fd0bd029c084923b4f40874 \
+    --hash=sha256:bdfbcb2685e10d156ac40f4dd5dc3875b781a3e0f2a7f51af7a166bbc084f288
```
zzzeid, line 305: This should be updated in `requirements.in`. Currently, `requirements.in` is showing `mercurial==5.1`. You can run `python -m piptools compile --allow-unsafe --generate-hashes`.

zzzeid (changes requested):
One important change (requirements), and the rest are nits.

zzzeid (approved):

migrations/ffc1d1b8652a_adds_formatted_replacements_field.py
```diff
@@ -0,0 +1,35 @@
+"""adds formatted_replacements field
+
+Revision ID: ffc1d1b8652a
+Revises: 0bf5ff89b7fd
```
zzzeid, line 4: This should be updated to reflect the actual revised revision.

cgsheeh:
This has landed as https://github.com/mozilla-conduit/lando-api/commit/dfa02df4bea01431b6064d8c8354cc4b77594fa5, outside of GitHub's "Rebase and merge" feature due to an issue with CircleCI.

PR #148, opened 2021-02-25
landing jobs: add stats endpoint (bug 1695047)

zzzeid:
Add an API endpoint to check average duration, counts, and number of
attempts of landings given a date range.

### DIFF
--- a/landoapi/api/landing_jobs.py
+++ b/landoapi/api/landing_jobs.py
@@ -4,7 +4,9 @@
 import logging
 
 from connexion import ProblemException
+from datetime import datetime, timedelta
 from flask import g
+from sqlalchemy import func, Date
 
 from landoapi import auth
 from landoapi.models.landing_job import LandingJob, LandingJobStatus, LandingJobAction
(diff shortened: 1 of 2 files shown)


landoapi/api/landing_jobs.py
```diff
@@ -70,3 +72,37 @@ def put(landing_job_id, data):
...
+        start_date: A string representing the start date (e.g. 2020-01-23).
+        end_date: A string representing the end date.
+
+    Returns:
+        Some meta data and the aggregated statistics.
+    """
+    if start_date is None:
+        start_date = datetime.now().date().isoformat()
+    if end_date is None:
+        end_date = (datetime.now() + timedelta(days=1)).date().isoformat()
```
cgsheeh, line 90: I think the start date/end date logic is backwards - start date default should be yesterday (ie `datetime.now() - timedelta(days=1)`) and end date default should be today. Correct me if I'm wrong. :)

Also I think we need to check for more edge cases around ensuring the start/end date represent a valid range of dates (ie if we specify end date earlier than our start date). We can hit this rather easily by specifying only start/end date and using the default based off of `datetime.now()`.

cgsheeh (changes requested):
Sanity check: since this is lando-api it's essentially our 'private' API and thus we don't need auth, right?

Can we test this easily?

landoapi/api/landing_jobs.py
zzzeid: Yes those are fair points -- this was essentially an approach of getting all records up until midnight of the current day. I can clean this up.

zzzeid: > Sanity check: since this is lando-api it's essentially our 'private' API and thus we don't need auth, right?
> 
> Can we test this easily?

We don't need auth for this, and the API is not private (those private endpoints are behind auth). The "private" API got deprecated when we decommissioned Autoland-Transplant.

Yes we can test this via `/stats` endpoint.

cgsheeh (approved):

PR #149, opened 2021-02-26
landing worker: fix AUTOLAND_REQUEST_USER (bug 1695315)

zzzeid:
The environment variable needs to be set before entering the repo
context manager.


cgsheeh (approved):
r+

PR #150, opened 2021-02-26
landing worker: handle request user in context manager (bug 1695335)

zzzeid:
- set and unset $AUTOLAND_REQUEST_USER in HgRepo context manager
- move string to own variable to reduce repetition/typos
- refactor context managers so that there is a separate one that will permit pushing to repo

### DIFF
--- a/landoapi/hg.py
+++ b/landoapi/hg.py
@@ -21,6 +21,9 @@ landing_worker_username = os.environ.get("LANDING_WORKER_USERNAME", "app")
 landing_worker_target_ssh_port = os.environ.get("LANDING_WORKER_TARGET_SSH_PORT", "22")
 REJECTS_PATH = Path("/tmp/patch_rejects")
 
+# Name of the environment variable that will store the push user's email address.
+REQUEST_USER_ENV_VAR = "AUTOLAND_REQUEST_USER"
+
 
 class HgException(Exception):
     @staticmethod
@@ -102,7 +105,7 @@ class HgRepo:
         "ui.merge": "internal:merge",
         "ui.ssh": (
             "ssh "
-            '-o "SendEnv AUTOLAND_REQUEST_USER" '
+            f'-o "SendEnv {REQUEST_USER_ENV_VAR}" '
             '-o "StrictHostKeyChecking no" '
             '-o "PasswordAuthentication no" '
             f'-o "User {landing_worker_username}" '
@@ -117,6 +120,7 @@ class HgRepo:
     def __init__(self, path, config=None):
         self.path = path
         self.config = copy.copy(self.DEFAULT_CONFIGS)
+        self.requester_email = ""
 
         # Somewhere to store patch headers for testing.
         self.patch_header = None
@@ -127,18 +131,32 @@ class HgRepo:
     def _config_to_list(self):
         return ["{}={}".format(k, v) for k, v in self.config.items() if v is not None]
 
+    def __call__(self, request_user):
+        self.request_user = request_user
+        return self
+
     def __enter__(self):
+        if self.requester_email:
+            # Set the environment variable the `set_landing_system` hg extension will
+            # use to send the push user override.
+            os.environ[REQUEST_USER_ENV_VAR] = self.requester_email
+            logger.debug(f"{REQUEST_USER_ENV_VAR} set to {self.requester_email}")
         self.hg_repo = hglib.open(
             self.path, encoding=self.ENCODING, configs=self._config_to_list()
         )
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        if REQUEST_USER_ENV_VAR in os.environ:
+            del os.environ[REQUEST_USER_ENV_VAR]
         try:
             self.clean_repo()
         except Exception as e:
             logger.exception(e)
         self.hg_repo.close()
+
+    def set_request_user(self, requester_email):
+        self.requester_email = requester_email
 
     def clone(self, source):
         # Use of robustcheckout here would work, but is probably not worth
--- a/landoapi/hg.py
+++ b/landoapi/hg.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import copy
+from contextlib import contextmanager
 import logging
 import os
 from pathlib import Path
@@ -120,7 +121,6 @@ class HgRepo:
     def __init__(self, path, config=None):
         self.path = path
         self.config = copy.copy(self.DEFAULT_CONFIGS)
-        self.requester_email = ""
 
         # Somewhere to store patch headers for testing.
         self.patch_header = None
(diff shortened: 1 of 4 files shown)


landoapi/hg.py
```diff
@@ -22,6 +22,9 @@
 landing_worker_target_ssh_port = os.environ.get("LANDING_WORKER_TARGET_SSH_PORT", "22")
 REJECTS_PATH = Path("/tmp/patch_rejects")
 
+# Name of the environment variable that will store the push user's email address.
+REQUEST_USER_ENV_VAR = "AUTOLAND_REQUEST_USER"
```
mitchhentges, line 26: Nice! Arguably, this minor refactor could be in a separate commit, but I understand how stack management is a lot more rough in Github-based workflows :)

landoapi/hg.py
```diff
@@ -129,18 +133,27 @@ def _config_to_list(self):
         return ["{}={}".format(k, v) for k, v in self.config.items() if v is not None]
 
     def __enter__(self):
+        if self.requester_email:
+            # Set the environment variable the `set_landing_system.py` hg
```
mitchhentges, line 137: There might be a word or punctuation missing here.

landoapi/landing_worker.py
```diff
@@ -246,18 +246,37 @@ def run_job(self, job, repo, hgrepo, treestatus, patch_bucket):
             )
             return False
 
-        # Set the environment variable the `set_landing_system.py` hg
-        # extension will use to send the push user override.
-        os.environ["AUTOLAND_REQUEST_USER"] = job.requester_email
+        hgrepo.set_request_user(job.requester_email)
+        with hgrepo:
```
mitchhentges, line 250: This smells a bit like the context manager should be parameterized with the request user, maybe something like:
```
with hgrepo.with_user(job.requester_email):
    ...
```

landoapi/landing_worker.py
```diff
@@ -268,154 +287,126 @@ def run_job(self, job, repo, hgrepo, treestatus, patch_bucket):
                     )
                     self.notify_user_of_landing_failure(job)
                     return True
+                patch_bufs.append((revision_id, patch_buf))
 
-                # Download all patches locally from S3.
```
mitchhentges, line 272: Why is this patch so girthy if most of it is an indentation change? Weird :thinking:

mitchhentges (approved):
Looks good!
There's a few nits, but I trust your judgment here.
Also, heads up that in the commit message "repitition" should be "repetition"

landoapi/landing_worker.py
zzzeid: It's mostly undoing a previous patch that was also an indentation change. Just the way that GitHub parses the diff I guess?

landoapi/hg.py
zzzeid: This might be a remnant of something I copy and pasted, nice catch, will change.

landoapi/hg.py
zzzeid: These commits will be squashed regardless. I'd say because it's part of moving everything to this module, it's also good to keep it in the same PR.

landoapi/hg.py
zzzeid: Ended up just adjusting the line width here, I think the sentence makes sense but kind of awkward to read.

landoapi/hg.py
```diff
@@ -127,19 +131,33 @@ def __init__(self, path, config=None):
     def _config_to_list(self):
         return ["{}={}".format(k, v) for k, v in self.config.items() if v is not None]
 
+    def __call__(self, request_user):
+        self.request_user = request_user
+        return self
```
mitchhentges, line 136: For what it's worth, this could lead to the following situation:
```
with repo("user_a"):
    # user_a is active

with repo:
    # user_a is active, again, despite not being set.
```

This mutable behaviour could accidentally be depended on in the future, which would be bad.
Fortunately, it can be resolved by having `__call__` return a disposable context manager:
```
lil_ctx_manager = repo("user_a")
with lil_ctx_manager:
    # user_a is set

# or, how this would be used:
with repo("user_a")
    # user_a is set

with repo:
    # doesn't have a user set, as expected, because one isn't provided.
```

This adds some more code though, so the tradeoff is up to you :)

mitchhentges (approved):
I sure wish that GitHub could do a nice diff after a force push. Ah well.
Heads up that "repetition" should be spellcheck'd in the commit message :)

zzzeid:
> I sure wish that GitHub could do a nice diff after a force push. Ah well.

I did a force push because otherwise the fix for the conflict with master would have resulted in some ugly commits that I didn't want to push, but, nonetheless I agree 100% with this statement.

> Heads up that "repetition" should be spellcheck'd in the commit message :)
Yep, it's fixed in the actual commit, the PR description is not in sync, but that should be fixed now as well.

landoapi/hg.py
zzzeid: Good point -- I think the existing implementation with `__enter__` and `__exit__` may be too complex to maintain with this approach, so I refactored it into two separate context managers.

landoapi/hg.py
```diff
@@ -321,6 +336,9 @@ def apply_patch(self, patch_io_buf):
             )
 
     def push(self, target, bookmark=None):
+        if not os.getenv(REQUEST_USER_ENV_VAR):
+            raise ValueError(f"{REQUEST_USER_ENV_VAR} not set while attempting to push")
```
mitchhentges, line 340: This state-check is giving a little smell that perhaps the handle for pushing/pulling should be contextual. This ensures that functions like `push()` here can only be called when the appropriate handle has been instantiated, e.g.:
```
class HgPushHandle():
    def __init__(repo):
        self._repo = repo
    
    def push(target, bookmark=None):
        # perhaps set the env here?


class HgRepo():
    # ...
    @contextmanager
    def for_push():
        hg_handle = self._open()
        hg_push_handle = HgPushHandle(hg_handle)
        yield hg_push_handle
        self._clean_and_close(hg_handle)
```

Of course, there's a balance here - increased weight on a Python structural system, but better guard rails against API misuse. I like to lean more on better guard rails, but I'm completely on board if you feel that it adds too many types and that it may not be worth it :)

mitchhentges (approved):
I've left a nit on what flavour I'd prefer, but I think that this is great regardless, and I'm a fan of the contextual `for_push()` and `for_pull()` context managers. Nice work! :tada:

landoapi/hg.py
zzzeid: I think this is out of scope for this PR, but is a good suggestion, though it might need some tweaking. The `for_pull` context manager doesn't only provide the push functionality, it also will provide everything else the `for_pull` context manager does. It would be kind of awkward to do all of that via the "push handle".

mitchhentges: True, though if the `for_push` handle is only used for pushing, it's ok to strip it of non-push functionality, even though it's technically possible. I think that you have better context here, so :+1: from me!

PR #152, opened 2021-03-04
landing worker: fix failed paths parsing (bug 1696396)

zzzeid:
- do not fetch failed paths form raw output
    - look at .rej paths instead
- trim last 4 characters instead of using .rstrip to get file path

### DIFF
--- a/landoapi/landing_worker.py
+++ b/landoapi/landing_worker.py
@@ -230,6 +230,8 @@ class LandingWorker:
         # TODO: capture reason for patch failure, e.g. deleting non-existing file, or
         # adding a pre-existing file, etc...
         reject_paths = rejs_re.findall(exception)
+
+        # Collect all failed paths by removing `.rej` extension.
         failed_paths = [path[:-4] for path in reject_paths]
 
         return failed_paths, reject_paths
@@ -340,6 +342,8 @@ class LandingWorker:
                                     reject["content"] = f.read()
                             except Exception as e:
                                 logger.exception(e)
+                            # Use actual path of file to store reject data, by removing
+                            # `.rej` extension.
                             breakdown["reject_paths"][r[:-4]] = reject
 
                         message = (
--- a/landoapi/landing_worker.py
+++ b/landoapi/landing_worker.py
@@ -227,18 +227,12 @@ class LandingWorker:
             re.MULTILINE,
         )
 
-        # RE to capture files that could not be merged for one reason or another.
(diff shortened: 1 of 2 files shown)


landoapi/landing_worker.py
```diff
@@ -345,7 +340,7 @@ def run_job(self, job, repo, hgrepo, treestatus, patch_bucket):
                                     reject["content"] = f.read()
                             except Exception as e:
                                 logger.exception(e)
-                            breakdown["reject_paths"][r.rstrip(".rej")] = reject
+                            breakdown["reject_paths"][r[:-4]] = reject
```
cgsheeh, line 343: Let's add a comment explaining the significance of `:-4` here.

landoapi/landing_worker.py
```diff
@@ -227,18 +227,12 @@ def extract_error_data(exception: str) -> tuple[list[str], list[str]]:
...
-        # RE to capture files that could not be merged for one reason or another.
-        patterns = (
-            "(?:patching file (.*))",
-            "(?:file (.*) already exists)",
-            "(?:unable to find '(.*)' for patching)",
-        )
+        # TODO: capture reason for patch failure, e.g. deleting non-existing file, or
+        # adding a pre-existing file, etc...
+        reject_paths = rejs_re.findall(exception)
+        failed_paths = [path[:-4] for path in reject_paths]
```
cgsheeh, line 233: Same here.

landoapi/landing_worker.py
```diff
@@ -227,18 +227,12 @@ def extract_error_data(exception: str) -> tuple[list[str], list[str]]:
             re.MULTILINE,
         )
 
-        # RE to capture files that could not be merged for one reason or another.
```
cgsheeh, line 230: If I understand correctly, we have some error messages interspersed with some confirmation/successful messages, and our workaround to fix the problem is to stop parsing for both. Could we (or do we plan to) re-enable this parsing and account for the potential "successfully applied hunk" error message?

cgsheeh (changes requested):

landoapi/landing_worker.py
zzzeid: Yes - I'll file a bug for this. That particular functionality would be needed in order to determine that cause of the conflict, but not specifically to find the failed paths (which can be done just by looking at the .rej paths).

cgsheeh (approved):

PR #156, opened 2021-04-22
ci: push images based on release tags to Docker Hub (bug 1704897)

zzzeid:


mitchhentges (approved):
I don't fully understand the rationale for this change: the bug says that its to always push to `latest`, but this change changes CI to _never_ push to `latest`.

Also, adding a "release number" feels like it could add some friction to the process - perhaps, if you want to have a deterministic release version, you can use the short hash of the current commit?

-----

I may be missing context, however. So long as you've already considered the above questions, then send it ✈️

zzzeid:
> I don't fully understand the rationale for this change: the bug says that its to always push to `latest`, but this change changes CI to _never_ push to `latest`.

I updated the bug to match this implementation. We will deprecate the `latest` tag, and instead, only push release tags to Docker Hub. Jenkins will look only at tags matching the regex to trigger the deployment pipeline.

> Also, adding a "release number" feels like it could add some friction to the process - perhaps, if you want to have a deterministic release version, you can use the short hash of the current commit?

Yes there is a new friction that was added intentionally, so that we can be explicit about our releases. We can add some logic to check that the tag is indeed incremental before deploying, which would reduce the chance of error in terms of accidentally skipping a version number. We want to use tags, and not commit IDs so that we can have a 1:1 mapping between Git tags and Docker Hub tags.

jbuck (approved):

PR #168, opened 2021-07-21
Bug 1721637: return post-formatting hashes for `formatted_replacement…

cgsheeh:
…s` field

At the moment the `formatted_replacements` field is actually the
pre-formatting changeset hashes instead of the post-formatting
hashes, making the value mostly meaningless. Change the value
to the post-formatting hashes to make the value more useful.

### DIFF
--- a/landoapi/hg.py
+++ b/landoapi/hg.py
@@ -391,14 +391,20 @@ class HgRepo:
         # Update the working directory to the latest change.
         self.run_hg(["update", "-C", "-r", "tip"])
 
-        formatted_replacements = check_fix_output_for_replacements(fix_output)
-        logger.info(
-            f"revisions were reformatted: {', '.join(formatted_replacements)}"
-            if formatted_replacements
-            else "autoformatting did not change any revisions"
+        # Exit if no revisions were reformatted.
+        pre_formatting_hashes = check_fix_output_for_replacements(fix_output)
+        if not pre_formatting_hashes:
+            return None
+
+        post_formatting_hashes = (
+            self.run_hg(["log", "-r", "stack()", "-T", "{node}\n"])
+            .decode("utf-8")
+            .splitlines()[len(pre_formatting_hashes) - 1 :]
         )
 
-        return formatted_replacements
+        logger.info(f"revisions were reformatted: {', '.join(post_formatting_hashes)}")
+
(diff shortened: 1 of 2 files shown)


landoapi/hg.py
```diff
@@ -391,14 +391,21 @@ def format(self) -> Optional[List[str]]:
...
 
-        formatted_replacements = check_fix_output_for_replacements(fix_output)
-        logger.info(
-            f"revisions were reformatted: {', '.join(formatted_replacements)}"
-            if formatted_replacements
-            else "autoformatting did not change any revisions"
+        # Exit if no revisions were reformatted.
+        pre_formatting_hashes = check_fix_output_for_replacements(fix_output)
+        if not pre_formatting_hashes:
+            logger.info("autoformatting did not change any revisions")
```
mitchhentges, line 397: Unsure if this log line is valuable, but that's up to you.

mitchhentges, line 397: Unsure if this log line is valuable, but that's up to you.

landoapi/hg.py
```diff
@@ -391,14 +391,21 @@ def format(self) -> Optional[List[str]]:
...
+            return None
+
+        post_formatting_hashes = (
+            self.run_hg(["log", "-r", "stack()", "-T", "{node}\n"])
+            .decode("utf-8")
+            .splitlines()[len(pre_formatting_hashes) - 1 :]
         )
 
-        return formatted_replacements
+        logger.info(f"revisions were reformatted: {', '.join(post_formatting_hashes)}")
```
mitchhentges, line 406: Note that [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1721637#c0) says that you'll print both the pre- and post-formatting hashes. I'm indifferent, and I'm guessing that the pre-hash is sufficiently unhelpful that it's not worth printing.

mitchhentges (approved):
Neat :+1:

landoapi/hg.py
cgsheeh: Yes, I thought about it again and decided to just post the node of the commit that would actually land.

landoapi/hg.py
cgsheeh: I agree it's not super useful - dumping it.

PR #169, opened 2021-08-11
black: update to 21.5b2 and fix formatting issues (bug 1714229)

zzzeid:


Dockerfile
```diff
@@ -40,6 +40,7 @@ RUN addgroup --gid 10001 app \
 ENV HGPYTHON3=
 
 COPY requirements.txt /python_requirements.txt
+RUN pip install pip --upgrade
```
cgsheeh, line 43: I think this needs to be `pip install --upgrade pip`

Dockerfile-dev
```diff
@@ -28,6 +28,7 @@ ENV HGPYTHON3=
 
 # TODO: there are some packages here that we do not need.
 COPY requirements.txt /python_requirements.txt
+RUN pip install pip --upgrade
```
cgsheeh, line 31: `pip install --upgrade pip`

tests/test_landing_job.py
```diff
@@ -90,7 +90,9 @@ def test_cancel_landing_job_fails_bad_input(db, client, landing_job, auth0_mock)
     )
 
     assert response.status_code == 400
-    assert response.json["detail"] == ("'IN_PROGRESS' is not one of ['CANCELLED']")
+    assert response.json["detail"] == (
```
cgsheeh, line 93: Is the addition of ` - 'status'` leftover from when this PR had Flask v2? Confused as to how this changes after this PR.

cgsheeh (changes requested):

tests/test_landing_job.py
zzzeid: This is because of the upgrade to connexion, not Flask.

PR #171, opened 2021-08-18
landing jobs: allow cancelling deferred jobs (bug 1656271)

zzzeid:


cgsheeh (changes requested):
Seems simple enough. Can you add a test to `test_landing_job.py`?

cgsheeh (approved):
Ship it!

PR #176, opened 2021-11-23
requirements: remove typing and update black (Bug 1742720)

cgsheeh:
- requirements: remove `typing` module (Bug 1742720)
- requirements: update black to 21.11b1 (Bug 1742720)

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -23,5 +23,4 @@ redis==3.2.1
 requests-mock==1.6.0
 requests==2.25.1
 rs-parsepatch==0.3.3
-typing==3.6.6
 uWSGI==2.0.18
--- a/requirements.in
+++ b/requirements.in
@@ -2,7 +2,7 @@ Flask-Alembic==2.0.1
 Flask-Caching==1.10.1
 Flask==1.1.4
 atomicwrites==1.3.0
-black==21.10b0
+black==21.11b1
 blinker==1.4
 celery==4.3.0
 connexion==2.7.0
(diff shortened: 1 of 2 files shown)


zzzeid (approved):
LGTM. FWIW General approach in Lando is to squash commits into a single one with general description in the first line and any detail that breaks down other commit descriptions in the description. I noticed the commits on Nov. 18 were not squashed (e.g. the 3 autoformatting ones)

cgsheeh:
@zzzeid I just pressed "Rebase and merge" in the PR web view, which is the default for me. Personally would I strongly prefer the rebase and merge workflow and it might be worth examining when/why we decided to squash commits on this repo (I can see a few examples of stacks being rebased and landed without squashing from when we reviewed changes on Phabricator).

I'm fine with either but we should disable the rebasing option in Github if we want everything squashed.

zzzeid:
> @zzzeid I just pressed "Rebase and merge" in the PR web view, which is the default for me. Personally would I strongly prefer the rebase and merge workflow and it might be worth examining when/why we decided to squash commits on this repo (I can see a few examples of stacks being rebased and landed without squashing from when we reviewed changes on Phabricator).
> 
> I'm fine with either but we should disable the rebasing option in Github if we want everything squashed.

Good point, definitely worth a review. I suppose both are fine and the rebasing option specifically is great when there are multiple things being impacted (instead of say having multiple PRs depending on each other).

The main thing that caught my attention was the 3 commits affecting the same bug number and concept (autoformatting). Ultimately it doesn't matter much how we do this as long as the commits that land are clean.

PR #177, opened 2021-11-26
revisions: add revision worker and mots integration (bug 1740107)

zzzeid:
This is a work in progress, do not merge!

- add new command (lando-cli revision-worker) to start new worker
- add new `revision-worker` that pre-processes revisions
- add method to parse diff and list affected files (should move to mots)

TODO:
- [ ] refactor revision worker to extend a general worker


zzzeid:
This is now implemented in #178

PR #179, opened 2021-12-09
Bug 1745292: enable using alternate endpoint url for s3

cgsheeh:
This commit adds some new code to enable using a non-standard AWS
endpoint for S3 operations. An `S3_ENDPOINT_URL` environment variable
is added that can be used to specify the endpoint for use. If the
environment variable is omitted the production S3 url will be used
instead. This enables passing an url such as `http://mock-s3` which
will direct to a moto server instance for simple testing.

Since the moto server would not have an appropriate S3 bucket created
for use, we add a `--init-s3` flag that can be passed to the lando-cli
database init script which creates a new bucket named using the
`PATCH_BUCKET` environment variable.


zzzeid (approved):
lgtm! This should tide us over until we deprecate S3.

PR #180, opened 2021-12-15
Bug 1745293: amend release managers review check to check all required uplift conditions

cgsheeh:
This commit extends the existing uplift blocking check to include
checking for a submitted uplift request form somewhere on the
stack. The function name is updated to reflect the change.

### DIFF
--- a/landoapi/repos.py
+++ b/landoapi/repos.py
@@ -96,6 +96,11 @@ class Repo:
         )
 
 
+SCM_ALLOW_DIRECT_PUSH = AccessGroup(
+    active_group="active_scm_allow_direct_push",
+    membership_group="all_scm_allow_direct_push",
+    display_name="Above Level 3 Commit Access",
+)
 SCM_LEVEL_3 = AccessGroup(
     active_group="active_scm_level_3",
     membership_group="all_scm_level_3",
@@ -176,6 +181,7 @@ REPO_CONFIG = {
             push_path="ssh://autoland.hg//repos/third-repo",
             pull_path="http://hg.test/third-repo",
             config_override={"fix.black:command": "black -- -"},
+            approval_required=True,
         ),
         # Approval is required for the uplift dev repo
         "uplift-target": Repo(
@@ -198,6 +204,7 @@ REPO_CONFIG = {
             access_group=SCM_CONDUIT,
             commit_flags=[DONTBUILD],
             config_override={"fix.black:command": "black -- -"},
+            approval_required=True,
         ),
         "vct": Repo(
             tree="vct",
@@ -280,6 +287,12 @@ REPO_CONFIG = {
             tree="mozilla-build",
             url="https://hg.mozilla.org/mozilla-build",
             access_group=SCM_LEVEL_3,
+        ),
+        "beta": Repo(
+            tree="mozilla-beta",
+            url="https://hg.mozilla.org/releases/mozilla-beta",
+            access_group=SCM_ALLOW_DIRECT_PUSH,
+            approval_required=True,
         ),
     },
 }
--- a/landoapi/reviews.py
+++ b/landoapi/reviews.py
@@ -4,12 +4,15 @@
 
 import logging
 from collections import namedtuple
-from typing import List
+from typing import List, Tuple
 
 from landoapi.phabricator import (
     PhabricatorClient,
     PhabricatorCommunicationException,
     ReviewerStatus,
+)
+from landoapi.projects import (
+    RELMAN_PROJECT_SLUG,
 )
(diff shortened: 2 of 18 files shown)


tests/mocks.py
```diff
@@ -235,8 +236,12 @@ def revision(
             "ccs": [],
             "hashes": [],
             "bugzilla.bug-id": bug_id,
+            "uplift.request": uplift,
             "repositoryPHID": repo["phid"] if repo is not None else None,
-            "fields": {"repositoryPHID": repo["phid"] if repo is not None else None},
+            "fields": {
+                "repositoryPHID": repo["phid"] if repo is not None else None,
+                "uplift.request": uplift,
+            },
```
cgsheeh, line 244: I added to the main dict in the same way as `bugzilla.bug-id`, but the object doesn't return as expected. Not sure if we test `bugzilla.bug-id` but perhaps it's supposed to go in the `fields` dict?

landoapi/uplift.py
```diff
@@ -197,3 +208,12 @@ def create_approval_request(phab: PhabricatorClient, revision: dict, form_conten
         "revision_id": rev_id,
         "revision_phid": rev_phid,
     }
+
+
+def stack_uplift_form_submitted(stack_data) -> bool:
+    """Return `True` if the stack has a valid uplift request form submitted."""
+    # TODO just checks that any of the revisions in the stack have the uplift form
+    # submitted.
```
cgsheeh, line 216: This check is basically "does some revision in the stack have an uplift request form submitted". The `TODO` is mostly to check if others have opinions on whether this should be a more complicated check. For example we could check the tip commit of the changeset being landed, which would mean partial landings of stacks require a submitted form on each submitssion.

zzzeid: Maybe this should be a `NOTE`?
```suggestion
    # NOTE: this just checks that any of the revisions in the stack have the uplift
    # form submitted.
```

landoapi/repos.py
```diff
@@ -281,6 +288,12 @@ def autoformat_enabled(self) -> bool:
             url="https://hg.mozilla.org/mozilla-build",
             access_group=SCM_LEVEL_3,
         ),
+        "beta": Repo(
+            tree="",
+            url="https://hg.mozilla.org/releases/mozilla-beta",
+            access_group=SCM_ALLOW_DIRECT_PUSH,
```
zzzeid, line 294: Are we technically overloading this access group? I imagine this might be flagged by security team as something that should have an explicit access group.

landoapi/reviews.py
```diff
@@ -165,3 +168,41 @@ def reviewers_for_commit_message(
...
+        sec_approval_phid: The PHID string of the sec-approval project.
+
+    Returns:
+        A list of strings.
+    """
+    # Approvals are reviews where the user is in the `release-managers` group.
+    approvals = [
+        reviewer_identity(phid, users, projects).identifier
+        for phid, r in reviewers.items()
+        if phid in relman_phids and r["status"] is ReviewerStatus.ACCEPTED
```
zzzeid, line 198: There is a bit of inconsistency when comparing against `ReviewerStatus` enums -- in some places, `is` is used, in others, `==` is used. We should probably stick to one or the other. It seems that in most cases `collate_reviewer_attachments` is used, which maps strings back to enum. It should be safe to just do a normal comparison by value. See [this line](https://github.com/mozilla-conduit/lando-api/pull/180/files#diff-d1d8509fc2fac61f40b9c9231b59a27dcff8fb6e2d443fa1e32e240898c55723R162).

zzzeid (changes requested):
Few minor comments on first pass, but looking good otherwise.

landoapi/repos.py
cgsheeh: We could look at creating a separate access group to tighten restrictions on the number of people who can push vs the number of people who can land uplifts through this new workflow. For now this just reflects the current state of the world, we only want relman + sheriffs and a select few others (releng etc) doing uplift landings. I'll file a follow up bug for this.

landoapi/reviews.py
cgsheeh: I think `==` makes more sense here as well. I pulled this line from `reviewers_for_commit_message` above, which uses `is`. We can make things more consistent as a follow up.

cgsheeh:
Example landing on dev showing rewrite from `r=` to `a=`: https://hg.mozilla.org/conduit-testing/m-c/rev/190dce644ae84f911c9a16d262d3bafb507dacad

zzzeid (approved):
This looks good to me!

PR #181, opened 2022-01-13
Bug 1749754: add a landing warning to check if a patch has `WIP:` in the title

cgsheeh:
Adds a revision warning to assert a revision does not have `WIP:`
as the beginning of it's revision title. The presence of `WIP:`
indicates a work-in-progress revision and should be scrubbed before
landing by the user.

We could follow up with some code to automatically scrub the `WIP`
text from the commit message, but this is a good start for now.

### DIFF
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -295,6 +295,13 @@ def warning_diff_warning(*, revision, diff, **kwargs):
         return [w.data for w in warnings]
 
 
+@RevisionWarningCheck(7, "Revision is marked as WIP.")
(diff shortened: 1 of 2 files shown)


zzzeid (approved):
lgtm -- would be nice to see a screenshot of Lando UI as an example.

cgsheeh:
I'll deploy this to dev quickly and post a screenshot. :+1:

cgsheeh:
When I tried to make a screenshot I realized I didn't add the check to the list of checks in `check_landing_warnings` so it wasn't actually running. That's fixed in the latest version.

Here's a screenshot:
wip_warning

PR #185, opened 2022-02-17
treestatus: consider "approval required" status as open for landing (Bug 1756036)

cgsheeh:
Within the new uplift workflow, repositories on Lando with
"approval required" status should be considered open for landing.
The approval is checked in Lando by asserting `release-managers`
review on the revision and the hg hook asserts `a=<reviewer>` is
present in the first line of the commit message. Thus we can
consider an "approval required" repo as landable and enabled.

### DIFF
--- a/landoapi/treestatus.py
+++ b/landoapi/treestatus.py
@@ -16,8 +16,16 @@ class TreeStatus:
 
     DEFAULT_URL = "https://treestatus.mozilla-releng.net"
 
+    # A repo is considered open for landing when either of these
+    # statuses are present.
+
+    # For the "approval required" status Lando will enforce the appropriate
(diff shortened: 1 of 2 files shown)


zzzeid (approved):
This looks good to me. As long as the enforcement is happening appropriately elsewhere (which it looks like both Lando and hgmo are going to be doing).

zzzeid:
> This looks good to me. As long as the enforcement is happening appropriately elsewhere (which it looks like both Lando and hgmo are going to be doing).

To clarify, by enforcement in Lando, I meant that the final commit message should be sanitized using only actual approvals and that users shouldn't be able to coerce/inject an approval by other means.

PR #186, opened 2022-02-17
commit-message: expand instances of review specifiers (Bug 1756048)

cgsheeh:
Remove the conservative `R_SPECIFIER` and replace with `SPECIFIER`
everywhere. This expands the values used for review specifiers
to catch and scrub `a=` and others when landing code.


zzzeid (commented):
checks are failing here, does this need to be updated to make tests pass?

cgsheeh:
> checks are failing here, does this need to be updated to make tests pass?

Turns out this doesn't work. It needs to be updated.

zzzeid (approved):
lgtm.

PR #188, opened 2022-02-22
approvals: only rewrite RelMan reviews as `a=` for approval required repositories (Bug 1755932)

cgsheeh:
Fixes a regression caused by #180 where all RelMan reviews are
rewritten from `r=reviewer` to `a=reviewer`. This should only
occur on an uplift repository, so we amend usages of
`approvals_for_commit_message` to only apply when this flag is
set.

Some of the secure revision tests fail after this change due to
a missing repository tagged on the revision so we add a repo to
these tests.

### DIFF
--- a/landoapi/reviews.py
+++ b/landoapi/reviews.py
@@ -186,10 +186,12 @@ def approvals_for_commit_message(
         reviewers: Dict of {reviewer_phid: reviewer_data}
         users: List of Phabricator Users that were involved in the revision.
         projects: List of Phabricator Projects that were involved in the revision.
+        relman_phids: List of Phabricator users in the `release-managers` group.
         sec_approval_phid: The PHID string of the sec-approval project.
 
     Returns:
-        A list of strings.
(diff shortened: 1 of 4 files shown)


landoapi/reviews.py
```diff
@@ -186,10 +186,12 @@ def approvals_for_commit_message(
         reviewers: Dict of {reviewer_phid: reviewer_data}
         users: List of Phabricator Users that were involved in the revision.
         projects: List of Phabricator Projects that were involved in the revision.
+        relman_phids: List of Phabricator users in the `release-managers` group.
         sec_approval_phid: The PHID string of the sec-approval project.
 
     Returns:
-        A list of strings.
+        A tuple of lists of strings, representing `r=` reviewers and `a=` approvers
+        respectively.
```
cgsheeh, line 194: I realized I forgot to update the docstring in #180 so I added this as a ride-along.

zzzeid (approved):
lgtm

PR #190, opened 2022-03-03
repos: fix typing issue for `repos.commit_flags` (Bug 1759890)

cgsheeh:
PyRight is showing errors in my editor as the value of
`commit_flags` defaults to `None` but the type should
only be a list of tuples. Use the `field` function to
define a default factory for the value, resulting in
a consistent type definition for the fields.

### DIFF
--- a/landoapi/repos.py
+++ b/landoapi/repos.py
@@ -6,7 +6,10 @@ import logging
 import pathlib
 import urllib
 from collections import namedtuple
-from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


cgsheeh:
There are a bunch of these small typing issues in Lando, that I plan to pick off as I see them. I can file a bug and aggregate them there if we need a bug for this change.

zzzeid (approved):
lgtm

PR #193, opened 2022-04-06
migrations: migrate to flask-migrate (bug 1694242)

zzzeid:
See note below before landing and deploying. This deployment will need to be coordinated with the Cloud Ops team, and needs to be tested on the development environment with a fresh clone of the production DB before we deploy to production.

- remove flask-alembic dependency
- add flask-migrate dependency
- update code to work with flask-migrate
- upgrade alembic, sqlalchemy packages
- remove all previous migrations
- reinitialize all migrations based on latest state
- remove init cli command (keep init_s3 for now)
- update tasks
- update readme with instructions to generate migrations

NOTE:

This PR will need to be deployed manually, and the `alembic_version` table needs to be updated with `version_num` set to `7883d80258fb` manually, before the migration step is performed.


migrations/versions/c02a75d4f418_.py
```diff
@@ -0,0 +1,160 @@
+"""empty message
```
zzzeid, line 1: Should replace this with a message describing this initial migration as matching the previous state of the database.

landoapi/cli.py
```diff
@@ -52,24 +52,16 @@ def cli():
...
-            aws_secret_key=os.environ["AWS_SECRET_KEY"],
-            endpoint_url=os.environ["S3_ENDPOINT_URL"],
-        )
-        s3.create_bucket(Bucket=os.environ["PATCH_BUCKET_NAME"])
+    s3 = patches.create_s3(
+        aws_access_key=os.environ["AWS_ACCESS_KEY"],
+        aws_secret_key=os.environ["AWS_SECRET_KEY"],
+        endpoint_url=os.environ["S3_ENDPOINT_URL"],
+    )
+    s3.create_bucket(Bucket=os.environ["PATCH_BUCKET_NAME"])
```
zzzeid, line 64: Maybe there's a better place for this elsewhere?

migrations/alembic.ini
```diff
@@ -0,0 +1,50 @@
+# A generic, single database configuration.
```
zzzeid, line 1: Should verify that the settings in this file are being read correctly in all environments.

landoapi/cli.py
```diff
@@ -52,24 +52,16 @@ def cli():
...
-def init(init_s3):
-    """Initialize Lando API (Create the DB, etc.)"""
-    # Create the database and set the alembic version to
-    # head revision.
-    from landoapi.storage import alembic, db
-
-    db.create_all()
-    alembic.stamp("head")
-
+def init_s3(init_s3):
```
cgsheeh, line 55: If we keep this function you'll need to remove the `init_s3` argument for it to work.

landoapi/cli.py
cgsheeh: If you can think of somewhere else to put it that would be r+ from me. Though IIRC you have some work upcoming to remove S3 from Lando entirely, I would be fine with leaving this here until that lands and then this becomes redundant.

migrations/alembic.ini
cgsheeh, line 1: Needs a license header.

migrations/env.py
```diff
@@ -0,0 +1,89 @@
+from __future__ import with_statement
```
cgsheeh, line 1: License.

migrations/versions/c02a75d4f418_.py
cgsheeh, line 1: License.

cgsheeh (changes requested):
Just need to add license headers then this is r+. Thanks for doing this, it was a huge headache when working on autoformat.

migrations/versions/c02a75d4f418_.py
zzzeid: We don't normally put license headers on these autogenerated migration files (check all deleted migration files below/above).

cgsheeh (approved):

PR #196, opened 2022-04-11
migrations, requirements: general improvements (bugs 1694242, 1754816)

zzzeid:
See note below before landing and deploying. This deployment will need to be coordinated with the Cloud Ops team, and needs to be tested on the development environment with a fresh clone of the production DB before we deploy to production.

- remove flask-alembic dependency
- add flask-migrate dependency
- update code to work with flask-migrate
- upgrade alembic, sqlalchemy packages
- remove all previous migrations
- reinitialize all migrations based on latest state
- remove init cli command (keep init_s3 for now)
- update tasks
- update readme with instructions to generate migrations

NOTE:

This PR will need to be deployed manually, and the `alembic_version` table needs to be updated with `version_num` set to `7883d80258fb` manually, before the migration step is performed.

Also includes:

requirements: major package upgrades to Flask et al (bug 1754816)

- migrate to using sentry-sdk from raven
- add before_send hook to sentry (bug 1766626)
- upgrade Flask, Connexion, Mercurial
- remove old, stale requirements (blinker, py, pluggy, atomicwrites)


migrations/README.md
```diff
@@ -0,0 +1,33 @@
...
+-------------------------
+
+To create a new migration file (e.g. after updating your models), follow these steps:
+1. Make sure that your database is created and up to date, by running: 
+
+```
+$ docker-compose run lando-api db upgrade
+```
+
+4. Generate a new revision by running
```
cgsheeh, line 11: Docs, very nice!

Guessing there were more steps before hand, so this should now be 2, then 3 etc.

migrations/alembic.ini
```diff
@@ -0,0 +1,54 @@
...
+
+# A generic, single database configuration.
+
+[alembic]
+# template used to generate migration files
+# file_template = %%(rev)s_%%(slug)s
+
+# set to 'true' to run the environment during
+# the 'revision' command, regardless of autogenerate
+# revision_environment = false
```
cgsheeh, line 13: Should these be removed/uncommented?

migrations/versions/7883d80258fb_initial.py
```diff
@@ -0,0 +1,160 @@
+"""initial
```
cgsheeh, line 1: I think this is auto-generated, but if it isn't we need a license header.

cgsheeh (changes requested):

migrations/README.md
zzzeid: Heh. Yes indeed, will fix.

migrations/alembic.ini
zzzeid: This file is itself autogenerated as well, a one-time deal. I don't see a harm in removing or keeping these comments, but there is no need to uncomment them. I will remove them to keep this file clean.

migrations/versions/7883d80258fb_initial.py
zzzeid: It is autogenerated-- I think we discussed it in the previous iteration of this PR.

cgsheeh (approved):

PR #198, opened 2022-04-11
transplants: deprecate legacy transplants (bug 1654646)

zzzeid:
WIP DO NOT LAND

- remove code flows no longer in use
- remove deprecated tests, factories, etc.
- update health check tests


zzzeid:
This is now implemented as part of #178

PR #199, opened 2022-04-12
meta: add type/assertion hints and upgrade pytest

cgsheeh:
- testing: add some assertion hints to test cases (Bug 1759890)
- typing: add type hints in a few places (Bug 1759890)
- testing: upgrade pytest to 7.1.1 (Bug 1754816)

### DIFF
--- a/landoapi/treestatus.py
+++ b/landoapi/treestatus.py
@@ -29,7 +29,7 @@ class TreeStatus:
         self.url = self.url if self.url[-1] == "/" else self.url + "/"
         self.session = session or self.create_session()
 
-    def is_open(self, tree):
+    def is_open(self, tree: str) -> bool:
         if not tree:
             raise ValueError("tree must be a non-empty string")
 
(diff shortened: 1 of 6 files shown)


cgsheeh:
These are a few of the smaller changes in a larger patch stack - pulling them out and submitting them to make the stack smaller/review of the larger change easier.

landoapi/treestatus.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, *, url=None, session=None):
         self.url = self.url if self.url[-1] == "/" else self.url + "/"
         self.session = session or self.create_session()
 
-    def is_open(self, tree):
+    def is_open(self, tree) -> bool:
```
zzzeid, line 32: nit: Why not also add type hints for the arguments as well, here and below -- since you're here already?

zzzeid (approved):
Minor comment otherwise looks good.

landoapi/treestatus.py
cgsheeh: Good point, I'll do that now.

PR #200, opened 2022-04-26
uplift: fix endpoint and add support for stack uplifts (Bug 1764213)

cgsheeh:
Remove code related to "approval requests" from the uplift endpoint.
Add `move_drev_to_original` from moz-phab to keep summary consistent.
Change the incorrect `commit` key on the diff to the correct
`rev` key to fix the endpoint. Add some type hints to various pieces
of code.

### DIFF
--- a/landoapi/api/uplift.py
+++ b/landoapi/api/uplift.py
@@ -15,11 +15,23 @@ from landoapi.transplants import convert_path_id_to_phid
 from landoapi.uplift import (
     create_uplift_revision,
     get_uplift_conduit_state,
+    get_uplift_repositories,
 )
 from landoapi.validation import parse_landing_path
 from landoapi.decorators import require_phabricator_api_key
 
 logger = logging.getLogger(__name__)
+
+
+@auth.require_auth0(scopes=("lando", "profile", "email"), userinfo=True)
+def get(_):
+    """"""
+    phab: PhabricatorClient = g.phabricator
+    repos = [
+        phab.expect(repo, "fields", "name") for repo in get_uplift_repositories(phab)
+    ]
+
+    return {"repos": repos}, 201
 
 
 @require_phabricator_api_key(optional=False)
--- a/landoapi/api/uplift.py
+++ b/landoapi/api/uplift.py
@@ -116,7 +116,7 @@ def create(data):
 
         if not base_revision:
             # Base revision hash is available on the diff fields.
-            refs = {ref["type"]: ref for ref in phab.expect(revision, "fields", "refs")}
+            refs = {ref["type"]: ref for ref in phab.expect(diff, "fields", "refs")}
             base_revision = refs["base"]["identifier"]
 
         # Get the parent commit PHID from the stack if available.
--- a/landoapi/api/uplift.py
+++ b/landoapi/api/uplift.py
@@ -125,6 +125,11 @@ def create(data):
                 attachments={"commits": True},
                 constraints={"phids": [parent["diff_phid"]]},
             )
+            parent_diff = phab.single(parent_diff, "data", none_when_empty=True)
+
+            if not parent_diff:
+                raise Exception("No parent diff commits.")
+
             commits = phab.expect(parent_diff, "attachments", "commits", "commits")
             parent_revision = commits[0] if commits else None
 
--- a/landoapi/api/uplift.py
+++ b/landoapi/api/uplift.py
@@ -14,6 +14,7 @@ from landoapi.repos import get_repos_for_env
 from landoapi.transplants import convert_path_id_to_phid
 from landoapi.uplift import (
     create_uplift_revision,
+    get_local_uplift_repo,
     get_uplift_conduit_state,
     get_uplift_repositories,
 )
@@ -75,6 +76,7 @@ def create(data):
             revision_id=tip_revision_id,
             target_repository_name=repo_name,
         )
+        local_repo = get_local_uplift_repo(phab, target_repository)
         logger.info("Approval state is valid")
     except ValueError as err:
         logger.exception(
@@ -117,7 +119,13 @@ def create(data):
         try:
             # Create the revision.
             rev = create_uplift_revision(
-                phab, revision, diff, parent_phid, relman_phid, target_repository
+                phab,
+                local_repo,
+                revision,
+                diff,
+                parent_phid,
+                relman_phid,
+                target_repository,
             )
             commit_stack.append(rev)
         except Exception as e:
--- a/landoapi/api/uplift.py
+++ b/landoapi/api/uplift.py
@@ -116,6 +116,9 @@ def create(data):
         # Get the parent commit PHID from the stack if available.
         parent_phid = commit_stack[-1]["revision_phid"] if commit_stack else None
 
+        # TODO set this properly.
+        # parent_revision = commit_stack[-1][""] if commit_stack else None
+
         try:
             # Create the revision.
             rev = create_uplift_revision(
@@ -124,6 +127,7 @@ def create(data):
                 revision,
                 diff,
                 parent_phid,
+                # parent_revision,
                 relman_phid,
                 target_repository,
             )
--- a/landoapi/api/uplift.py
+++ b/landoapi/api/uplift.py
@@ -24,7 +24,6 @@ logger = logging.getLogger(__name__)
 
 
 @require_phabricator_api_key(optional=True)
-@auth.require_auth0(scopes=("lando", "profile", "email"), userinfo=True)
 def get():
     """"""
     phab: PhabricatorClient = g.phabricator
--- a/landoapi/api/uplift.py
+++ b/landoapi/api/uplift.py
@@ -105,7 +105,7 @@ def create(data):
 
     landing_path = convert_path_id_to_phid(landing_path, revision_data)
     commit_stack = []
-    parent_phid = parent_revision = None
+    base_revision = parent_phid = parent_revision = None
     for rev_id, _diff_id in landing_path:
         # Get the relevant revision.
         revision = revision_data.revisions[rev_id]
@@ -113,6 +113,11 @@ def create(data):
         # Get the relevant diff.
         diff_phid = phab.expect(revision, "fields", "diffPHID")
         diff = revision_data.diffs[diff_phid]
+
+        if not base_revision:
+            # Base revision hash is available on the diff fields.
+            refs = {ref["type"]: ref for ref in phab.expect(revision, "fields", "refs")}
+            base_revision = refs["base"]["identifier"]
 
         # Get the parent commit PHID from the stack if available.
         if commit_stack:
@@ -138,6 +143,7 @@ def create(data):
             rev = create_uplift_revision(
                 phab,
                 local_repo,
+                base_revision,
                 revision,
                 diff,
                 parent_phid,
--- a/landoapi/api/uplift.py
+++ b/landoapi/api/uplift.py
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 
 
 @auth.require_auth0(scopes=("lando", "profile", "email"), userinfo=True)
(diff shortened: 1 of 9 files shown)


landoapi/api/uplift.py
```diff
@@ -23,49 +22,51 @@
...
 def create(data):
     """Create new uplift requests for requested repository & revision"""
+    repo_name, revision_id = data["repository"], data["revision_id"]
 
-    # Validate repository
+    # Validate repository.
     all_repos = get_repos_for_env(current_app.config.get("ENVIRONMENT"))
-    repository = all_repos.get(data["repository"])
+    repository = all_repos.get(repo_name)
     if repository is None or not repository.approval_required:
```
zzzeid, line 30: The second part of this `if` statement is redundant (see line 38 below):

```python
    if not repository.approval_required:
        return problem(
            400,
            f"Repository {repo_name} is not an uplift repository.",
            "",
        )
```

landoapi/api/uplift.py
```diff
@@ -23,49 +22,51 @@
...
-        revision_id=data["revision_id"],
-        target_repository_name=data["repository"],
+        revision_id=revision_id,
+        target_repository_name=repo_name,
     )
     logger.info(
         "Approval state is valid",
-        extra={"style": "approval" if is_approval else "uplift"},
+        extra={"style": "uplift"},
     )
```
zzzeid, line 61: Is there any case when the approval state is not "valid" after checking? Seems like `check_approval_state` always returns something if the revision/repo exist -- so it may be a misnomer (is it just checking whether the repo/revision exist?).

landoapi/cli.py
```diff
@@ -61,15 +61,18 @@ def init(init_s3):
 
     db.create_all()
     alembic.stamp("head")
+    logging.info("created lando db")
```
zzzeid, line 64: Is this for debugging?

landoapi/uplift.py
```diff
@@ -145,8 +136,9 @@ def create_uplift_revision(
     )
 
     # Append an uplift mention to the summary
-    summary = phab.expect(source_revision, "fields", "summary")
-    summary += f"\nNOTE: Uplifted from D{source_revision['id']}"
+    summary = str(phab.expect(source_revision, "fields", "summary"))
+    summary += f"\nOriginal Revision: {phab.url_base}/D{source_revision['id']}"
+    summary = summary.strip()
```
zzzeid, line 141: nit: is calling `str()` necessary here? Same for `summary.strip`. Perhaps we can strip the phab summary since the additional line is not contributing any extraneous blank space.

zzzeid (changes requested):
few small comments otherwise lgtm

landoapi/uplift.py
cgsheeh: I added the call to `str` because technically `phab.expect` doesn't return a concrete type, so my editor was complaining about type issues. I'm fine with removing it but it shouldn't change behaviour here, just give PyRight a hint that `summary` should be a `str`.

I added the call to `strip` to avoid having a blank line above `Original Revision` if the summary was empty before uplift. As I'm writing this I realize that we don't replace the `Differential Revision` line in the commit message in this iteration, so I'll need to change this anyways.

landoapi/api/uplift.py
cgsheeh: I will change the name of the function - in the previous format this function checked for uplift/approval but now it serves a different purpose.

landoapi/cli.py
cgsheeh: I added this but it doesn't actually display anything - I think because lando's logging hasn't been initiated yet in this code path. I'll remove it.

landoapi/uplift.py
```diff
@@ -22,6 +23,48 @@
...
+    original_revision = ORIGINAL_DIFF_REV_RE.search(body)
+
+    # If both match, we already have an `Original Revision` line.
+    if differential_revision and original_revision:
+        return body
+
+    def repl(match):
+        phab_url = match.group("phab_url")
+        rev = match.group("rev")
+        return f"\nOriginal Revision: {phab_url}/D{rev}"
```
zzzeid, line 62: Could be a little cleaner like this:

```suggestion
        return "\nOriginal Revision: {phab_url}/D{rev}".format(**match.groupdict())
```

Or perhaps `repl` is not necessary as a method.

cgsheeh: We already have this code in `moz-phab`: https://github.com/mozilla-conduit/review/blob/main/mozphab/helpers.py#L219

I copied this over to keep the logic for moving `Differential Revision` consistent between lando and moz-phab, but removed the extra `rev_id` parsing that we have in moz-phab. I'm probably going to carry that into `moz-phab` and have the `rev_id` parsing take place elsewhere.

`repl` is necessary for `re.sub`, and "substituting `Differential Revision` for `Original Revision`" is the plain-english description of what we're trying to achieve.

Moving it into a one-liner is a -1 from me, it doesn't make anything clearer and makes the regex groups used more implicit instead of explicit. It removes a few lines, but disk space is cheap. ;)

zzzeid: Should have prefixed this with "nit", and I won't block the PR on it.

I think naming these groups without using `groupdict` doesn't utilize these named groups to their full potential given what you're using it for. Since the keys, group names, and f-string placeholders (i.e. variable names) all match, this seems repetitive and less readable. A possible middle ground:

```suggestion
        result = match.groupdict()
        return f"\nOriginal Revision: {result['phab_url']}/D{result['rev']}"
```

But again, this is probably splitting hairs at this point.

zzzeid (approved):

landoapi/uplift.py
```diff
@@ -83,25 +132,18 @@ def create_uplift_revision(
...
+    refs = {ref["type"]: ref for ref in phab.expect(source_diff, "fields", "refs")}
     base_revision = refs["base"]["identifier"] if "base" in refs else None
 
     # The first commit in the attachment list is the current HEAD of stack
     # we can use the HEAD to mark the changes being created
-    commits = phab.expect(diff, "attachments", "commits", "commits")
+    commits = phab.expect(source_diff, "attachments", "commits", "commits")
     head = commits[0] if commits else None
 
     # Upload it on target repo
```
zzzeid, line 149: The below code (i.e. the API call to `call_conduit`) may need to be modified so that `sourceControlSystem` is not hardcoded (currently always set to `hg` which may not always be the case). Other fields may need to be reviewed as well (e.g. `branch`). `sourcePath` may also need to be set (looks like there's a TODO there).

Lastly, `sourceControlBaseRevision` may not always point to a changeset that exists in the target repo. This may cause some problems when showing the diff on Phabricator, but also when generating the raw diff for landing. Just wanted to flag this in case it was not tested with this scenario.

landoapi/api/uplift.py
```diff
@@ -23,49 +24,89 @@
...
             )
+            new_uplift_phid = rev["revision_phid"]
+
+            # Store the returned revision information for the response.
+            output[new_uplift_phid] = rev
+
+            # Set `parent_phid` so the next created revision has this one set as the parent.
+            parent_phid = new_uplift_phid
+
     except Exception as e:
```
zzzeid, line 105: We should log a little more info here, especially since it's possible that there are stray revisions or diffs or other data that would have been created but not to completion. We should also abandon or mark any revisions that were created in the process so that they don't cause pollution. Ideally we do some clean up as well but at least all revisions/diffs that were partially created should be logged in the error.

This could also be refactored so that the transactions and API calls are collected and then called separately (e.g. outside the for loop) and updated afterwards with any relevant new info as needed, making error handling a little more robust.

zzzeid (changes requested):
A few comments and spots to clean up, but otherwise looks good.

landoapi/api/uplift.py
cgsheeh: Switched the implementation to maintain a stack of previously created revisions instead of just setting `parent_phid`. Then we get the previous parent with `stack[-1]["phid"]` and we can log the stack if we hit a failure creating a revision.

landoapi/uplift.py
```diff
@@ -126,11 +133,12 @@ def create_uplift_revision(
     """
     # Check the target repository needs an approval
     repos = get_repos_for_env(current_app.config.get("ENVIRONMENT"))
-    local_repo = repos.get(target_repository["fields"]["shortName"])
-    assert local_repo is not None, f"Unknown repository {target_repository}"
+    repo_shortname = phab.expect(target_repository, "fields", "shortName")
+    local_repo = repos.get(repo_shortname)
+    assert local_repo is not None, f"Unknown repository {repo_shortname}"
```
zzzeid, line 138: nit: Ideally we do if checks and raise here instead of using `assert`.

zzzeid (approved):
lgtm.

Note the punted items from the last review: RE: `sourceControlBaseRevision` and dealing with potentially incomplete stack more robustly.

landoapi/api/uplift.py
```diff
@@ -8,68 +8,140 @@
...
             extra={
-                "revision": data["revision_id"],
-                "repository": repository,
-                "error": str(e),
+                "revision": tip_revision_id,
+                "target_repository": repo_name,
             },
         )
-        raise
+        revision_data, target_repository = get_uplift_conduit_state(
```
zzzeid, line 73: I think `get_uplift_conduit_state` could be doing a little more when it comes to the `target_repo`. For example, it could do some of the work being done in `create_uplift_revision` so that we can fail sooner if there's something wrong with the repo (and so that we don't have to pass the dictionary around since we're only using it for this one purpose.)

Alternatively, we could also parse the target repo after this call, but before `create_uplift_revision`.

zzzeid (changes requested):
Few more comments.

landoapi/uplift.py
```diff
@@ -38,86 +95,90 @@ def get_release_managers(phab: PhabricatorClient) -> dict:
...
+    revision = phab.single(revision, "data", none_when_empty=True)
+    if not revision:
+        raise ValueError(f"No revision found with id {revision_id}")
+
+    try:
+        nodes, _ = build_stack_graph(phab, phab.expect(revision, "phid"))
+    except PhabricatorAPIException:
+        # If a revision within the stack causes an API exception, treat the whole stack
+        # as not found.
+        raise ValueError(f"Missing revision info for stack ending in {revision_id}")
```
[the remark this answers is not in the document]
cgsheeh: Hmmm so this section is mostly copy-pasted from `landoapi/api/stack.py:get`, but you're right that we should log that exception. I'll update it. Perhaps we should add some logging in the original source as well.

landoapi/uplift.py
```diff
@@ -130,42 +145,55 @@ def get_uplift_conduit_state(
...
         raise Exception("Missing raw source diff, cannot uplift revision.")
 
-    # Base revision hash is available on the diff fields
+    # Base revision hash is available on the diff fields.
     refs = {ref["type"]: ref for ref in phab.expect(source_diff, "fields", "refs")}
     base_revision = refs["base"]["identifier"] if "base" in refs else None
 
     # The first commit in the attachment list is the current HEAD of stack
-    # we can use the HEAD to mark the changes being created
+    # we can use the HEAD to mark the changes being created.
```
zzzeid, line 192: nit

```suggestion
    # We can use the HEAD to mark the changes being created.
```

zzzeid (approved):
lgtm

Note: https://github.com/mozilla-conduit/lando-api/pull/200#discussion_r871724598 and https://github.com/mozilla-conduit/lando-api/pull/200#pullrequestreview-980522482.

cgsheeh:
^^^ The only significant change should be that the `parents.set` transaction is now in its own call to `differential.revision.edit` - for the life of me I couldn't figure out how to get the revisions to stop associating with the Original Revisions so I just took the easy route and overwrote them after creation.

PR #201, opened 2022-04-26
requirements: generate via docker-compose service (bug-1766281)

zzzeid:


cgsheeh (changes requested):

docker-compose.yml
```diff
@@ -4,6 +4,11 @@
 
 version: '2'
 services:
+  build-requirements:
+    image: python:3.9
+    volumes:
+      - ./:/app
+    command: sh -c "cd /app && pip install pip --upgrade && pip install pip-tools && python -m piptools compile requirements.in --generate-hashes --allow-unsafe"
```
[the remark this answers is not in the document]
zzzeid: Opted to put a comment instead, since the full Dockerfile is quite bloated to be used for this purpose, but we can revisit in the future to optimize a little more.

docker-compose.yml
```diff
@@ -5,7 +5,7 @@
 version: '2'
 services:
   build-requirements:
-    image: python:3.9
+    image: python:3.9  # NOTE: this should be updated to match what is in ./Dockerfile
```
cgsheeh, line 8: I meant for you to add the comment to `Dockerfile`, since that is more likely to be updated without appropriately updating this file.

cgsheeh (changes requested):

docker-compose.yml
cgsheeh: Can't we just add the lines from below to this job?

```yaml
    build:
      context: ./
      dockerfile: ./Dockerfile-dev
```

zzzeid: Yeah but that builds the entire image which we don't need -- but we could still accomplish this without building the entire image by having a base target which is just the Python image.

cgsheeh: Ah, I misunderstood your comment - what you're suggesting is exactly what I was hoping for. :)

cgsheeh (approved):

PR #204, opened 2022-05-05
hg: add `--id` to `hg id` call so output can be used directly

cgsheeh:
Noticed the TODO mentioning we should use a template. Even
easier is to add `--id` to the call to `hg id` which will
cause only the 12-character changeset hash to be returned.


zzzeid (approved):
lgtm

PR #206, opened 2022-05-20
Revert "requirements: upgrades (bug 1766699) (#202)"

zzzeid:
This reverts commit 49aac4ceffe6aad38d90d347d7530fa8448a37ca.


cgsheeh (approved):
We should re-open the bug as well.

zzzeid:
> We should re-open the bug as well.

done.

PR #207, opened 2022-06-10
uplift: add bugherder functionality when landing to uplift repo (Bug 1773315)

cgsheeh:
Add some bug updating functionality when landing to an
`approval_required` (uplift) repository. Add an app config
to hold the `BMO_URL` so enable testing against dev and stage.
Add a new email notification function which is similar to the
landing failure email except with a standard error message.

### DIFF
--- a/landoapi/landing_worker.py
+++ b/landoapi/landing_worker.py
@@ -27,7 +27,10 @@ from landoapi.hg import (
 )
 from landoapi.models.landing_job import LandingJob, LandingJobStatus, LandingJobAction
 from landoapi.models.configuration import ConfigurationVariable
-from landoapi.notifications import notify_user_of_landing_failure
+from landoapi.notifications import (
+    notify_user_of_bug_update_failure,
+    notify_user_of_landing_failure,
+)
 from landoapi.repos import (
     Repo,
     repo_clone_subsystem,
@@ -36,6 +39,9 @@ from landoapi.storage import db, SQLAlchemy
 from landoapi.treestatus import (
     TreeStatus,
     treestatus_subsystem,
+)
+from landoapi.uplift import (
+    update_bugs_for_uplift,
 )
 
 logger = logging.getLogger(__name__)
@@ -244,6 +250,21 @@ class LandingWorker:
         )
 
     @staticmethod
+    def notify_user_of_bug_update_failure(job, exception):
+        """Wrapper around notify_user_of_landing_failure for convenience.
+
+        Args:
+            job (LandingJob): A LandingJob instance to use when fetching the
+                notification parameters.
+        """
+        notify_user_of_bug_update_failure(
+            job.requester_email,
+            job.head_revision,
+            f"Failed to update Bugzilla after landing uplift revisions: {str(exception)}",
+            job.id,
+        )
+
+    @staticmethod
     def extract_error_data(exception: str) -> tuple[list[str], list[str]]:
         """Extract rejected hunks and file paths from exception message."""
         # RE to capture .rej file paths.
@@ -435,9 +456,19 @@ class LandingWorker:
                     self.notify_user_of_landing_failure(job)
                     return False
 
-            commit_id = hgrepo.run_hg(["log", "-r", ".", "-T", "{node}"]).decode(
-                "utf-8"
-            )
+            # Gather changesets hashes and commit titles for each node in the stack in
+            # case we need to further analyse. Use the null character as the separator.
+            stack_ids: list[tuple[str, str]] = [
+                tuple(line.split("\0", 1))
+                for line in hgrepo.run_hg(
+                    ["log", "-r", "stack()", "-T", "{node}\\0{desc|firstline}\n"]
+                )
+                .decode("utf-8")
+                .splitlines()
+            ]
+
+            # Get the changeset hash of the first node.
+            commit_id = stack_ids[0][0]
 
             try:
                 hgrepo.push(repo.push_path, bookmark=repo.push_bookmark or None)
@@ -476,4 +507,16 @@ class LandingWorker:
 
         job.transition_status(LandingJobAction.LAND, commit_id=commit_id)
         db.session.commit()
+
+        if repo.approval_required:
+            changeset_titles = [title for _node, title in stack_ids]
+
+            try:
+                # If we just landed an uplift, update the relevant bugs as appropriate.
+                update_bugs_for_uplift(current_app.config["BMO_URL"], changeset_titles)
+            except Exception as e:
+                # The landing will have gone through even if updating the bugs fails. Notify
+                # the landing user so they are aware and can update the bugs themselves.
+                self.notify_user_of_bug_update_failure(job, e)
+
         return True
--- a/landoapi/notifications.py
+++ b/landoapi/notifications.py
@@ -5,7 +5,10 @@ import logging
 
 import kombu
 
-from landoapi.tasks import send_landing_failure_email
+from landoapi.tasks import (
+    send_bug_update_failure_email,
+    send_landing_failure_email,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


landoapi/uplift.py
```diff
@@ -226,29 +230,38 @@ def stack_uplift_form_submitted(stack_data) -> bool:
...
-        # Set the status of a bug to fixed if it's fix got uplifted to a branch
-        # and the "leave-open" keyword is not set
-        payload["status"] = "RESOLVED"
-        payload["resolution"] = "FIXED"
 
-    if "[checkin-needed-beta]" in bug["whiteboard"]:
-        # Remove "[checkin-needed-beta]" etc. texts from the whiteboard (rarely used, only
-        # if something doesn't need an uplift approval because it is not part of the build).
-        payload["whiteboard"] = bug["whiteboard"].replace("[checkin-needed-beta]", "")
+    milestone_tracking_flag = f"cf_status_firefox{milestone}"
```
Archaeopteryx, line 241: As discussed in the bug, ESR support is TBD.

landoapi/landing_worker.py
```diff
@@ -243,6 +249,21 @@ def notify_user_of_landing_failure(job):
             job.requester_email, job.head_revision, job.error, job.id
         )
 
+    @staticmethod
+    def notify_user_of_bug_update_failure(job, exception):
+        """Wrapper around notify_user_of_landing_failure for convenience.
```
zzzeid, line 254:
```suggestion
        """Wrapper around notify_user_of_bug_update_failure for convenience.
```

landoapi/landing_worker.py
```diff
@@ -435,9 +456,20 @@ def run_job(
...
+            # case we need to further analyse. Use the null character as the separator.
+            # This variable is a list of (node, changeset title) pairs.
+            stack_ids: list[tuple[str, str]] = [
+                tuple(line.split("\0", 1))
+                for line in hgrepo.run_hg(
+                    ["log", "-r", "stack()", "-T", "{node}\\0{desc|firstline}\n"]
+                )
+                .decode("utf-8")
+                .splitlines()
+            ]
```
zzzeid, line 469: I would suggest moving this to where it is used below (and perhaps not including `{node}` since it does not appear to be used). And, keeping the original `commit_id` definition here. Sure we would be running two separate commands, but this command is less relevant for the general code flow and the node IDs are mostly not needed, with the exception of the first one.

landoapi/uplift.py
```diff
@@ -213,10 +221,76 @@ def create_approval_request(phab: PhabricatorClient, revision: dict, form_conten
...
+
+    return payload
+
+
+UPDATE_RETRIES = 3
+
+
+def update_bugs_for_uplift(
+    bmo_url: str,
+    bmo_api_key: str,
```
zzzeid, line 260: Since we have `current_app` imported here, we don't really need to pass these through from the caller.

landoapi/landing_worker.py
```diff
@@ -476,4 +508,30 @@ def run_job(
...
+                hgrepo.run_hg(["cat" "-r", ".", "config/milestone.txt"])
+                .decode("utf-8")
+                .split(".")[0]
+            )
+
+            try:
+                # If we just landed an uplift, update the relevant bugs as appropriate.
+                update_bugs_for_uplift(
+                    current_app.config["BUGZILLA_URL"],
+                    current_app.config["BUGZILLA_API_KEY"],
```
zzzeid, line 527: These could be accessed directly from `update_bugs_for_uplift`.

landoapi/notifications.py
```diff
@@ -48,3 +51,43 @@ def notify_user_of_landing_failure(email, revision, error, request_id):
...
+
+def notify_user_of_bug_update_failure(email, revision, error, request_id):
+    """Send out user notifications that a bug update failed.
+
+    Args:
+        email (str): Receipent's email address (e.g. `requester_email`.)
+        revision (str): The revision associated with the bug update failure.
+        error (str): The text of the error associated with the bug update failure.
+        request_id (int): A unique identifier identifying either a legacy request ID or
+            a LandingJob ID.
```
zzzeid, line 64:
```suggestion
        request_id (int): A unique identifier identifying a LandingJob ID.
```

Legacy requests are fully deprecated, and just need to be ripped out of the code base.

landoapi/tasks.py
```diff
@@ -69,6 +69,61 @@ def send_landing_failure_email(recipient_email: str, revision_id: str, error_msg
...
+def send_bug_update_failure_email(
+    recipient_email: str, revision_id: str, error_msg: str
+):
+    """Tell a user that the Transplant service couldn't land their code.
+
+    Args:
+        recipient_email: The email of the user receiving the failure notification.
+        revision_id: The Phabricator Revision ID that failed to land. e.g. D12345
+        error_msg: The error message returned by the Transplant service.
+    """
```
zzzeid, line 99: This doc string needs to be updated (looks like it was copied/pasted?)

landoapi/uplift.py
```diff
@@ -213,10 +221,76 @@ def create_approval_request(phab: PhabricatorClient, revision: dict, form_conten
...
     # submitted.
     return any(
         get_uplift_request_form(revision) for revision in stack_data.revisions.values()
     )
+
+
+def create_uplift_bug_update_payload(
+    bug: dict, repo_name: str, milestone: int
+) -> dict[str, Any]:
+    """Create a payload for updating a bug using the BMO REST API."""
```
zzzeid, line 236: Can you elaborate a little in this docstring on the different things being done in this method. A quick summary of the various flags, keywords, etc. that are being used and manipulated should suffice.

landoapi/uplift.py
```diff
@@ -213,10 +221,76 @@ def create_approval_request(phab: PhabricatorClient, revision: dict, form_conten
...
+
+    checkin_needed_flag = f"[checkin-needed-{repo_name}]"
+    if checkin_needed_flag in bug["whiteboard"]:
+        # Remove "[checkin-needed-beta]" etc. texts from the whiteboard.
+        payload["whiteboard"] = bug["whiteboard"].replace(checkin_needed_flag, "")
+
+    return payload
+
+
+UPDATE_RETRIES = 3
```
zzzeid, line 255: This could be moved to the top of `update_bugs_for_uplift`, or even to the top of the module.

landoapi/uplift.py
```diff
@@ -213,10 +221,76 @@ def create_approval_request(phab: PhabricatorClient, revision: dict, form_conten
...
+    repo_name: str,
+    milestone: int,
+):
+    """Update Bugzilla bugs for uplift."""
+    headers = {"X-Bugzilla-API-Key": bmo_api_key}
+    bugs = [str(bug) for title in changeset_titles for bug in parse_bugs(title)]
+    params = {
+        "ids": ",".join(bugs),
+    }
+    bug_endpoint = f"{bmo_url}/rest/bug"
```
zzzeid, line 271: This should also be moved to the top of the method definition.

landoapi/uplift.py
```diff
@@ -213,10 +221,76 @@ def create_approval_request(phab: PhabricatorClient, revision: dict, form_conten
...
+    resp_get = requests.get(bug_endpoint, headers=headers, params=params)
```
zzzeid, line 272: Ideally we wrap these calls in a thin client -- something like `BMO.get` so that we only pass the endpoint and params here, and so that this is easier to reuse in the future. With that we reduce the clutter in this method (e.g. the bmo token/url/etc.).

landoapi/uplift.py
```diff
@@ -213,10 +221,76 @@ def create_approval_request(phab: PhabricatorClient, revision: dict, form_conten
...
+            except requests.RequestException as e:
+                if i == UPDATE_RETRIES:
+                    raise e
+
+                logger.exception(
+                    f"Error while updating bugs after uplift on attempt {i}, retrying..."
+                )
+                logger.exception(str(e))
+
+                time.sleep(0.1 * i)
```
zzzeid, line 296: Is this meant to be updated before merging this PR? These requests will still happen very quickly, even on the last attempt.

zzzeid (changes requested):
First pass with a few remarks. Might need a little more test coverage around the landing worker changes.

landoapi/landing_worker.py
cgsheeh: `node` is used in `commit_id` right below. I figured getting all the information about the stack from `hg log` would be the most efficient but running the command twice works too.

landoapi/uplift.py
```diff
@@ -252,45 +299,44 @@ def create_uplift_bug_update_payload(
...
 def update_bugs_for_uplift(
-    bmo_url: str,
-    bmo_api_key: str,
-    changeset_titles: list[str],
     repo_name: str,
-    milestone: int,
+    hgrepo: HgRepo,
+    changeset_titles: list[str],
 ):
     """Update Bugzilla bugs for uplift."""
```
zzzeid, line 307: More cleanly, I think we could pass the milestone file directly (or a string) from where we are calling this method.

It's not clear that this method needs to be run from within the `hgrepo.for_push` or `hgrepo.for_pull` context managers.

landoapi/uplift.py
```diff
@@ -252,45 +299,44 @@ def create_uplift_bug_update_payload(
...
-    bug_endpoint = f"{bmo_url}/rest/bug"
-    resp_get = requests.get(bug_endpoint, headers=headers, params=params)
-    resp_get.raise_for_status()
 
-    bugs = resp_get.json()["bugs"]
+    # Get information about the parsed bugs.
+    bugs = bmo.get_bug(params).json()["bugs"]
+
+    # Get the major release number from `config/milestone.txt`.
+    milestone = get_milestone_major_version(hgrepo)
```
zzzeid, line 322: As mentioned elsewhere, this method should take a file object or a string/bytes of the file.

zzzeid (changes requested):
Few more comments.

landoapi/uplift.py
```diff
@@ -252,45 +299,44 @@ def create_uplift_bug_update_payload(
...
 
     for bug in bugs:
         payload = create_uplift_bug_update_payload(bug, repo_name, milestone)
```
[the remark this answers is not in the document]
zzzeid: > Good call. Note that this won't actually do what you're suggesting since `map` returns an iterator, so we will still be making requests here before all the payloads are constructed. We would also need to create a lambda wrapper to fill in the other arguments to `create_uplift_bug_update_payload`. I've switched it to a list comprehension which will give us the desired behaviour.

Ah, I mentioned this earlier in my edit in parentheses -- by evaluate, I meant to wrap it in `list` for example, or do a list comprehension. Either way, would be good to fail before we start making calls

cgsheeh:
I took all your suggestions except for the BMO class - I don't think we need a class just for an API call when all the variables are passed from the caller or globally available. Trying to reduce the repeated sections of code for just two functions feels like overkill for right now. The module is still small enough that if we decide we need a class we can expand it easily at a later time. :+1:

zzzeid (approved):
lgtm

PR #208, opened 2022-07-05
typing: add more type hints to some functions (Bug 1759890)

cgsheeh:
Add more type hints to function definitions in Lando. Also
remove the unused variable from `calculate_review_extra_state`.

### DIFF
--- a/landoapi/auth.py
+++ b/landoapi/auth.py
@@ -1,11 +1,17 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 import functools
 import hashlib
 import hmac
 import logging
 import os
+
+from typing import (
+    Optional,
+    Union,
+)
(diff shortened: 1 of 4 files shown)


landoapi/reviews.py
```diff
@@ -19,8 +19,10 @@
 
 
 def calculate_review_extra_state(
-    for_diff_phid, reviewer_status, reviewer_diff_phid, reviewer_voided_phid
-):
+    for_diff_phid: str,
+    reviewer_status: ReviewerStatus,
```
zzzeid, line 23: nit: this is repeated in the docstring below, which can probably be removed/reworded below to reduce redundancy.

landoapi/auth.py
```diff
@@ -295,12 +301,15 @@ def email(self):
 
         return self._email
 
-    def is_in_groups(self, *args):
+    def is_in_groups(self, *args) -> bool:
```
zzzeid, line 304: nit: can probably add a type for `args` while we're in the neighbourhood.

zzzeid (approved):

PR #210, opened 2022-08-02
uplift: cache list of uplift repositories (Bug 1764213)

cgsheeh:
Decorates the call to `get_uplift_repositories` enable
caching the results in Redis. Moves the default cache
key timeout from `projects.py` to `cache.py` as it
makes more sense in that module.


landoapi/cache.py
```diff
@@ -10,6 +10,8 @@
 from landoapi.redis import SuppressRedisFailure
 from landoapi.systems import Subsystem
 
+DEFAULT_CACHE_KEY_TIMEOUT = 86400  # 60s * 60m * 24h
```
zzzeid, line 13: nit
```suggestion
DEFAULT_CACHE_KEY_TIMEOUT_SECONDS = 60 * 60 * 24
```

zzzeid (approved):
lgtm

PR #211, opened 2022-08-03
typing: add type hints in a few files (Bug 1759890)

cgsheeh:
Also removed a few functions that are unused from `commit_message.py`.

### DIFF
--- a/landoapi/commit_message.py
+++ b/landoapi/commit_message.py
@@ -188,34 +188,6 @@ def replace_reviewers(
         return commit_summary.strip() + "\n" + commit_description
 
 
-def strip_commit_metadata(message: str) -> str:
-    """Strips metadata related to commit tracking.
-
-    Will strip lines like "MozReview-Commit-ID: foo" from the commit
-    message.
-    """
-    # TODO this parsing is overly simplified. There is room to handle
-    # empty lines before the metadata.
-    lines = [l for l in message.splitlines() if not METADATA_RE.match(l)]
-
-    while lines and not lines[-1].strip():
-        lines.pop(-1)
-
-    return "\n".join(lines)
-
-
-def parse_commit_id(message: str) -> Optional[str]:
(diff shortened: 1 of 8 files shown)


landoapi/commit_message.py
```diff
@@ -185,35 +188,28 @@ def replace_first_reviewer(matchobj):
         return commit_summary.strip() + "\n" + commit_description
 
 
-def strip_commit_metadata(s):
+def strip_commit_metadata(message: str) -> str:
```
zzzeid, line 191: Is this used anywhere?

landoapi/commit_message.py
```diff
@@ -185,35 +188,28 @@ def replace_first_reviewer(matchobj):
...
-        joiner = "\n"
-    else:
-        raise TypeError("do not know type of commit message: %s" % type(s))
-
-    return joiner.join(lines)
+    return "\n".join(lines)
 
 
-def parse_commit_id(s):
+def parse_commit_id(message: str) -> Optional[str]:
```
zzzeid, line 207: Is this used anywhere?

landoapi/ui.py
```diff
@@ -12,7 +14,7 @@
 class LandoUISubsystem(Subsystem):
     name = "lando_ui"
 
-    def ready(self):
+    def ready(self) -> Union[bool, str]:
```
zzzeid, line 17: Could probably use the newer `bool | str` if you import from \_\_future\_\_.

zzzeid (changes requested):

landoapi/commit_message.py
cgsheeh: Hm, neither of the functions you called out are used... I wonder if that's something we could configure Flake8 to look for?

In the meantime I'll remove them.

zzzeid (approved):
lgtm. Final commit message may have to be edited to reflect removed functions.

PR #212, opened 2022-08-16
autoformat: change from hg fix to tip commit/amend strategy (Bug 1785004)

cgsheeh:
Remove the `hg fix` based autoformatting and replace with
running raw `mach lint` on the tip commit, creating a tip
autoformat commit for stacks of changes and amending the
commit for landings of a single changeset. Add a call to
`mach bootstrap` in the repo clone subsystem when autoformat
is enabled for a given repo, to enabled downloading required
packages. Since we now run `mach` from within the landing repo,
stop running `hg purge` with `--all` since that will remove
the existing objdir required for linting virtualenvs. Add the
`MOZBUILD_STATE_PATH` environment variable to the Dockerfile
to avoid a prompt during `mach bootstrap`.


landoapi/hg.py
```diff
@@ -368,56 +361,75 @@ def apply_patch(self, patch_io_buf):
...
-            if not key.endswith(":pattern"):
-                continue
+    def format_stack_tip(self, mach_path: Path) -> Optional[str]:
+        """Add an autoformat commit to the top of the patch stack."""
+        # Update to the tip of the stack.
+        self.run_hg(["update", "-r", "tip"])
 
-            fix_hg_command += ["--config", f"fix.{key}={value}"]
+        # Run linters.
+        subprocess.run([mach_path, "lint", "-r", "stack()"], check=True)
```
zzzeid, line 384: Wondering if it would be useful to store the command and its output, and put that in the commit message for verbosity?

landoapi/hg.py
```diff
@@ -368,56 +361,75 @@ def apply_patch(self, patch_io_buf):
...
+            # Create a new commit.
+            self.run_hg(
+                ["commit"]
+                + ["--message", AUTOFORMAT_COMMIT_MESSAGE]
+                + ["--landing_system", "lando"]
+            )
+        except hglib.error.CommandError as exc:
+            if exc.out.strip() == b"nothing changed":
+                # If nothing changed after formatting we can just return.
+                return None
```
zzzeid, line 396:
```suggestion
                return
```

landoapi/hg.py
```diff
@@ -368,56 +361,75 @@ def apply_patch(self, patch_io_buf):
...
+            return None
+
+        try:
+            # If formatting each commit has failed, create an autoformatting
+            # commit that fixes each changed file in the stack.
+            formatting_changeset = self.format_stack_tip(mach_path)
 
-        logger.info(f"revisions were reformatted: {', '.join(post_formatting_hashes)}")
+            if formatting_changeset is None:
+                return None
```
zzzeid, line 425:
```suggestion
                return
```

zzzeid (changes requested):
Couple of small comments.

landoapi/hg.py
cgsheeh: That's not a bad idea - we should definitely at least log the command output.

landoapi/hg.py
```diff
@@ -368,56 +372,158 @@ def apply_patch(self, patch_io_buf):
...
+        """Amend the top commit in the patch stack with changes from formatting."""
+        try:
+            # Amend the current commit, using `--no-edit` to keep the existing commit message.
+            self.run_hg(["commit", "--amend", "--no-edit", "--landing_system", "lando"])
+
+            return [self.get_current_node().decode("utf-8")]
+        except hglib.error.CommandError as exc:
+            if exc.out.strip() == b"nothing changed":
+                # If nothing changed after formatting we can just return.
+                return None
```
zzzeid, line 459: nit
```suggestion
                return
```

landoapi/hg.py
```diff
@@ -368,56 +372,158 @@ def apply_patch(self, patch_io_buf):
...
+            )
 
-        # Update the working directory to the latest change.
-        self.run_hg(["update", "-C", "-r", "tip"])
+            raise exc
 
-        # Exit if no revisions were reformatted.
-        pre_formatting_hashes = check_fix_output_for_replacements(fix_output)
-        if not pre_formatting_hashes:
+    def format_stack_amend(self) -> Optional[List[str]]:
```
zzzeid, line 449: Not very certain why we should treat 1-revision stacks any differently than multi-revision stacks.

landoapi/hg.py
```diff
@@ -368,56 +372,158 @@ def apply_patch(self, patch_io_buf):
...
+            output = self.run_code_formatters()
 
-        logger.info(f"revisions were reformatted: {', '.join(post_formatting_hashes)}")
+            # When the stack is just a single commit, amend changes into it.
+            if stack_size == 1:
+                return self.format_stack_amend()
 
-        return post_formatting_hashes
+            # If the stack is more than a single commit, create an autoformat commit.
+            return self.format_stack_tip(output)
```
zzzeid, line 518: These could be moved outside the try/except block or split into their own.

landoapi/repos.py
```diff
@@ -377,6 +368,10 @@ def ready(self):
                 logger.info("Cloning repo.", extra={"repo": name})
                 r.clone(repo.pull_path)
 
+            # Ensure packages required for automated code formatting are installed.
+            if repo.autoformat_enabled:
+                r.run_mach_bootstrap()
```
zzzeid, line 373: This my cause some problems if there are changes to `mach bootstrap` that need to be applied after the instance is started (since the `ready` method is only triggered once per restart).

landoapi/landing_worker.py
```diff
@@ -451,16 +451,16 @@ def run_job(
                     self.notify_user_of_landing_failure(job)
                     return True
 
-            # Run `hg fix` configured formatters if enabled
+            # Run automated code formatters if enabled.
             if repo.autoformat_enabled:
```
zzzeid, line 455: As discussed offline, this may be a good candidate to move to the Revision Worker (based on PR #224) in the future, and implement a way for the revision worker to store updated stacks for the landing worker to use.

zzzeid (changes requested):
Mostly small comments or minor changes.

landoapi/hg.py
```diff
@@ -280,7 +284,7 @@ def clean_repo(self, *, strip_non_public_commits=True):
         except hglib.error.CommandError:
             pass
         try:
-            self.run_hg(["purge", "--all"])
```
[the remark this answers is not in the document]
cgsheeh: We wanted the working copy to be clean after calling `clean_repo`, so I assume it was added since we didn't have a need for any VCS ignored files to be retained after cleaning. However we now need to keep at least the objdir and parts of `third_party` in central, so just running `purge` should be good enough for our purposes.

landoapi/hg.py
cgsheeh: Meh, personally I prefer to be explicit that `None` is going to be returned from a function, especially when the type is an `Optional`. I usually only ever do a generic `return` when the function itself would never return something. I'll change it in this case. :+1:

landoapi/hg.py
cgsheeh: Since we know formatting changes made to 1-revision stacks all apply to the revision, we can just amend them in. This avoids creating any autoformatting commits for the most common type of landings.

landoapi/repos.py
cgsheeh: For now I think we can run this at repo setup time - we may need to add a mechanism to periodically re-call `mach bootstrap` to keep toolchains up to date. I'd like to land these changes first before optimizing too much. :)

zzzeid (approved):
lgtm!

PR #213, opened 2022-08-23
transplants: add code freeze warning (Bug 1552371)

grahamhub:
this PR adds a new `RevisionWarningCheck` to compare the landing request time to the soft code freeze window, and if it falls inside that window it will surface a warning

### DIFF
--- a/landoapi/repos.py
+++ b/landoapi/repos.py
@@ -109,6 +109,11 @@ class Repo:
 
         return any(config.startswith("fix") for config in self.config_override.keys())
 
+    @property
+    def phab_identifier(self) -> str:
+        """Return a valid Phabricator identifier as a `str`."""
+        return self.short_name if self.short_name else self.tree
+
 
 SCM_ALLOW_DIRECT_PUSH = AccessGroup(
     active_group="active_scm_allow_direct_push",
--- a/landoapi/repos.py
+++ b/landoapi/repos.py
@@ -29,11 +29,6 @@ AccessGroup = namedtuple(
     ),
 )
 
-PRODUCT_DETAILS_LOCAL = "http://product-details.test"
-# TODO: update dev once json file is on merurcial server.
-PRODUCT_DETAILS_DEV = "https://product-details.mozilla.org"
-PRODUCT_DETAILS_PROD = "https://product-details.mozilla.org"
-
 
 @dataclass
 class Repo:
@@ -64,9 +59,6 @@ class Repo:
         config_override (dict): Parameters to override when loading the Mercurial
             configuration. The keys and values map directly to configuration keys and
             values. Defaults to `None`.
-        codefreeze_enabled (bool): Whether or not this repo has a code freeze window.
-            This is used to warn users when previewing a landing within the window.
-            Defaults to `False`.
         product_details_url (str): The URL which contains product-related information
             relevant to the repo. Defaults to an empty string.
     """
@@ -82,7 +74,6 @@ class Repo:
     approval_required: bool = False
     commit_flags: list[tuple[str, str]] = field(default_factory=list)
     config_override: dict = field(default_factory=dict)
-    codefreeze_enabled: bool = False
     product_details_url: str = ""
 
     def __post_init__(self):
@@ -181,8 +172,7 @@ REPO_CONFIG = {
             tree="test-repo",
             url="http://hg.test/test-repo",
             access_group=SCM_LEVEL_1,
-            codefreeze_enabled=True,
-            product_details_url=f"{PRODUCT_DETAILS_LOCAL}/1.0/firefox_versions.json"
+            product_details_url="http://product-details.test/1.0/firefox_versions.json",
         ),
         "first-repo": Repo(
             tree="first-repo",
@@ -228,8 +218,8 @@ REPO_CONFIG = {
             commit_flags=[DONTBUILD],
             config_override={"fix.black:command": "black -- -"},
             approval_required=True,
-            codefreeze_enabled=True,
-            product_details_url=f"{PRODUCT_DETAILS_DEV}/1.0/firefox_versions.json",
+            product_details_url="https://hg.mozilla.org/conduit-testing/m-c/raw-file"
+            "/tip/product-details/1.0/firefox_versions.json",
         ),
         "vct": Repo(
             tree="vct",
@@ -278,6 +268,8 @@ REPO_CONFIG = {
             short_name="mozilla-central",
             commit_flags=[DONTBUILD],
             config_override={"fix.black:command": "black -- -"},
+            product_details_url="https://product-details.mozilla.org"
+            "/1.0/firefox_versions.json",
         ),
         "comm-central": Repo(
             tree="comm-central",
--- a/landoapi/repos.py
+++ b/landoapi/repos.py
@@ -29,6 +29,10 @@ AccessGroup = namedtuple(
     ),
 )
 
+# Ideally we would have a difference here, will get in touch with service owners.
+PRODUCT_DETAILS_DEV = "https://product-details.mozilla.org"
+PRODUCT_DETAILS_PROD = "https://product-details.mozilla.org"
+
 
 @dataclass
 class Repo:
@@ -59,6 +63,11 @@ class Repo:
         config_override (dict): Parameters to override when loading the Mercurial
             configuration. The keys and values map directly to configuration keys and
             values. Defaults to `None`.
+        codefreeze_enabled (bool): Whether or not this repo has a code freeze window.
+            This is used to warn users when previewing a landing within the window.
+            Defaults to `False`.
+        product_details_url (str): The URL which contains product-related information
+            relevant to the repo. Defaults to an empty string.
     """
 
     tree: str
@@ -72,6 +81,8 @@ class Repo:
     approval_required: bool = False
     commit_flags: list[tuple[str, str]] = field(default_factory=list)
     config_override: dict = field(default_factory=dict)
+    codefreeze_enabled: bool = False
+    product_details_url: str = ""
 
     def __post_init__(self):
         """Set defaults based on initial values.
@@ -207,6 +218,8 @@ REPO_CONFIG = {
             commit_flags=[DONTBUILD],
             config_override={"fix.black:command": "black -- -"},
             approval_required=True,
+            codefreeze_enabled=True,
+            product_details_url=f"{PRODUCT_DETAILS_DEV}/1.0/firefox_versions.json",
         ),
         "vct": Repo(
             tree="vct",
--- a/landoapi/repos.py
+++ b/landoapi/repos.py
@@ -218,8 +218,8 @@ REPO_CONFIG = {
             commit_flags=[DONTBUILD],
             config_override={"fix.black:command": "black -- -"},
             approval_required=True,
-            product_details_url="https://hg.mozilla.org/conduit-testing/m-c/raw-file"
-            "/tip/product-details/1.0/firefox_versions.json",
+            product_details_url="https://raw.githubusercontent.com/mozilla-conduit"
+            "/suite/main/docker/product-details/1.0/firefox_versions.json",
         ),
         "vct": Repo(
             tree="vct",
--- a/landoapi/repos.py
+++ b/landoapi/repos.py
@@ -29,7 +29,8 @@ AccessGroup = namedtuple(
     ),
 )
 
-# Ideally we would have a difference here, will get in touch with service owners.
+PRODUCT_DETAILS_LOCAL = "http://product-details.test"
+# TODO: update dev once json file is on merurcial server.
 PRODUCT_DETAILS_DEV = "https://product-details.mozilla.org"
 PRODUCT_DETAILS_PROD = "https://product-details.mozilla.org"
 
@@ -172,7 +173,11 @@ REPO_CONFIG = {
     "default": {},
     "localdev": {
         "test-repo": Repo(
-            tree="test-repo", url="http://hg.test/test-repo", access_group=SCM_LEVEL_1
+            tree="test-repo",
+            url="http://hg.test/test-repo",
+            access_group=SCM_LEVEL_1,
+            codefreeze_enabled=True,
+            product_details_url=f"{PRODUCT_DETAILS_LOCAL}/1.0/firefox_versions.json"
         ),
         "first-repo": Repo(
             tree="first-repo",
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -5,7 +5,9 @@ import functools
 import hashlib
 import json
 import logging
+import requests
 from collections import namedtuple
+from datetime import datetime
 
 from connexion import ProblemException
 
@@ -301,6 +303,33 @@ def warning_wip_commit_message(*, revision, **kwargs):
         return "This revision is marked as a WIP. Please remove `WIP:` before landing."
 
 
+@RevisionWarningCheck(8, "Repository is under a soft code freeze.", True)
+def warning_code_freeze(*, repo, **kwargs):
+    repo_fields = repo.get("fields", {})
+    short_name = repo_fields.get("shortName")
+    if short_name == "mozilla-central":
+        product_details = requests.get(
+            "https://product-details.mozilla.org/1.0/firefox_versions.json"
+        ).json()
+
+        freeze_date = datetime.strptime(
+            product_details.get("NEXT_SOFTFREEZE_DATE"),
+            "%Y-%m-%d",
+        )
+        merge_date = datetime.strptime(
+            product_details.get("NEXT_MERGE_DATE"),
+            "%Y-%m-%d",
+        )
+
+        warning = {
+            "message": f"Repository is under a soft code freeze "
+            f"(ends {merge_date.strftime('%Y-%m-%d')})."
+        }
+
+        if freeze_date <= datetime.today() <= merge_date:
+            return [warning]
+
+
 def user_block_no_auth0_email(*, auth0_user, **kwargs):
     """Check the user has a proper auth0 email."""
     return (
@@ -345,6 +374,7 @@ def check_landing_warnings(
         warning_revision_missing_testing_tag,
         warning_diff_warning,
         warning_wip_commit_message,
+        warning_code_freeze,
     ],
 ):
     assessment = TransplantAssessment()
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -308,26 +308,36 @@ def warning_code_freeze(*, repo, **kwargs):
     repo_fields = repo.get("fields", {})
     short_name = repo_fields.get("shortName")
     if short_name == "mozilla-central":
-        product_details = requests.get(
-            "https://product-details.mozilla.org/1.0/firefox_versions.json"
-        ).json()
+        message = ""
+        try:
+            product_details = requests.get(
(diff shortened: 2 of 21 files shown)


landoapi/transplants.py
```diff
@@ -301,6 +303,33 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
 
 
+@RevisionWarningCheck(8, "Repository is under a soft code freeze.", True)
+def warning_code_freeze(*, repo, **kwargs):
+    repo_fields = repo.get("fields", {})
+    short_name = repo_fields.get("shortName")
+    if short_name == "mozilla-central":
+        product_details = requests.get(
+            "https://product-details.mozilla.org/1.0/firefox_versions.json"
+        ).json()
```
zzzeid, line 313: Good to wrap this with a try/except in case this service is down. Caught error can alter the warning message to say that soft freeze state could not be checked, for example.

landoapi/transplants.py
```diff
@@ -301,6 +303,33 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
+        )
+        merge_date = datetime.strptime(
+            product_details.get("NEXT_MERGE_DATE"),
+            "%Y-%m-%d",
+        )
+
+        warning = {
+            "message": f"Repository is under a soft code freeze "
+            f"(ends {merge_date.strftime('%Y-%m-%d')})."
+        }
```
zzzeid, line 327: This should be moved to the if statement below so that it does not run needlessly if the below condition is not met.

landoapi/transplants.py
```diff
@@ -301,6 +303,33 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
+
+        if freeze_date <= datetime.today() <= merge_date:
+            return [warning]
```
zzzeid, line 330: Do we know what timezone these dates are in, and how they compare to the server timezone? Would be good to double check this and leave a comment or two.

landoapi/transplants.py
```diff
@@ -301,6 +303,33 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
+@RevisionWarningCheck(8, "Repository is under a soft code freeze.", True)
+def warning_code_freeze(*, repo, **kwargs):
+    repo_fields = repo.get("fields", {})
+    short_name = repo_fields.get("shortName")
+    if short_name == "mozilla-central":
+        product_details = requests.get(
+            "https://product-details.mozilla.org/1.0/firefox_versions.json"
+        ).json()
+
+        freeze_date = datetime.strptime(
```
zzzeid, line 315: Can probably return early if `freeze_date` is in the future.

zzzeid (changes requested):
I know this is still in draft, but quick drive-by comments. Probably good to ensure full test coverage with the different time conditions.

landoapi/transplants.py
```diff
@@ -301,6 +303,43 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
+            today = datetime.now(tz=timezone.utc)
+            freeze_date = datetime.strptime(
+                f"{product_details.get('NEXT_SOFTFREEZE_DATE')} -0800",
+                "%Y-%m-%d %z",
+            ).replace(tzinfo=timezone.utc)
+            if today < freeze_date:
+                return
+
+            merge_date = datetime.strptime(
+                f"{product_details.get('NEXT_MERGE_DATE')} -0800",
```
cgsheeh, line 326: Is `-0800` relevant in some way? Maybe that should be a variable somewhere and added in as part of the format string.

cgsheeh (changes requested):

landoapi/transplants.py
grahamhub: forgot to add a comment in the code - the freeze dates are based on pacific timezone (utc offset of -0800), so this is included to be correctly converted to utc. but agreed it'll look better with a variable

landoapi/transplants.py
```diff
@@ -301,6 +303,43 @@ def warning_wip_commit_message(*, revision, **kwargs):
         return "This revision is marked as a WIP. Please remove `WIP:` before landing."
 
 
+@RevisionWarningCheck(8, "Repository is under a soft code freeze.", True)
+def warning_code_freeze(*, repo, **kwargs):
+    repo_fields = repo.get("fields", {})
+    short_name = repo_fields.get("shortName")
+    if short_name == "mozilla-central":
```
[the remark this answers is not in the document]
grahamhub: based on the conversation in the original bug, it seems central is the only (relevant?) repo that undergoes a code freeze

landoapi/transplants.py
```diff
@@ -301,6 +303,47 @@ def warning_wip_commit_message(*, revision, **kwargs):
         return "This revision is marked as a WIP. Please remove `WIP:` before landing."
 
 
+@RevisionWarningCheck(8, "Repository is under a soft code freeze.", True)
+def warning_code_freeze(*, repo, **kwargs):
+    repo_fields = repo.get("fields", {})
+    short_name = repo_fields.get("shortName")
```
zzzeid, line 309: Slightly cleaner:
```suggestion
    try:
        short_name = repo_fields["fields"]["shortName"]
    except KeyError:
        return
```

Or:
```suggestion
    if "fields" in repo_fields and "shortName" in repo_fields["fields"]:
        short_name = repo_fields["fields"]["shortName"]
    else:
        return
```

Although both those keys should always exist here, so I'm tempted to forego this prudence. Lots of things will break if `shortName` is missing from the repo. You should also consider using `PhabricatorClient.expect` which takes care of this.

landoapi/transplants.py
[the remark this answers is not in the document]
cgsheeh: We still shouldn't check the value of `shortName` to determine whether this check is relevant. We should use `shortName` to get the `Repo` object for the current env using something like this: https://github.com/mozilla-conduit/lando-api/blob/04a1f37078c2206d92cb06fa206542dd2a2ea8d7/landoapi/api/uplift.py#L48

Then check a flag on the object similar to just below: https://github.com/mozilla-conduit/lando-api/blob/04a1f37078c2206d92cb06fa206542dd2a2ea8d7/landoapi/api/uplift.py#L58

Then it will look like:

```python
all_repos = get_repos_for_env(current_app.config.get("ENVIRONMENT")
short_name = repo.get("fields", {}).get("shortName")
local_repo = all_repos.get(short_name)

if not local_repo.soft_code_freeze:
    return

# rest of the code here
```

landoapi/transplants.py
```diff
@@ -301,6 +303,47 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
+    if short_name == "mozilla-central":
+        message = ""
+        try:
+            product_details = requests.get(
+                "https://product-details.mozilla.org/1.0/firefox_versions.json"
+            ).json()
```
zzzeid, line 315: This value should probably be also stored in the `Repo` object -- in case we want to apply different product URLs to different repos/products (e.g. [comm-central] (https://product-details.mozilla.org/1.0/thunderbird.json) or different URLs in DEV/PROD/local.

landoapi/transplants.py
```diff
@@ -301,6 +303,47 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
+
+            if "NEXT_SOFTFREEZE_DATE" not in product_details:
+                raise KeyError
```
zzzeid, line 318: This is redundant, as `product_details["NEXT_SOFTFREEZE_DATE"]` will raise `KeyError` if the key does not exist. Either return a different value or handle this differently.

landoapi/transplants.py
```diff
@@ -301,6 +303,47 @@ def warning_wip_commit_message(*, revision, **kwargs):
         return "This revision is marked as a WIP. Please remove `WIP:` before landing."
 
 
+@RevisionWarningCheck(8, "Repository is under a soft code freeze.", True)
+def warning_code_freeze(*, repo, **kwargs):
+    repo_fields = repo.get("fields", {})
+    short_name = repo_fields.get("shortName")
+    if short_name == "mozilla-central":
+        message = ""
+        try:
```
zzzeid, line 312: This `try` block is too large. See [pep 8](https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations).

> Additionally, for all try/except clauses, limit the try clause to the absolute minimum amount of code necessary. Again, this avoids masking bugs.

landoapi/transplants.py
```diff
@@ -301,6 +303,47 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
+            ).json()
+
+            if "NEXT_SOFTFREEZE_DATE" not in product_details:
+                raise KeyError
+
+            # The code freeze dates generally correspond to PST work days.
+            utc_offset = "-0800"
+            today = datetime.now(tz=timezone.utc)
+            freeze_date = datetime.strptime(
+                f"{product_details.get('NEXT_SOFTFREEZE_DATE')} {utc_offset}",
```
zzzeid, line 324: No need for `.get` here and below, we should be checking the proper format of these keys before generating the message.

landoapi/transplants.py
```diff
@@ -301,6 +303,47 @@ def warning_wip_commit_message(*, revision, **kwargs):
         return "This revision is marked as a WIP. Please remove `WIP:` before landing."
 
 
+@RevisionWarningCheck(8, "Repository is under a soft code freeze.", True)
+def warning_code_freeze(*, repo, **kwargs):
+    repo_fields = repo.get("fields", {})
+    short_name = repo_fields.get("shortName")
+    if short_name == "mozilla-central":
+        message = ""
```
zzzeid, line 311: `message` should not be blank -- if there is an unhandled case, the message should be explicit about that. Cleaning up / refactoring the try/except block should make it clearer as to what is happening with `message`.

zzzeid (changes requested):

cgsheeh (changes requested):

landoapi/transplants.py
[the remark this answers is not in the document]
zzzeid: > We still shouldn't check the value of `shortName` to determine whether this check is relevant. We should use `shortName` to get the `Repo` object for the current env.

+1, and would also help with setting a URL for our testing repos (see [this comment](https://github.com/mozilla-conduit/lando-api/pull/213/files#r958801182))

landoapi/repos.py
```diff
@@ -72,6 +82,8 @@ class Repo:
     approval_required: bool = False
     commit_flags: list[tuple[str, str]] = field(default_factory=list)
     config_override: dict = field(default_factory=dict)
+    codefreeze_enabled: bool = False
+    product_details_url: str = ""
```
cgsheeh, line 86: Will `codefreeze_enabled` ever be `True` without `product_details_url` being set to a non-empty string? Maybe we could get rid of the boolean and just do `if repo.product_details_url` instead.

cgsheeh (changes requested):

landoapi/repos.py
grahamhub: in this situation no, but if we were to use this url for other repos in the future, then it wouldn't be a reliable way to tell. however, we could just check for the `NEXT_CODEFREEZE_DATE` key in the returned json for the given product

landoapi/repos.py
```diff
@@ -29,6 +29,11 @@
     ),
 )
 
+PRODUCT_DETAILS_LOCAL = "http://product-details.test"
+# TODO: update dev once json file is on merurcial server.
+PRODUCT_DETAILS_DEV = "https://product-details.mozilla.org"
+PRODUCT_DETAILS_PROD = "https://product-details.mozilla.org"
```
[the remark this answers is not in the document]
grahamhub: after talking with zeid, decided to just hardcode the URLs per repo since it's just the base URL that's different and each repo will have a different URL regardless

landoapi/transplants.py
```diff
@@ -301,6 +304,56 @@ def warning_wip_commit_message(*, revision, **kwargs):
...
+
+        if freeze_date <= today <= merge_date:
+            return [
+                {
+                    "message": (
+                        f"Repository is under a soft code freeze "
+                        f"(ends {merge_date_str})."
+                    )
+                }
+            ]
```
zzzeid, line 354: slight cleanup here to reduce nesting and make sure not to return an incorrect message if a repo does not implement the product details URL:

```suggestion
    if not repo_details.product_details_url:
        # Repo does not have a product details URL.
        return

    try:
        product_details = requests.get(repo_details.product_details_url).json()
    except requests.exceptions.RequestException as e:
        logger.exception(e)
        return [{"message": "Could not retrieve repository's code freeze status."}]

    # If the JSON doesn't have these keys, this warning isn't applicable
    try:
        freeze_date_str = product_details["NEXT_SOFTFREEZE_DATE"]
        merge_date_str = product_details["NEXT_MERGE_DATE"]
    except KeyError:
        return

    # The code freeze dates generally correspond to PST work days.
    utc_offset = "-0800"
    today = datetime.now(tz=timezone.utc)
    freeze_date = datetime.strptime(
        f"{freeze_date_str} {utc_offset}",
        "%Y-%m-%d %z",
    ).replace(tzinfo=timezone.utc)
    if today < freeze_date:
        return

    merge_date = datetime.strptime(
        f"{merge_date_str} {utc_offset}",
        "%Y-%m-%d %z",
    ).replace(tzinfo=timezone.utc)

    if freeze_date <= today <= merge_date:
        return [
            {
                "message": (
                    f"Repository is under a soft code freeze "
                    f"(ends {merge_date_str})."
                )
            }
        ]                                                                                                                                                                                  

```

tests/test_transplants.py
```diff
@@ -143,6 +143,114 @@ def test_dryrun_reviewers_warns(client, db, phabdouble, auth0_mock):
...
+        },
+    )
+    monkeypatch.setattr("landoapi.transplants.datetime", codefreeze_datetime())
+    mc_repo = Repo(
+        tree="mozilla-conduit",
+        url="https://hg.test/mozilla-conduit",
+        access_group=SCM_CONDUIT,
+        commit_flags=[DONTBUILD],
+        config_override={"fix.black:command": "black -- -"},
+        approval_required=True,
```
zzzeid, line 171: Likely don't need these two things in this test.

zzzeid (changes requested):

landoapi/transplants.py
```diff
@@ -312,46 +312,47 @@ def warning_code_freeze(*, repo, **kwargs):
...
+    except requests.exceptions.RequestException as e:
+        logger.exception(e)
+        return [{"message": "Could not retrieve repository's code freeze status."}]
+
+    # If the JSON doesn't have these keys, this warning isn't applicable
+    try:
+        freeze_date_str = product_details["NEXT_SOFTFREEZE_DATE"]
+        merge_date_str = product_details["NEXT_MERGE_DATE"]
+    except KeyError:
+        return
```
zzzeid, line 330: At this point in the refactor, I think this would be a little cleaner and more explicit, as well as covers the case when the input is null/blank (if that's ever the case):

```suggestion
freeze_date_str = product_details.get("NEXT_SOFTFREEZE_DATE")
merge_date_str = product_details.get("NEXT_MERGE_DATE")

if not freeze_date_str or not merge_date_str:
    return
```

tests/conftest.py
```diff
@@ -428,3 +429,44 @@ def hg_clone(hg_server, tmpdir):
...
+    )
+
+
+@pytest.fixture
+def codefreeze_datetime(request_mocker):
+    today = datetime(2000, 1, 5, 0, 0, 0, tzinfo=timezone.utc)
+    freeze_date = datetime(2000, 1, 3, 0, 0, 0)
+    merge_date = datetime(2000, 1, 6, 0, 0, 0)
+    outside_freeze_date = datetime(2000, 2, 3, 0, 0, 0)
+    outside_merge_date = datetime(2000, 2, 6, 0, 0, 0)
```
zzzeid, line 452: nit

```suggestion
    today = datetime(2000, 1, 5, 0, 0, 0, tzinfo=timezone.utc)
    two_days_ago = datetime(2000, 1, 3, 0, 0, 0)
    tomorrow = datetime(2000, 1, 6, 0, 0, 0)
    one_month_from_today = datetime(2000, 2, 3, 0, 0, 0)
    one_month_and_two_days_from_today = datetime(2000, 2, 6, 0, 0, 0)
```

Maybe the last two, if they don't need to be in this exact interval can be made to be `four_weeks_from_today` and `five_weeks_from_today` or something to simplify the wording a little more.

tests/conftest.py
```diff
@@ -428,3 +429,44 @@ def hg_clone(hg_server, tmpdir):
...
+
+        @classmethod
+        def strptime(cls, date_string, fmt):
+            if date_string == "freeze_date -0800":
+                return freeze_date
+            elif date_string == "merge_date -0800":
+                return merge_date
+            elif date_string == "outside_freeze_date -0800":
+                return outside_freeze_date
+            elif date_string == "outside_merge_date -0800":
```
zzzeid, line 467: nit (to match what you have in `transplants.warning_code_freeze`)

```suggestion
            utc_offset = "-0800"
            if date_string == f"freeze_date {utc_offset}":
                return freeze_date
            elif date_string == f"merge_date {utc_offset}":
                return merge_date
            elif date_string == f"outside_freeze_date {utc_offset}":
                return outside_freeze_date
            elif date_string == f"outside_merge_date {utc_offset}":
```

zzzeid (approved):
Couple of nits / nice to haves, otherwise lgtm.

tests/test_transplants.py
```diff
@@ -143,6 +143,110 @@ def test_dryrun_reviewers_warns(client, db, phabdouble, auth0_mock):
...
+    monkeypatch,
+    request_mocker,
+):
+    product_details = "https://product-details.mozilla.org/1.0/firefox_versions.json"
+    request_mocker.register_uri(
+        "GET",
+        product_details,
+        json={
+            "NEXT_SOFTFREEZE_DATE": "freeze_date",
+            "NEXT_MERGE_DATE": "merge_date",
```
zzzeid, line 161: If these are renamed to reflect the day relative to "today" (from the test's perspective), it will be easier to understand what this test is trying to do (i.e. in this case NEXT_SOFTFREEZE_DATE would say "two_days_ago", and "NEXT_MERGE_DATE" would say "tomorrow".)

grahamhub: thank you for the re-work - i was trying to think of a way to make it more straightforward

tests/test_transplants.py
```diff
@@ -143,6 +143,110 @@ def test_dryrun_reviewers_warns(client, db, phabdouble, auth0_mock):
...
+            ]
+        },
+        headers=auth0_mock.mock_headers,
+    )
+
+    assert 200 == response.status_code
+    assert "application/json" == response.content_type
+    assert response.json["warnings"]
+    assert response.json["warnings"][0]["id"] == 8
+    assert response.json["confirmation_token"] is not None
```
cgsheeh, line 196: When you write `assert`s you should include a statement describing what the assert is specifically trying to test, ie `assert response.json["warnings"], "warnings should be non-empty for a repo with a code freeze"`. This makes updating/removing/fixing tests easier in the future since you've added the relevant context to the test itself, so another dev can decide if the test needs to be fixed/can be removed etc. Please do this as much as possible when you write tests.

As an example, what is `assert response.json["warnings"][0]["id"] == 8` testing? If I update tests and the number `8` becomes `9` is that a failure? Adding context like `"ID of the returned warning should match the ID for the code freeze warning"` makes this much easier to understand.

cgsheeh (changes requested):

PR #214, opened 2022-08-26
stacks: update build_stack_graph to use new endpoint (bug 1787516)

zzzeid:
- use updated response from differential.revision.search endpoint
- translate `stackGraph` field to phids and edges
- update `tests.mocks` such that revision response includes new field

### DIFF
--- a/landoapi/stacks.py
+++ b/landoapi/stacks.py
@@ -27,10 +27,6 @@ def build_stack_graph(
 ) -> Tuple[Set[str], Set[Tuple[str, str]]]:
     """Return a graph representation of a revision stack.
 
-    This function is expensive and can make up to approximately
-    n/2 calls to phabricator for a linear stack where n is the
-    number of revisions in the stack.
-
     Args:
         phab: A landoapi.phabricator.PhabricatorClient.
         revision_phid: String PHID of a revision in the stack.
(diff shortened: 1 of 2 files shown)


tests/mocks.py
```diff
@@ -96,6 +97,50 @@ def validate_change(change):
...
+    # from child to parent. This is enough to represent the graph.
+    edges = {
+        (edge["sourcePHID"], edge["destinationPHID"])
+        for edge in edges
+        if edge["edgeType"] == "revision.parent"
+    }
+
+    depends_on = defaultdict(list)
+    for edge in edges:
+        # Check that destination/source depend on a node.
```
zzzeid, line 133:
```suggestion
        # Check that destination phid has a corresponding source phid.
```

landoapi/stacks.py
```diff
@@ -42,35 +38,21 @@ def build_stack_graph(
...
     """
-    phids = set()
-    new_phids = {revision_phid}
+    result = phab.call_conduit(
+        "differential.revision.search",
+        constraints={"phids": [revision_phid]},
+        limit=1,
+    )
+
+    stack_graph = result["data"][0]["fields"]["stackGraph"]
```
cgsheeh, line 47: This should use `phab.single` and `phab.expect`.

landoapi/stacks.py
```diff
@@ -42,35 +38,21 @@ def build_stack_graph(
...
-    edges = {
-        (edge["sourcePHID"], edge["destinationPHID"])
-        for edge in edges
-        if edge["edgeType"] == "revision.parent"
-    }
+    for source_node, destination_nodes in stack_graph.items():
+        for predecessor_node in destination_nodes:
+            edges.append((source_node, predecessor_node))
+
+    edges = set(edges)
```
cgsheeh, line 55: We can just make `edges` a set to begin with and `.add` the nodes instead of converting from a list.

tests/mocks.py
```diff
@@ -96,6 +97,50 @@ def validate_change(change):
...
+    # Treat the stack like a commit DAG, we only care about edges going
+    # from child to parent. This is enough to represent the graph.
+    edges = {
+        (edge["sourcePHID"], edge["destinationPHID"])
+        for edge in edges
+        if edge["edgeType"] == "revision.parent"
+    }
+
+    stack_graph = defaultdict(list)
+    for edge in edges:
```
cgsheeh, line 132: Here we're always using `edge` with some index - can we just unpack it in the loop so it's more obvious what `edge[0]` and `edge[1]` are supposed to represent?

cgsheeh (changes requested):

landoapi/stacks.py
zzzeid: Done.

landoapi/stacks.py
zzzeid: Good idea.

tests/mocks.py
zzzeid: Done, + added a `sources` list to make checking for root node a little more readable.

cgsheeh (approved):

PR #215, opened 2022-08-30
uplift: use `packaging.version.Version` to handle milestone version parsing (Bug 1788078)

cgsheeh:
The regex used to parse the major milestone version failed
on the content of the current beta (which is now in the tree
as a test case as `MILESTONE_TEST_CONTENTS_2`). This commit
changes the parsing to instead take the last line of the file
and pass to a `Version` object, which is returned from the
parsing function.

### DIFF
--- a/tests/test_uplift.py
+++ b/tests/test_uplift.py
@@ -4,15 +4,19 @@
 
 import pytest
 
+from packaging.version import (
+    Version,
+)
+
 from landoapi.phabricator import PhabricatorClient
 from landoapi.stacks import build_stack_graph
 from landoapi.uplift import (
     create_uplift_bug_update_payload,
     move_drev_to_original,
-    parse_milestone_major_version,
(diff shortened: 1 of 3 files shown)


tests/test_uplift.py
```diff
@@ -28,11 +32,31 @@
...
+
+105.0
+"""
+
 
-def test_parse_milestone_major_version():
-    assert (
-        parse_milestone_major_version(MILESTONE_TEST_CONTENTS) == 84
-    ), "Test milestone file should have 84 as major milestone version."
+def test_parse_milestone_version():
```
zzzeid, line 52: Can test coverage be added to cover the not so happy path of `InvalidVersion`/`ValueError`?

zzzeid (changes requested):

zzzeid (approved):
lgtm

PR #216, opened 2022-09-01
phabricator: call `diffusion.looksoon` on landing to trigger repo sync (Bug 1455468, Bug 1788722)

cgsheeh:
Adds a call to Phabricator's `diffusion.looksoon` API to trigger
syncing from the upstream repo. This should cause revisions
to close faster, especially on uplift repositories.

As part of this patch some tests were failing due to out of date
`moto` conflicting with `responses`. Upgrading `moto` caused the
removal of several packages that are used directly in testing and
the main application. These packages are added to `requirements.in`
so they are explicit dependencies.

### DIFF
--- a/landoapi/landing_worker.py
+++ b/landoapi/landing_worker.py
@@ -13,6 +13,7 @@ import subprocess
 import time
 
 import hglib
+import kombu
 from flask import current_app
 
 from landoapi import patches
@@ -31,12 +32,12 @@ from landoapi.notifications import (
     notify_user_of_bug_update_failure,
     notify_user_of_landing_failure,
 )
-from landoapi.phabricator import PhabricatorAPIException, PhabricatorClient
 from landoapi.repos import (
     Repo,
     repo_clone_subsystem,
 )
 from landoapi.storage import db, SQLAlchemy
+from landoapi.tasks import phab_trigger_repo_update
 from landoapi.treestatus import (
     TreeStatus,
     treestatus_subsystem,
@@ -264,6 +265,23 @@ class LandingWorker:
             f"Failed to update Bugzilla after landing uplift revisions: {str(exception)}",
             job.id,
         )
+
+    @staticmethod
+    def phab_trigger_repo_update(phab_identifier: str):
+        """Wrapper around `phab_trigger_repo_update` for convenience.
+
+        Args:
+            phab_identifier: `str` to be passed to Phabricator to identify
+            repo.
+        """
+        try:
+            # Send a Phab repo update task to Celery.
+            phab_trigger_repo_update.apply_async(args=(phab_identifier,))
+        except kombu.exceptions.OperationalError as e:
+            # Log the exception but continue gracefully.
+            # The repo will eventually update.
+            logger.exception("Failed sending repo update task to Celery.")
+            logger.exception(e)
 
     @staticmethod
     def extract_error_data(exception: str) -> tuple[list[str], list[str]]:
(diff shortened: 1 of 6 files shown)


landoapi/landing_worker.py
```diff
@@ -521,4 +522,17 @@ def run_job(
...
+                current_app.config["PHABRICATOR_URL"],
+                current_app.config["PHABRICATOR_UNPRIVILEGED_API_KEY"],
+            )
+            phab.call_conduit("diffusion.looksoon", repositories=[repo.phab_identifier])
+
+        except PhabricatorAPIException as e:
+            # Log the exception but continue gracefully.
+            logger.exception("Failed calling `diffusion.looksoon` to update Phab.")
+            logger.exception(e)
+
```
zzzeid, line 537: Couple of things to note here:

- This will be triggered quite frequently (maybe more than needed) when the landing worker is under a heavy load. What kind of performance impact is it going to have on Phabricator?
- This should be moved to an async command (similar to `admin_remove_phab_project`) so that it doesn't interfere with the Landing Worker
- Perhaps we can throttle the task somehow to make sure it isn't called after every single landing.

zzzeid (changes requested):

landoapi/landing_worker.py
cgsheeh: > Couple of things to note here:
> 
>     * This will be triggered quite frequently (maybe more than needed) when the landing worker is under a heavy load. What kind of performance impact is it going to have on Phabricator?

Very little - the call itself is lightweight as it just passes the callsign to Phab, and Phab schedules a repo update sooner than the standard update frequency.

>     * This should be moved to an async command (similar to `admin_remove_phab_project`) so that it doesn't interfere with the Landing Worker

Makes sense, I'll do that.

>     * Perhaps we can throttle the task somehow to make sure it isn't called after every single landing.

I think that's mostly the point, we want the repo to update ASAP after landings take place. Do we know how long it usually takes the landing worker to complete each push? I would assume the processing time for each job would be long enough that hitting Phab with one of these requests wouldn't have a huge impact on performance.

zzzeid: > Very little - the call itself is lightweight as it just passes the callsign to Phab, and Phab schedules a repo update sooner than the standard update frequency.

I think as long as Phabricator can handle these calls and they don't cause any problems from the phabricator daemon perspective, this makes sense to me.

> I think that's mostly the point, we want the repo to update ASAP after landings take place. Do we know how long it usually takes the landing worker to complete each push? I would assume the processing time for each job would be long enough that hitting Phab with one of these requests wouldn't have a huge impact on performance.

Given the above, this may be OK. We do store the job duration on each job in the DB so that can be checked, but it's definitely in the multiple seconds at the least.

landoapi/tasks.py
```diff
@@ -160,3 +160,20 @@ def admin_remove_phab_project(
         objectIdentifier=revision_phid,
         transactions=transactions,
     )
+
+
+@celery.task(
+    autoretry_for=(IOError, PhabricatorCommunicationException),
+    default_retry_delay=20,
+    max_retries=3 * 20,  # 20 minutes
```
zzzeid, line 168: `3 * 20` seems like a large number of max retries? (i.e. 60). The comment is not very clear as to what 20 minutes is referring to (maybe it's meant for `default_retry_delay`?)

zzzeid (approved):
Looks good other than minor comment.

PR #217, opened 2022-09-15
hg: add email validation for hg users (bug 1691635)

grahamhub:
while moz-phab will hopefully catch invalid emails before they make it to lando now, this PR adds an extra safeguard against incorrectly configured mercurial users' commits landing in a repo

[Bug 1691635 - Lando landed a commit without a valid email address](https://bugzilla.mozilla.org/show_bug.cgi?id=1691635)

### DIFF
--- a/landoapi/hg.py
+++ b/landoapi/hg.py
@@ -12,11 +12,12 @@ import shutil
 import tempfile
 import uuid
 
-from typing import List, Optional
+from typing import List, Optional, Union
 
 import hglib
 
 from landoapi.hgexports import PatchHelper
+from landoapi.validation import is_valid_email
 
 logger = logging.getLogger(__name__)
 
@@ -353,9 +354,12 @@ class HgRepo:
             # --landing_system is provided by the set_landing_system hgext.
             date = patch_helper.header("Date")
             user = patch_helper.header("User")
-
             if not user:
                 raise ValueError("Missing `User` header!")
+
+            email = self.extract_email_from_username(user)
+            if not is_valid_email(email):
+                raise ValueError("Invalid email configured for Mercurial user!")
 
             if not date:
                 raise ValueError("Missing `Date` header!")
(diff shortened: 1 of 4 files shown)


landoapi/validation.py
```diff
@@ -6,6 +6,7 @@
 from connexion import ProblemException
 
 REVISION_ID_RE = re.compile(r"^D(?P<id>[1-9][0-9]*)$")
+VALID_EMAIL_RE = re.compile(r"[^@ \t\r\n]+@[^@ \t\r\n]+\.[^@ \t\r\n]+")
```
zzzeid, line 9: This may be a misnomer since it could catch invalid email addresses too. Maybe `ACCEPTED_EMAIL_RE` is a better name. There probably is a more suitable regex that is available on the internet.

An example of an invalid email that will pass the above: `-@...`

landoapi/validation.py
```diff
@@ -35,3 +36,8 @@ def parse_landing_path(landing_path: list[dict]) -> list[tuple[int, int]]:
             f"The provided landing_path was malformed.\n{str(e)}",
             type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
         )
+
+
+def is_valid_email(email: str) -> bool:
+    """Given a string, determines if it is a valid email."""
```
zzzeid, line 42: You can probably put the regex pattern here, and describe in more detail in the docstring what it's doing.

zzzeid (changes requested):

landoapi/hg.py
```diff
@@ -519,3 +523,9 @@ def read_checkout_file(self, path: str) -> str:
 
         with checkout_file_path.open() as f:
             return f.read()
+
+    @staticmethod
+    def extract_email_from_username(username: Union[str, bytes]) -> str:
+        """Extracts an email from a Mercurial username, if it exists.
+        Not guaranteed to return a valid email, make sure to validate."""
```
zzzeid, line 530: See pep 257.

```suggestion
        """Extracts an email from a Mercurial username, if it exists.
        
        Not guaranteed to return a valid email, make sure to validate."""
```

zzzeid (changes requested):
Thought I would leave the comments here explicitly (though we chatted about them a while ago over Zoom). Main comment is re: the use of `replace`.

PR #220, opened 2022-09-20
transplants: mv CODE_FREEZE_OFFSET and test codefreeze_datetime

grahamhub:


landoapi/constants.py
```diff
@@ -0,0 +1,2 @@
+# The code freeze dates generally correspond to PST work days.
```
cgsheeh, line 1: This doesn't need its own file, just put it at the top level in the relevant module, `landoapi.models.revisions` seems appropriate.

cgsheeh (approved):

landoapi/models/revisions.py
```diff
@@ -19,6 +19,9 @@
 
 logger = logging.getLogger(__name__)
 
+# The code freeze dates generally correspond to PST work days.
+CODE_FREEZE_OFFSET = "-0800"
```
zzzeid, line 23: Nit: I think this should be moved to `landoapi.transplants`, where it is used.

zzzeid, line 23: Nit: I think this should be moved to landoapi.transplants, where it is used.

zzzeid (changes requested):

PR #233, opened 2022-11-25
autoformat: add bug numbers to autoformat commit messages (Bug 1802558)

cgsheeh:
Instead of having `No bug` in the commit message title,
use the bug numbers from all changesets in the stack.
Move collection of stack changeset titles earlier in
the job processing and parse bug IDs at the landing
job level, passing the ids into the formatting function
as well as into the uplift code where they were previously
parsed. Add a function `bug_list_to_commit_string` to
convert these bug IDs into an appropriate bug string
and add this string to the autoformat commit message
template. Update the test commit messages with a bug
number to ensure the parsing is correct.

### DIFF
--- a/landoapi/commit_message.py
+++ b/landoapi/commit_message.py
@@ -198,3 +198,11 @@ def split_title_and_summary(msg: str) -> Tuple[str, str]:
     tail = parts[1:]
     summary = "\n".join(tail).strip()
     return title, summary
+
+
+def bug_list_to_commit_string(bug_ids: List[str]):
+    """Convert a list of `str` bug IDs to a string for a commit message."""
+    if not bug_ids:
+        return "No bug"
+
+    return f"Bug {', '.join(bug_ids)}"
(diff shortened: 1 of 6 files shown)


landoapi/commit_message.py
```diff
@@ -198,3 +198,11 @@ def split_title_and_summary(msg: str) -> Tuple[str, str]:
...
     summary = "\n".join(tail).strip()
     return title, summary
+
+
+def bug_list_to_commit_string(bug_ids: List[str]) -> str:
+    """Convert a list of `str` bug IDs to a string for a commit message."""
+    if not bug_ids:
+        return "No bug"
+
+    return f"Bug {', '.join(bug_ids)}"
```
zzzeid, line 208: Probably worth supporting singular and plural forms here. E.g.

```
Bug XXXX
Bug XXXX, Bug YYYY, Bug ZZZZ
```

Do we have an existing convention for multiple bugs? Thinking of things that parse commit messages that look for this form.

zzzeid (changes requested):

PR #235, opened 2022-11-29
autoformatting: add a details field to `AutoformattingException` and populate with mach output (Bug 1803169)

cgsheeh:
Add a field to `AutoformattingException` that contains relevant details
for display in failure emails.

Also add type hints to `transition_status` while we are here.

### DIFF
--- a/landoapi/hg.py
+++ b/landoapi/hg.py
@@ -105,7 +105,14 @@ class PatchConflict(PatchApplicationFailure):
 class AutoformattingException(Exception):
     """Exception when autoformatting fails to format a patch stack."""
 
-    pass
+    def __init__(self, *args: object, details: Optional[str] = None) -> None:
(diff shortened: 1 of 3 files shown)


landoapi/hg.py
```diff
@@ -105,7 +105,14 @@ class PatchConflict(PatchApplicationFailure):
...
 
-    pass
+    def __init__(self, *args: object, details: Optional[str] = None) -> None:
+        super().__init__(*args)
+
+        self._details = details
+
+    def details(self) -> str:
+        """Return error details for display."""
+        return self._details if self._details else str(self)
```
zzzeid, line 115: Might make more sense to move this logic to landing_worker.py, i.e.:

```python
f"{exc.details or exc}
```

zzzeid (approved):
lgtm with minor comment.

PR #237, opened 2022-12-07
typing: add/clarify type hints in several places (Bug 1759890)

cgsheeh:
- Add type hints to some functions in `revisions.py`.
- Add type hints to `get_project_phid`.
- Make `flags` kwarg an `Optional`.
- Make `action` a `LandingJobAction` instead of a `str`.
- Remove `diff_id` from base exception handling as it is irrelevant, may be unbound.
- Add `dict` type for list elements in `patch_to_changes`.
- Use a different variable name than `f`, add docstring.
- Add a docstring for `serialize_hunk`.
- Add type hints for `RevisionStack` constructor.
- Fix types for `RevisionStack`.
- Use new variable name for start/end datetimes.

### DIFF
--- a/landoapi/commit_message.py
+++ b/landoapi/commit_message.py
@@ -72,7 +72,7 @@ def format_commit_message(
     approvals: List[str],
     summary: str,
     revision_url: str,
-    flags: List[str] = None,
+    flags: Optional[List[str]] = None,
 ) -> Tuple[str, str]:
     """
     Creates a default format commit message using revision metadata.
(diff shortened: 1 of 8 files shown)


landoapi/commit_message.py
```diff
@@ -72,7 +72,7 @@ def format_commit_message(
     approvals: List[str],
     summary: str,
     revision_url: str,
-    flags: List[str] = None,
+    flags: Optional[List[str]] = None,
```
zzzeid, line 75: I personally prefer the `List[str] | None` notation to using `Optional`, here and elsewhere. Personal preference only.

zzzeid (approved):

PR #239, opened 2022-12-07
uplift: stop using a `landing_path` for uplift endpoint (Bug 1801926)

cgsheeh:
Change the swagger schema for the uplift endpoint to take
a revision ID instead of a landing path. Add simple stack
walking methods to `RevisionStack` and make `get_uplift_conduit_state`
return a `RevisionStack`. Use these stack walking methods in
place of the passed `landing_path` to determine the order
in which to create uplift revisions.

### DIFF
--- a/landoapi/stacks.py
+++ b/landoapi/stacks.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import logging
 from collections import namedtuple
+from collections.abc import Iterator
 from typing import (
     Callable,
     Iterable,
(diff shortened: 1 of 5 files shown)


tests/test_stacks.py
```diff
@@ -670,3 +671,20 @@ def test_integrated_stack_has_revision_security_status(
     revisions = {r["phid"]: r for r in response.json["revisions"]}
     assert not revisions[public_revision["phid"]]["is_secure"]
     assert revisions[secure_revision["phid"]]["is_secure"]
+
+
+def test_revisionstack():
```
zzzeid, line 676: How does `iter_stack_from_base` behave if there is more than one node in `stack.base_revisions()`? E.g., if a stack of 3 revisions has a revision with two predecessors. Would be good to add a test for this, and maybe document it in the docstring as well.

landoapi/stacks.py
```diff
@@ -142,6 +143,22 @@ def __init__(self, nodes: set[str], edges: set[tuple[str, str]]):
             self.children[parent].add(child)
             self.parents[child].add(parent)
 
+    def base_revisions(self) -> Iterator[str]:
+        """Return the set of base revisions in the stack."""
```
zzzeid, line 147: Would be good to add a couple of examples in the docstring giving a couple of stack configurations as examples.

zzzeid (changes requested):

tests/test_stacks.py
cgsheeh: As of right now it simply picks the first available base revision and walks the stack from there. I will document this in the docstring. I suppose I could alternatively disallow any attempt to create uplift revisions where there is a non-linear stack path, just to keep things simple.

I had to change the implementation to work in this way because the `landing_path` variable is pulled from `calculate_landable_subgraphs`, where closed revisions are counted as "not landable". This is a problem because we want to be able to uplift closed revisions.

zzzeid (approved):

PR #240, opened 2022-12-07
uplift: remove `move_drev_to_original` and instead place line directly from `revision.fields.uri` (Bug 1804508)

cgsheeh:
The `summary` field in Phabricator does not include the
`Differential Revision` line, so `move_drev_to_original`
effectively does nothing. Instead place the value of
`revision.fields.uri` in the summary as `Original Revision`
directly.

### DIFF
--- a/landoapi/uplift.py
+++ b/landoapi/uplift.py
@@ -44,7 +44,7 @@ MAX_UPLIFT_STACK_SIZE = 5
 UPLIFT_BUG_UPDATE_RETRIES = 3
 
 
-def add_original_revision_line(summary: str, uri: str) -> str:
+def add_original_revision_line_if_needed(summary: str, uri: str) -> str:
     """Return the summary with `Original Revision` added."""
     if any(line.startswith("Original Revision:") for line in summary.splitlines()):
         return summary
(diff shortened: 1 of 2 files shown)


landoapi/uplift.py
```diff
@@ -42,46 +41,15 @@
...
-        rev = match.group("rev")
-        return f"\nOriginal Revision: {phab_url}/D{rev}"
+UPLIFT_BUG_UPDATE_RETRIES = 3
 
-    # Update the commit message.
-    return ARC_DIFF_REV_RE.sub(repl, body)
 
+def add_original_revision_line(summary: str, uri: str) -> str:
+    """Return the summary with `Original Revision` added."""
+    if any(line.startswith("Original Revision:") for line in summary.splitlines()):
```
zzzeid, line 49: I think this check should be in `create_uplift_revision`. Not sure if this method is necessary altogether, but if you want to keep it it should be renamed to `add_original_revision_line_if_needed`.

zzzeid (changes requested):

landoapi/uplift.py
cgsheeh: This function is here mostly so we can have a unit test of the behaviour. I'll update the function name.

zzzeid (approved):

PR #242, opened 2022-12-08
uplift: add check for bug IDs to get_uplift_conduit_state (Bug 1804506)

cgsheeh:
`create_uplift_revision` requires the revision to have a bug
ID present in the `bugzilla.bug-id` field. Add a check to
`get_uplift_conduit_state` to assert each revision being uplifted
is associated with a bug.

### DIFF
--- a/landoapi/uplift.py
+++ b/landoapi/uplift.py
@@ -96,16 +96,15 @@ def get_release_managers(phab: PhabricatorClient) -> dict:
     return phab.single(groups, "data")
 
 
-def get_revisions_without_bugs(
-    phab: PhabricatorClient, revisions: dict[str, dict]
-) -> set[str]:
+def get_revisions_without_bugs(phab: PhabricatorClient, revisions: dict) -> set[str]:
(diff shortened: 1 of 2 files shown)


landoapi/uplift.py
```diff
@@ -140,6 +154,13 @@ def get_uplift_conduit_state(
             f"Cannot create uplift for stack > {MAX_UPLIFT_STACK_SIZE} revisions."
         )
 
+    missing_bugs = get_revisions_without_bugs(phab, stack_data.revisions)
+    if missing_bugs:
+        missing = ", ".join(missing_bugs)
+        raise ValueError(
+            f"Every uplifted patch must have an associated bug ID: {missing} do not."
```
zzzeid, line 161: I think it may be generally useful to have revision IDs (perhaps with PHIDs if needed) here, as it's not straightforward to look up revisions with a PHID, if one is trying to do some debugging.

zzzeid (approved):
with one nit.

PR #243, opened 2022-12-12
stacks: attach `projects` for repos in `request_extended_revision_data` (Bug 1805265)

cgsheeh:
In https://github.com/mozilla-conduit/lando-api/pull/236 I removed an
unnecessary call to `diffusion.repository.search` as we were already doing
a search for the same repo earlier in the program logic and passing the
returned object into the check. After removing and deploying this call,
Lando started failing when trying to access the projects attachment on
the repo object. This is because the call to `diffusion.repository.search`
doesn't include the projects attachment parameter.

We need to add `projects: True` to the `diffusion.repository.search` call
so the projects data is available.

### DIFF
--- a/landoapi/stacks.py
+++ b/landoapi/stacks.py
@@ -121,6 +121,7 @@ def request_extended_revision_data(
     if repo_phids:
         repos = phab.call_conduit(
             "diffusion.repository.search",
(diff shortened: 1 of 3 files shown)


dklawren (approved):
LGTM. r=dkl

PR #246, opened 2022-12-14
meta: add license headers to files where it is missing (Bug 1805674)

cgsheeh:


zzzeid (changes requested):
nit: I feel like there's a bit of grey area when it comes to `script.py.mako`, since this is part of the initialization of alembic and is automatically generated by it, not by us and not based on any of our input.

Also, it is not a Python file, and if you intended for the comment to apply to the mako file you'll need two `#` per line (see [this](https://docs.makotemplates.org/en/latest/syntax.html#comments)).

If we're going to add the license header to generated files as well (i.e., all the migrations) we could also add the license header to the template, at which point it may make sense to also add the license header to the top of the mako file itself.

cgsheeh:
> 

Ah that makes sense. I'll just remove it instead, I didn't plan to add licenses to generated files.

cgsheeh:
Hm, maybe I should be doing the same for the other migrations file then...

zzzeid (approved):

PR #248, opened 2022-12-16
repos: add release uplift repos (Bug 1806141)

cgsheeh:


cgsheeh:
This is blocked on the repos being added to Phab, but getting a review so I can land when needed would be great. :+1:

zzzeid (approved):
This will likely cause a bit of a delay/log jam after deployment since the landing worker has to clone all three repos from scratch.

PR #252, opened 2022-12-22
typing: fix a few typing issues (Bug 1759890)

cgsheeh:
- Pass `landable_repos` as a `set` instead of a `dict`.
- Add types for `result_list_to_phid_dict`.
- Fix types for some functions in `reviewer_identity`.
- Add a `None` check for `sec_approval_project_phid` and add
  add sec-approval project to tests where it is needed.

### DIFF
--- a/landoapi/api/stacks.py
+++ b/landoapi/api/stacks.py
@@ -91,7 +91,7 @@ def get(phab: PhabricatorClient, revision_id: str):
     )
 
     landable, blocked = calculate_landable_subgraphs(
-        stack_data, edges, set(landable_repos), other_checks=other_checks
+        stack_data, edges, landable_repos, other_checks=other_checks
     )
     uplift_repos = [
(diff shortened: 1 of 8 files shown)


zzzeid (changes requested):
Feels that it would be a little cleaner to either make `get_landable_repos_for_revision_data` return a set, or have `calculate_landable_subgraphs` accept a list.

cgsheeh:
> Feels that it would be a little cleaner to either make `get_landable_repos_for_revision_data` return a set, or have `calculate_landable_subgraphs` accept a list.

We can't do the former because `get_landable_repos_for_revision_data` is used elsewhere and we need a `dict` to be returned for those use cases.

We really only need to do `object in landable_repos` with that value (ie `__contains__`), so we should probably make the type hint `Collection` and then pass the `dict` (see https://docs.python.org/3/library/collections.abc.html?highlight=collections#collections.abc.Container).

PR #256, opened 2023-01-06
transplants: add warning for revisions with unresolved comments (Bug 1504720)

grahamhub:
[Bug 1504720 - Require confirmation when a revision has unresolved comments](https://bugzilla.mozilla.org/show_bug.cgi?id=1504720)
- right now if a revision has unresolved comments in the stack, that information is not surfaced to the user
- adds a warning which will check the transactions for the revisions in `to_land` and determine if any of the inline comments are still unresolved

### DIFF
--- a/landoapi/models/revisions.py
+++ b/landoapi/models/revisions.py
@@ -35,7 +35,7 @@ class DiffWarningGroup(enum.Enum):
 class DiffWarning(Base):
     """Represents a warning message associated with a particular diff and revision."""
 
-    # A Phabricator revision and diff ID (NOTE: revision ID does not inlude a prefix.)
+    # A Phabricator revision and diff ID (NOTE: revision ID does not include a prefix.)
     revision_id = db.Column(db.Integer, nullable=False)
     diff_id = db.Column(db.Integer, nullable=False)
 
--- a/landoapi/transactions.py
+++ b/landoapi/transactions.py
@@ -74,7 +74,9 @@ def get_raw_comments(transaction):
     ]
 
 
-def get_inline_comments(phab: PhabricatorClient, object_identifer: str) -> filter:
+def get_inline_comments(
+    phab: PhabricatorClient, object_identifer: str
+) -> Iterator[Transaction]:
     """Returns an iterable of inline comments for the requested object.
 
     Args:
--- a/landoapi/transactions.py
+++ b/landoapi/transactions.py
@@ -72,3 +72,17 @@ def get_raw_comments(transaction):
         PhabricatorClient.expect(comment, "content", "raw")
         for comment in PhabricatorClient.expect(transaction, "comments")
     ]
+
+
+def get_inline_comments(phab: PhabricatorClient, object_identifer: str) -> filter:
+    """Returns an iterable of inline comments for the requested object.
+
+    Args:
+        phab: A PhabricatorClient instance.
+        object_identifer: An object identifier (PHID or monogram) whose inline
+            comments we want to fetch.
+    """
+    return filter(
+        lambda transaction: transaction["type"] == "inline",
+        transaction_search(phab, object_identifer),
+    )
--- a/landoapi/transactions.py
+++ b/landoapi/transactions.py
@@ -77,7 +77,7 @@ def get_raw_comments(transaction):
 def get_inline_comments(
     phab: PhabricatorClient, object_identifer: str
 ) -> Iterator[Transaction]:
-    """Returns an iterable of inline comments for the requested object.
+    """Returns an iterator of inline comments for the requested object.
 
     Args:
         phab: A PhabricatorClient instance.
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -358,10 +358,10 @@ def warning_code_freeze(*, repo, **kwargs):
 
 @RevisionWarningCheck(9, "Revision has unresolved comments.")
 def warning_unresolved_comments(*, phab, revision, **kwargs):
-    if False in [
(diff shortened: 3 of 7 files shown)


landoapi/transplants.py
```diff
@@ -355,6 +356,15 @@ def warning_code_freeze(*, repo, **kwargs):
         ]
 
 
+@RevisionWarningCheck(9, "Revision has unresolved comments.")
+def warning_unresolved_comments(*, phab, revision, **kwargs):
+    if False in [
+        PhabricatorClient.expect(inline, "fields", "isDone")
+        for inline in get_inline_comments(phab, f"D{revision['id']}")
+    ]:
```
cgsheeh, line 364: This is the more Pythonic way:

```python
if not all(
    phab.expect(inline, "fields", "isDone")
    for inline in get_inline_comments(phab, f"D{revision['id']}")
):
```
It uses the [`all` builtin](https://docs.python.org/3/library/functions.html#all) which is more efficient and takes an iterator, so we don't need to create a full `list` - if one of the first few elements is `False` the function will return `False` immediately.

Nit: please use `phab.expect` here since we already have an instance of the class available.

landoapi/transactions.py
```diff
@@ -72,3 +72,17 @@ def get_raw_comments(transaction):
         PhabricatorClient.expect(comment, "content", "raw")
         for comment in PhabricatorClient.expect(transaction, "comments")
     ]
+
+
+def get_inline_comments(phab: PhabricatorClient, object_identifer: str) -> filter:
```
cgsheeh, line 77: The return type here is actually `Iterator[Transaction]`.

landoapi/models/revisions.py
```diff
@@ -35,7 +35,7 @@ class DiffWarningGroup(enum.Enum):
 class DiffWarning(Base):
     """Represents a warning message associated with a particular diff and revision."""
 
-    # A Phabricator revision and diff ID (NOTE: revision ID does not inlude a prefix.)
+    # A Phabricator revision and diff ID (NOTE: revision ID does not include a prefix.)
```
cgsheeh, line 38: Good catch. :+1:

cgsheeh (changes requested):

landoapi/transactions.py
zzzeid: I don't think `Transaction` is a type or a class here... Probably just a `dict`?

cgsheeh: It's [just an alias to `dict`](https://github.com/mozilla-conduit/lando-api/blob/main/landoapi/transactions.py#L10) but still technically correct. It's at least nice to know the structure of the `dict` is the one for a transaction. :)

In the future it would be nice to make the available values more defined (ie create some proper types for things like `Revision`, `Diff` etc that we pull from Phabricator).

tests/test_transplants.py
```diff
@@ -1085,3 +1085,131 @@ def test_codefreeze_datetime_mock(codefreeze_datetime):
...
+
+
+def test_unresolved_comment_stack(
+    client,
+    db,
+    phabdouble,
+    auth0_mock,
+    release_management_project,
+):
+    repo = phabdouble.repo()
```
zzzeid, line 1159: I find it useful to add a docstring for tests, describing what the test is doing in plain language, particularly for larger tests. E.g., "Ensure a warning is generated when a revision in a stack has unresolved commits.".

zzzeid (changes requested):

tests/test_transplants.py
```diff
@@ -1085,3 +1085,131 @@ def test_codefreeze_datetime_mock(codefreeze_datetime):
...
+    )
+
+    phabdouble.transaction(
+        transaction_type="inline",
+        object=r3,
+        comments=["this is done"],
+        fields={"isDone": True},
+    )
+
+    # Function should filter out unrelated transaction types.
```
cgsheeh, line 1193: What is `Function` here?

tests/test_transplants.py
```diff
@@ -1085,3 +1085,131 @@ def test_codefreeze_datetime_mock(codefreeze_datetime):
...
+    d1 = phabdouble.diff()
+    r1 = phabdouble.revision(diff=d1, repo=phabdouble.repo())
+    phabdouble.reviewer(r1, phabdouble.user(username="reviewer"))
+    phabdouble.transaction(
+        transaction_type="inline",
+        object=r1,
+        comments=["this is done"],
+        fields={"isDone": True},
+    )
+    # Function should filter out unrelated transaction types.
```
cgsheeh, line 1106: What is `Function` here?

cgsheeh (changes requested):
Just clarify those comments in the tests and then this is ready to go. :+1:

tests/test_transplants.py
```diff
@@ -1085,3 +1085,146 @@ def test_codefreeze_datetime_mock(codefreeze_datetime):
...
+
+def test_unresolved_comment_stack(
+    client,
+    db,
+    phabdouble,
+    auth0_mock,
+    release_management_project,
+):
+    """Ensure a warning is generated when a revision in the stack has unresolved
+    comments.
```
zzzeid, line 1169: First line of docstring should fit on one line (see [PEP 257](https://peps.python.org/pep-0257/#multi-line-docstrings)).

cgsheeh (approved):
Well done. :+1:

zzzeid (changes requested):
Looks good to me just one minor thing.

zzzeid (approved):
I think the docstring you changed should probably go on its own line/be reworded slightly since I think it's over the 88 characters. Otherwise lgtm!

PR #257, opened 2023-01-11
transactions: remove unused `get_raw_comments` (Bug 1809730)

cgsheeh:
Remove `get_raw_comments` as it is an unused function.
While we're here also add type hints to the other functions
in this file.

### DIFF
--- a/landoapi/transactions.py
+++ b/landoapi/transactions.py
@@ -1,8 +1,14 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 """Functions for working with Phabricator transactions."""
-from typing import Iterator, NewType
+
+from typing import (
+    Iterator,
+    NewType,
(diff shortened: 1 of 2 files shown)


landoapi/transactions.py
```diff
@@ -14,7 +20,10 @@
 
 
 def transaction_search(
-    phabricator, object_identifier, transaction_phids=None, limit=100
+    phabricator: PhabricatorClient,
+    object_identifier: str,
+    transaction_phids: Optional[list[str]] = None,
```
grahamhub, line 25: What version of Python does lando run on? depending on that might need to change to `List[str]`, but other than that looks goood.

zzzeid (approved):

landoapi/transactions.py
cgsheeh: Lando runs on Python 3.9 at the moment. You can find the Python version in the [Dockerfile for Lando](https://github.com/mozilla-conduit/lando-api/blob/4244a2078bbe98187fcec727f8d43ee939c058aa/Dockerfile#L5) as we build Lando from the `python` image repository.

PR #258, opened 2023-01-12
patches: add type hints everywhere (Bug 1759890)

cgsheeh:


landoapi/patches.py
```diff
@@ -88,14 +93,14 @@ def upload(
...
+    diff_id: int,
+    s3_bucket: str,
     *,
-    aws_access_key,
-    aws_secret_key,
-    endpoint_url=None
-):
+    aws_access_key: str,
+    aws_secret_key: str,
+    endpoint_url: Optional[str] = None,
```
zzzeid, line 102: nit: Could use `|` here to be more consistent (and I personally prefer that but take it or leave it).
```suggestion
    endpoint_url: str | None = None,
```

landoapi/patches.py
```diff
@@ -35,24 +38,26 @@ def create_s3(
...
+    patch: str,
+    s3_bucket: str,
     *,
-    aws_access_key,
-    aws_secret_key,
-    endpoint_url=None
-):
+    aws_access_key: str,
+    aws_secret_key: str,
+    endpoint_url: Optional[str] = None,
```
zzzeid, line 59:
```suggestion
    endpoint_url:str | None = None,
```

zzzeid (approved):

landoapi/patches.py
cgsheeh: I use `Optional` for the most part because of `Option` in Rust and other languages. 

I like this syntax in favor of `Union` (ie below for `bool | str`), but for `something | None` I generally prefer Optional. :)

PR #259, opened 2023-01-13
phabricator: add type hints and remove some unused code (Bug 1759890)

cgsheeh:
Add type hints everywhere in `phabricator.py`. While we're here,
we remove the unused `from_deprecated_id` function in `RevisionStatus`.
Also remove the unused `voidable` function from `ReviewerStatus`, which
allows us to remove `meta` by moving its remaining contents into
`diff_specific`.


grahamhub (approved):
lgtm

PR #261, opened 2023-01-17
api, models: finish adding type hints, make `ConfigurationKey` a proper enum (Bug 1759890)

cgsheeh:
Finish adding type hints to all modules in `api/` and `models/`.

Type hints for `put` are updated from `int` to `str` to reflect
the correct state in `swagger.yml`.

Add a type alias for configuration values that is an alias to
`bool | int | str`. Also raise at the end of the function to
avoid an accidental `None` return in some edge cases.
Convert `ConfigurationKey` to a proper `Enum` type, add
type hints for it and convert usages of it's underlying `str`
to use the `.value` attribute.

Add `flask_sqlalchemy.BaseQuery` as the return type for
model queries.

### DIFF
--- a/landoapi/models/configuration.py
+++ b/landoapi/models/configuration.py
@@ -18,7 +18,8 @@ logger = logging.getLogger(__name__)
 ConfigurationValueType = Union[bool, int, str]
 
 
-class ConfigurationKey:
+@enum.unique
+class ConfigurationKey(enum.Enum):
     """Configuration keys used throughout the system."""
 
     LANDING_WORKER_PAUSED = "LANDING_WORKER_PAUSED"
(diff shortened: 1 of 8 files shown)


landoapi/models/configuration.py
```diff
@@ -4,12 +4,19 @@
...
+    Union,
+)
+
 from landoapi.cache import cache
 from landoapi.models.base import Base
 from landoapi.storage import db
 
 logger = logging.getLogger(__name__)
 
+ConfigurationValue = Union[bool, int, str]
```
zzzeid, line 18: Something about the name that makes me think this is a value and not a type. Perhaps we could just append "Type" to this.

```suggestion
ConfigurationValueType = Union[bool, int, str]
```

landoapi/models/transplant.py
```diff
@@ -87,23 +90,23 @@ def update_from_transplant(self, landed, error="", result=""):
...
+    def head_revision(self) -> str:
         """Human-readable representation of the branch head's Phabricator revision ID."""
         assert (
             self.revision_order
         ), "head_revision should never be called without setting self.revision_order!"
         return "D" + self.revision_order[-1]
 
     @classmethod
-    def revisions_query(cls, revisions):
+    def revisions_query(cls, revisions: list[str]) -> Query:
```
zzzeid, line 105: I think this is technically `flask_sqlalchemy.BaseQuery`.

zzzeid (changes requested):

cgsheeh:
Added some changes around `ConfigurationKey` to this patch FYI.

zzzeid (changes requested):
I think the enum change no longer matches with the title of this commit, it should go in its own commit or the title should be changed.

cgsheeh:
@zzzeid I updated the title, let me know if this is better or if you have a suggestion otherwise.

zzzeid (approved):
I would shorten it to something like:
`api, models: add more type hints, use enum (bug 1759890)`

And put the rest in the message.

PR #265, opened 2023-01-23
uplift: check `target_repo` for `None` (Bug 1759890)

cgsheeh:
Technically this can be `None`, so we should check
it to avoid problems down the line.


zzzeid (approved):

zzzeid:
On a different note, if this target repo is coming in from the API call, we should probably return a 400 error if the target repo is `None`.

PR #224, opened 2022-11-15
workers: revision worker implementation

zzzeid:

PR #266, opened 2023-01-23
stacks: implement `RevisionStack` with `networkx` (Bug 1811469)

cgsheeh:
Re-implement `RevisionStack` with `networkx`. Implement
the class as a subclass of `nx.DiGraph` to use its representation.
Update `child` and `parent` mappings to instead use `predecessor`
and `successor`, bringing it more in line with the usual
terminology used within Conduit.

`base_revisions` is updated to instead iterate over the
elements in the graph with an in-degree of zero. The
in-degree is the number of directed edges that point
to the node (ie the number of predecessors). We also
rename the function to `root_revisions` to stay aligned
with the `networkx` graph terminology.

Add a `dest` argument to `iter_stack_from_base` to set
the "destination" node and change the function to use
the "simple paths" algorithm from `networkx`. Rename
the function to `iter_stack_from_root` to align with
`networkx` terminology.


landoapi/stacks.py
```diff
@@ -135,14 +137,13 @@ def request_extended_revision_data(
...
-        self.parents = {phid: set() for phid in self.nodes}
-        for child, parent in self.edges:
-            self.children[parent].add(child)
-            self.parents[child].add(parent)
+        self.graph = nx.DiGraph(
+            # Reverse the order of the nodes in the edges set as `networkx`
+            # represents `a -> b` as `(a, b)` but Lando uses `(b, a)`.
+            (successor, predecessor)
+            for predecessor, successor in edges
+        )
```
zzzeid, line 145: Haven't looked at the entire diff yet, but there's a `reverse` method that you may be able to use. See my changes in [revision_worker.py](https://github.com/mozilla-conduit/lando-api/pull/224/files#diff-b666a8fe8b2e34d21a8bf41d80705719f945165e15483472bc0d625707e3c8e1R58) for an example.

cgsheeh: We don't need to reverse the order of the edges list, but the order of the node relationships (ie the elements of the list) from `(dest, source)` to `(source, dest)`.

Touching some of the edge/graph related code in Lando/moz-phab the last little while has made me think we should use a more concrete type like a `namedtuple` to represent the edges, that way the order doesn't actually matter and we can just do `edge.source` and `edge.dest`. That's out of scope for this change of course. :)

cgsheeh:
> FYI I have a related change in the revision worker changes where I add networkx as a dependency, and use it for some stack graph functionality (see [here](https://github.com/mozilla-conduit/lando-api/pull/224/files#diff-b666a8fe8b2e34d21a8bf41d80705719f945165e15483472bc0d625707e3c8e1R29)). Thought I'd mention this here, though these changes can probably be landed independently and I can update my patch to reflect that.

Looking at your changes is what made me think to use `networkx`, in fact.

zzzeid (changes requested):
Looks good, with a couple of small comments/suggestions.

landoapi/stacks.py
```diff
@@ -156,32 +157,40 @@ def base_revisions(self) -> Iterator[str]:
...
 
-    def iter_stack_from_base(self) -> Iterator[str]:
+    def iter_stack_from_base(self, head: str) -> Iterator[str]:
         """Iterate over the revisions in the stack starting from the base.
 
-        NOTE: assumes the stack is linear, with one base revision and each
-        subsequent revision having a single child. If there are multiple paths
-        that could be walked in the stack, this function will naively pick one.
+        Walks from one of the root nodes of the graphs to `head`. If multiple
+        root nodes exist, it will select one naively.
```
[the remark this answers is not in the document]
cgsheeh: I switched us to using `root` instead of `base`, and switched to `dest` instead of `head` to align with networkx.

zzzeid (approved):
lgtm!

PR #268, opened 2023-01-23
landing_worker: use `ConfigurationKey` instead of string variable (Bug 1759890)

cgsheeh:
Noticed this lingering usage of a `str` during testing.


grahamhub (approved):

zzzeid (approved):
good catch!

PR #269, opened 2023-02-02
transplants: separate `None` check for required variables (Bug 1759890)

cgsheeh:
Move `None` check for required variables into their own
`if` statements. This makes the exception more explicit
about what values are missing as the previous implementation
would print a `None` to the screen instead of showing
which variable could not be resolved. This also has the
effect of letting type checkers know the value is not
`Optional` after that check.

### DIFF
--- a/landoapi/api/transplants.py
+++ b/landoapi/api/transplants.py
@@ -267,10 +267,19 @@ def post(phab: PhabricatorClient, data: dict):
 
     assessment.raise_if_blocked_or_unacknowledged(confirmation_token)
 
-    if not all((to_land, landing_repo, stack_data)):
+    if not to_land:
         raise ValueError(
-            "One or more values missing in access transplant request: "
-            f"{to_land}, {landing_repo}, {stack_data}"
+            "Could not retrieve revision information from transplant request."
+        )
+
+    if not landing_repo:
+        raise ValueError(
+            "Could not retrieve landing repo information from transplant request."
+        )
+
+    if not stack_data:
+        raise ValueError(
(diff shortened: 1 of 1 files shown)


zzzeid (changes requested):
I think a better approach would be to store the error strings, and then raise an exception after checking all the variables. The original behaviour would inform you of one or more issues, whereas the new behaviour will only inform you of one issue (and it it happens to be the first one, then a new request needs to be submitted to realize there is an additional error.)

I don't know theoretically how often there is more than one issue, but it makes more sense to detect as many issues per request as possible, and then return an error or raise an exception, since this is an API request.

cgsheeh:
> I think a better approach would be to store the error strings, and then raise an exception after checking all the variables. The original behaviour would inform you of one or more issues, whereas the new behaviour will only inform you of one issue (and it it happens to be the first one, then a new request needs to be submitted to realize there is an additional error.)
> 
> I don't know theoretically how often there is more than one issue, but it makes more sense to detect as many issues per request as possible, and then return an error or raise an exception, since this is an API request.

Looking at this again, we have two separate use cases for the `_assess_transplant_request` function. In one case we only care about the `assessment`, so we return `None` for `to_land` etc. In the other case those variables are required, but we are checking they are non-`None` in the caller. I think we should split those two use cases out and make the new function for the second use case return a type without `Optional`, raising if they can't resolve the required values.

cgsheeh:
This will become obsolete after https://github.com/mozilla-conduit/lando-api/pull/388.

PR #271, opened 2023-02-07
types: add type hints to `secapproval.py`, `sentry.py`, `transplant_client.py` (Bug 1759890)

cgsheeh:
Also rename `phabclient` to `phab` to match the usual Lando convention.

### DIFF
--- a/landoapi/secapproval.py
+++ b/landoapi/secapproval.py
@@ -47,7 +47,9 @@ class CommitDescription(NamedTuple):
     sanitized: bool
 
 
-def send_sanitized_commit_message_for_review(revision_phid, message, phabclient):
+def send_sanitized_commit_message_for_review(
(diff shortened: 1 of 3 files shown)


landoapi/secapproval.py
```diff
@@ -86,7 +88,7 @@ def send_sanitized_commit_message_for_review(revision_phid, message, phabclient)
             {"type": "reviewers.add", "value": [f"blocking({sec_approval_phid})"]},
         ],
     )
-    return PhabricatorClient.expect(response, "transactions")
```
zzzeid, line 89: nit: This change is not just a rename, should probably just include that detail in the commit message.

zzzeid (approved):

landoapi/secapproval.py
cgsheeh: It is included in the commit message.

PR #274, opened 2023-02-22
bmo: switch from standard REST API to uplift endpoint (Bug 1818281)

cgsheeh:


cgsheeh:
I'm going to test this on dev before landing, just waiting on a BMO dev deploy.

landoapi/bmo.py
```diff
@@ -8,8 +8,8 @@
 
 
 def bmo_bug_endpoint() -> str:
```
zzzeid, line 10: Should rename this to `bmo_uplift_endpoint` or something else less general, since there is nothing in the module name or method name hinting that this is an uplift related method.

zzzeid (changes requested):

cgsheeh:
The current state has been tested on dev, this should be ready to go.

zzzeid (approved):

PR #276, opened 2023-02-27
workers: add Worker class, extend to LandingWorker (bug 1744327)

zzzeid:
- abstract worker into a reusable class
- implement better functionality to stop workers gracefully


landoapi/workers/landing_worker.py
```diff
@@ -75,170 +69,65 @@ def job_processing(worker: LandingWorker, job: LandingJob, db: SQLAlchemy):
...
+        # Check if any closed trees reopened since the beginning of this iteration
+        if len(self.enabled_repos) != len(self.applicable_repos):
+            self.refresh_enabled_repos()
 
-        self.running = True
-
-        # Initialize state
-        self.refresh_enabled_repos()
-        last_job_finished = True
+        if self.last_job_finished is False:
```
cgsheeh, line 95: `if not self.last_job_finished:`

landoapi/workers/base.py
```diff
@@ -0,0 +1,136 @@
...
+            loops += 1
+
+        logger.info(f"{self} exited after {loops} loops.")
+
+    @property
+    def throttle_seconds(self):
+        return ConfigurationVariable.get(self.THROTTLE_KEY, 3)
+
+    def throttle(self, seconds=None):
+        sleep(seconds if seconds is not None else self.throttle_seconds)
```
cgsheeh, line 118: `self.throttle_seconds` can technically be a `bool` - we should ensure it is a `float` before using it.

cgsheeh (changes requested):
This patch removes the `job_processing` attribute from `LandingWorker` but doesn't account for it being removed in `landoapi.workers.landing_worker.job_processing`.

landoapi/workers/base.py
zzzeid: How can it be a `bool`? The default is `3` (which is an int). It is not set in the DB (yet).

Once we deploy this, we can initialize the database value, and `variable_type` would have to be set to an `INT` on a one-time basis, but it makes no sense for it to be set to a `bool`. The type checking is the responsibility of whomever sets the initial values for these variables in the database (or if there is a command, like in `cli` which sets these values.) They are not designed to be changed arbitrarily.

cgsheeh: The returned type from `ConfigurationVariable.get` is `ConfigurationValueType = Union[bool, int, str]`, so the type checker detects this can possibly be a `bool`.

Perhaps instead of a `get` method for retrieving the variables we should have a `get_bool`, `get_int` etc.

landoapi/workers/base.py
```diff
@@ -0,0 +1,136 @@
...
+            input=ssh_private_key,
+            capture_output=True,
+            universal_newlines=True,
+        )
+        if add_process.returncode != 0:
+            raise Exception(add_process.stderr)
+        logger.info("Added private SSH key from environment.")
+
+    @property
+    def _paused(self):
```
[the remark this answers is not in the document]
zzzeid: I am technically following [PEP8](https://peps.python.org/pep-0008/#documentation-strings) here:
>  Docstrings are not necessary for non-public methods, but you should have a comment that describes what the method does.

I suppose I can add a comment to those that are missing comments.

landoapi/workers/landing_worker.py
```diff
@@ -75,170 +69,65 @@ def job_processing(worker: LandingWorker, job: LandingJob, db: SQLAlchemy):
...
+            logger.info("Starting landing job", extra={"id": job.id})
+            self.last_job_finished = self.run_job(
+                job,
+                repo,
+                hgrepo,
+                treestatus_subsystem.client,
+                current_app.config["PATCH_BUCKET_NAME"],
+            )
             logger.info("Finished processing landing job", extra={"id": job.id})
         logger.info("Landing worker exited")
```
zzzeid, line 129: This line should be removed, it is handled in `self._start`.

landoapi/workers/base.py
zzzeid: > The returned type from `ConfigurationVariable.get` is `ConfigurationValueType = Union[bool, int, str]`, so the type checker detects this can possibly be a `bool`.
> 
> Perhaps instead of a `get` method for retrieving the variables we should have a `get_bool`, `get_int` etc.

We don't have any enforcement of type checking (not sure if you have a way to set your IDE to ignore this particular case if it is bugging you). This would be out of scope for this PR, however I agree that there could be some improvements made to how the DB configuration variable system works. Perhaps you can file a bug for this, as I think it requires further discussion.

landoapi/workers/base.py
```diff
@@ -0,0 +1,136 @@
...
+    def refresh_enabled_repos(self):
+        self.enabled_repos = [
+            r
+            for r in self.applicable_repos
+            if treestatus_subsystem.client.is_open(repo_clone_subsystem.repos[r].tree)
+        ]
+        logger.info(f"{len(self.enabled_repos)} enabled repos: {self.enabled_repos}")
+
+    def start(self, max_loops=None):
+        if ConfigurationVariable.get(self.STOP_KEY, False):
```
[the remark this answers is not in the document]
zzzeid: There is no `START_KEY`, `STOP_KEY` is used to determine whether a worker can be started or not (as seen in this method).

I don't see a harm in moving this to a property as you described, since this is an abstract class and it would make sense to have unimplemented properties that should exist in all extended class here. This should probably be upgraded to a proper abstract base class at some point.

landoapi/workers/base.py
cgsheeh: We can discuss adding helper methods for getting variables of the correct type from the db.

For now we should just wrap the value in a `float` before we use it.

zzzeid: > We can discuss adding helper methods for getting variables of the correct type from the db.
> 
> For now we should just wrap the value in a `float` before we use it.

But it's an `int`, not a `float`. And what difference would that make other than suppress the warning in your IDE? It will throw an error regardless if it gets the wrong input.

I would like to avoid adding code that will just help with IDE-related issues, unless it also has a more general benefit.

Adding a try/except on the other hand may accomplish something if you're trying to handle an incorrect type here.

zzzeid: See: https://github.com/mozilla-conduit/lando-api/blob/main/landoapi/models/configuration.py#L64

landoapi/workers/base.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @property
+    def _paused(self):
+        # When the pause variable is True, the worker is temporarily paused. The worker
+        # resumes when the key is reset to False.
+        return ConfigurationVariable.get(self.PAUSE_KEY, False)
+
+    @property
+    def _running(self):
+        # When the stop variable is True, the worker will exist and will not restart,
```
zzzeid, line 106:
```suggestion
        # When the stop variable is True, the worker will exit and will not restart,
```

landoapi/workers/base.py
```diff
@@ -0,0 +1,160 @@
...
+        # until the value is changed to False.
+        return not ConfigurationVariable.get(self.STOP_KEY, False)
+
+    def _setup(self):
+        """Perform various setup actions."""
+        if hasattr(self, "ssh_private_key"):
+            # Run various ssh configuration commands.
```
zzzeid, line 113:
```suggestion
```

landoapi/workers/base.py
zzzeid: Probably worth noting that `time.sleep` does not care about the variable type it gets (and in fact would work if provided a bool, though it would interpret those as integers 0 and 1). I think [this](https://github.com/python/cpython/blob/main/Python/pytime.c#L553) is where it interprets the argument.

landoapi/workers/base.py
```diff
@@ -0,0 +1,165 @@
...
+        # When the stop variable is True, the worker will exit and will not restart,
+        # until the value is changed to False.
+        return not ConfigurationVariable.get(self.STOP_KEY, False)
+
+    def _setup(self):
+        """Perform various setup actions."""
+        if hasattr(self, "ssh_private_key"):
+            self._setup_ssh(self.ssh_private_key)
+
+    def _start(self, max_loops=None, *args, **kwargs):
```
cgsheeh, line 121: Type hint on `max_loops`.

landoapi/workers/base.py
```diff
@@ -0,0 +1,165 @@
...
+                self.throttle(self.sleep_seconds)
+            self.loop(*args, **kwargs)
+            loops += 1
+
+        logger.info(f"{self} exited after {loops} loops.")
+
+    @property
+    def throttle_seconds(self) -> int:
+        """The duration to pause for when the worker is being throttled."""
+        return ConfigurationVariable.get(self.THROTTLE_KEY, 3)
```
cgsheeh, line 140:
```suggestion
        return int(ConfigurationVariable.get(self.THROTTLE_KEY, 3))
```

landoapi/workers/landing_worker.py
```diff
@@ -65,179 +60,81 @@ def job_processing(worker: LandingWorker, job: LandingJob, db: SQLAlchemy):
...
         job.duration_seconds = (datetime.now() - start_time).seconds
         db.session.commit()
 
 
-class LandingWorker:
-    def __init__(self, sleep_seconds: float = 5.0):
-        SSH_PRIVATE_KEY_ENV_KEY = "SSH_PRIVATE_KEY"
+class LandingWorker(Worker):
+    @property
+    def STOP_KEY(self) -> str:
```
cgsheeh, line 73:
```suggestion
    def STOP_KEY(self) -> ConfigurationKey:
```

landoapi/workers/base.py
```diff
@@ -0,0 +1,165 @@
...
+    def refresh_enabled_repos(self):
+        """Refresh the list of repositories based on treestatus."""
+        self.enabled_repos = [
+            r
+            for r in self.applicable_repos
+            if treestatus_subsystem.client.is_open(repo_clone_subsystem.repos[r].tree)
+        ]
+        logger.info(f"{len(self.enabled_repos)} enabled repos: {self.enabled_repos}")
+
+    def start(self, max_loops=None):
```
cgsheeh, line 155: Type hint for `max_loops`.

landoapi/workers/base.py
cgsheeh: I'll just fix it myself after this lands.

landoapi/workers/base.py
```diff
@@ -0,0 +1,165 @@
...
+from time import sleep
+from landoapi.repos import repo_clone_subsystem
+from landoapi.treestatus import treestatus_subsystem
+from landoapi.models.configuration import ConfigurationVariable, ConfigurationKey
+
+
+logger = logging.getLogger(__name__)
+
+
+class Worker:
```
cgsheeh, line 20: Docstring for this class while we're here.

cgsheeh (changes requested):

landoapi/workers/base.py
cgsheeh: As mentioned elsewhere I'll fix the type issue here after this patch lands.

zzzeid: The suggestion you made here is technically redundant, as I [linked to](https://github.com/mozilla-conduit/lando-api/blob/main/landoapi/models/configuration.py#L64) elsewhere, this exact conversion is already made when calling `get`, and this change would be only to satisfy the type-checker/IDE (which is, in my opinion not a good enough reason to make this change.) Thus we should refactor this to make it possible to satisfy the type checker without having redundant code, but that refactoring would require some discussion and is out of scope for this PR.

cgsheeh (approved):

PR #278, opened 2023-02-28
storage: move _lock_table_for to Base model (bug 1819332)

zzzeid:


landoapi/storage.py
```diff
@@ -1,16 +1,35 @@
...
 from landoapi.systems import Subsystem
 
+if TYPE_CHECKING:
+    from landoapi.models.base import Base
+
 db = SQLAlchemy()
 migrate = Migrate()
 
 
+def _lock_table_for(
```
cgsheeh, line 19: Since this is something we intend to be used outside this module we should rename it to remove the `_` prefix, ie `lock_table_for`.

Out of scope for this change: maybe this function makes more sense as a method on `Base`? Then we could just do `Base.lock_table()` for example.

cgsheeh (changes requested):

landoapi/api/transplants.py
```diff
@@ -389,7 +374,8 @@ def post(phab: PhabricatorClient, data: dict):
 
     if not landing_repo.legacy_transplant:
         with db.session.begin_nested():
-            _lock_table_for(model=LandingJob)
+            LandingJob.lock_table()
+            LandingJob
```
zzzeid, line 378:
```suggestion
```

landoapi/models/base.py
```diff
@@ -43,3 +43,13 @@ def __repr__(self) -> str:
         For example, `<Transplant: 1235>`.
         """
         return f"<{self.__class__.__name__}: {self.id}>"
+
+    @classmethod
+    def lock_table(cls, mode: str = "SHARE ROW EXCLUSIVE MODE"):
+        """Locks a given table in the given database with the given mode.
```
zzzeid, line 49:
```suggestion
        """Locks the table for the model with the given mode.
```

cgsheeh (approved):

PR #279, opened 2023-02-28
stacks: refactor build_stack_graph endpoint and tests (bug 1819408)

zzzeid:


landoapi/stacks.py
```diff
@@ -25,18 +25,15 @@
...
 
-def build_stack_graph(
-    phab: PhabricatorClient, revision_phid: str
-) -> tuple[set[str], set[tuple[str, str]]]:
+def build_stack_graph(revision: dict) -> tuple[set[str], set[tuple[str, str]]]:
     """Return a graph representation of a revision stack.
 
     This function is expensive and can make up to approximately
     n/2 calls to phabricator for a linear stack where n is the
     number of revisions in the stack.
```
zzzeid, line 33: No longer relevant.
```suggestion
```

tests/conftest.py
```diff
@@ -464,3 +464,23 @@ def strptime(cls, date_string, fmt):
             return dates[f"{date_string}"]
 
     return Mockdatetime
+
+
+@pytest.fixture
+def revision_from_api(phabdouble):
+    """Gets revision from the Phabricator API, given a revision.
```
zzzeid, line 471:
```suggestion
    """Get revision from the Phabricator API, given a revision.
```

tests/mocks.py
```diff
@@ -96,6 +97,51 @@ def validate_change(change):
...
+    # Treat the stack like a commit DAG, we only care about edges going
+    # from child to parent. This is enough to represent the graph.
+    edges = {
+        (edge["sourcePHID"], edge["destinationPHID"])
+        for edge in edges
+        if edge["edgeType"] == "revision.parent"
+    }
+
+    stack_graph = defaultdict(list)
+    sources = [edge[0] for edge in edges]
```
cgsheeh, line 132:
```suggestion
    sources = [source for source, dest in edges]
```

cgsheeh (changes requested):
Looks mostly good, I just ask that we use the `api_object_for` test APIs and a few nits.

landoapi/stacks.py
```diff
@@ -45,36 +42,13 @@ def build_stack_graph(
...
-        for edge in edges
-        if edge["edgeType"] == "revision.parent"
-    }
+    stack_graph = PhabricatorClient.expect(revision, "fields", "stackGraph")
+    phids = set(stack_graph.keys())
+    edges = set()
 
+    for source_node, destination_nodes in stack_graph.items():
+        for predecessor_node in destination_nodes:
+            edges.add((source_node, predecessor_node))
```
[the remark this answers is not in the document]
zzzeid: Good point -- I think the main thing that could be simplified here is `destination_nodes` and we can drop the "source" reference. I think you are referring to the child/parent, ancestor/descendant terminology which we got rid of, and which implied a hierarchy that isn't implied by successor/predecessor.

Source/destination here were intended as more neutral terminology within the scope of the for loop. Maybe this alternative approach is easier to read (and illustrates what I'm trying to get at here).

```python
for node, predecessors in stack_graph.items():
    for predecessor in predecessors:
        edges.add((node, predecessor))
```

It wouldn't make sense to use both "successor" and "predecessor" reference to describe an edge, since those words describe a relative relationship -- and we lose the "relative to what" in that case.

tests/mocks.py
zzzeid: FWIW, this is the same source/dest concept as in the [other comment](https://github.com/mozilla-conduit/lando-api/pull/279/files#r1132474294).

landoapi/stacks.py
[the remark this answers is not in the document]
cgsheeh: That looks good to me. :+1:

landoapi/stacks.py
```diff
@@ -25,18 +25,15 @@
...
-    phab: PhabricatorClient, revision_phid: str
-) -> tuple[set[str], set[tuple[str, str]]]:
+def build_stack_graph(revision: dict) -> tuple[set[str], set[tuple[str, str]]]:
     """Return a graph representation of a revision stack.
 
     This function is expensive and can make up to approximately
     n/2 calls to phabricator for a linear stack where n is the
     number of revisions in the stack.
 
     Args:
```
zzzeid, line 35:
```suggestion
    """Return a graph representation of a revision stack.

    Args:
```

cgsheeh (approved):

PR #280, opened 2023-03-01
transplants: remove legacy code and migrate tests (bug 1654646)

zzzeid:
- remove landoapi.api.landings module
- remove landoapi.transplant_client module
- remove all usages of TransplantClient
- update all tests to use LandingWorker or LandingJob as needed
- remove code logic that uses Repo.legacy_transplant flag
- remove PINGBACK_URL and PINGBACK_ENABLED implementations
- remove endpoints  and tests that update Transplant objects
- fix bug where completed landing jobs were not triggering a warning


tests/test_notifications.py
```diff
@@ -88,24 +85,6 @@ def test_email_content():
...
-    # happy-path runs cleanly.
-    transplant = Transplant(revision_order=["1"])
-    notify_user_of_landing_failure(
-        transplant.requester_email,
-        transplant.head_revision,
-        transplant.error,
-        transplant.request_id,
-    )
-
-
```
zzzeid, line 108: This should be rewritten to use `LandingJob`.

landoapi/transplants.py
```diff
@@ -211,9 +211,9 @@ def warning_previously_landed(*, revision, diff, **kwargs):
...
 
     landed_transplant = (
-        Transplant.revisions_query([revision_id])
-        .filter_by(status=TransplantStatus.landed)
-        .order_by(Transplant.updated_at.desc())
+        LandingJob.revisions_query([revision_id])
+        .filter_by(status=LandingJobStatus.LANDED)
+        .order_by(LandingJob.updated_at.desc())
         .first()
     )
```
cgsheeh, line 218: Did we have entries for each landing in both tables (`LandingJob` and `Transplant`) before? I understand how this is a replacement from the logic side of things, but I don't get how this would work if we didn't have both tables up to date with the latest landings.

cgsheeh (approved):
Code LGTM, just have a question re: one of the `Transplant` -> `LandingJob` changes.

landoapi/transplants.py
zzzeid: > Did we have entries for each landing in both tables (`LandingJob` and `Transplant`) before?

Once a repo was switched over, we stopped creating `Transplant` entries. This backwards compatibility was really in place for a very small number of stacks that had parts of them landed with the old system, and other parts landed with the new system.

This particular line change actually fixes a bug, as this warning was not doing anything for non-legacy landings. I will add a note about that in the description.

zzzeid: This is how the warning should look like, and how it will look like with this fix.

image

PR #283, opened 2023-03-09
treestatus: implement Treestatus API as a function of Lando (Bug 1817472)

cgsheeh:
This commit implements the Treestatus API as a feature of Lando.
The Treestatus OpenAPI spec is added to Lando's API spec and most
endpoints are re-implemented here. The following Treestatus endpoints
are removed in the Lando implementation:

- `/v0/*` endpoints.
- `DELETE /trees2`

The following features are also removed in the Lando implementation:

- StatusPage updating.
- Pulse message publishing.

Treestatus' database, caching and logging is changed to integrate with
Lando's Flask platform.

A new `group` argument is added to the Auth0 decorators to specify the
user must be a member of the specified groups to be permitted to
access certain API endpoints, and two new `TREESTATUS_ADMIN` and
`TREESTATUS_USER` group name constants are added. A `user_id()`
function is added to `A0User` to retrieve the current user ID string
for usage in the `who` field of status updates.

Integration tests for the endpoints are added to show typical usage
patterns and ensure the endpoints return output in a format consistent
with the existing Treestatus API.

A few database adjustments are made in this migration. First, the `tree`
field of `Tree` is now a unique index instead of a primary key, as Lando
models all have an `id` field as the primary key. This also involves
updating some `.get()` calls in SQLALchemy to instead by `filter_by`.
Second, the custom `UTCDatetime` object used by the existing Treestatus
implementation is replaced with a normal `DateTime` SQLAlchemy object.

Since Treestatus is now a part of Lando after this PR, the existing
Treestatus client is removed, and replaced with an `is_open` function
that directly queries the DB to retrieve the status of the tree. The
TreestatusDouble test mocker is removed and replaced with a simple
fixture to create a tree and set it to a specific status in the
tests. The `TreestatusSubsystem` is also removed as it is no longer
necessary.

After the above changes, many more code cleanups and refactorings
were completed to make the code easier to reason about and more
maintainable.


cgsheeh:
This is ready to review now. CircleCI is having trouble but the [build is passing](https://app.circleci.com/pipelines/github/cgsheeh/lando-api/4/workflows/25019c50-9e2e-41d8-8936-d3ff9dfb37f5) on my Lando-API fork.

The failing tests were due to several pieces of the DB query code having `order_by` parameters based on a time field, but in the tests we were mocking the time and returning a constant datetime, so the sorting wasn't deterministic. This plus a slight DB difference between the local test DB images and the image we use in CI caused the failures. I fixed this by adding a mock that increases in time with each call, so actions taken in the test appear to happen with time moving forward.

 I also added a migration file for the new models.

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,619 @@
...
+        session.commit()
+    except (sa.exc.IntegrityError, sa.exc.ProgrammingError):
+        raise ProblemException(
+            400,
+            f"Tree {tree} already exists.",
+            "Tree already exists.",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
+
+    return new_tree.to_dict(), 200
```
zzzeid, line 532:
```suggestion
    return new_tree.to_dict(), 201
```

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,619 @@
...
+    session = db.session
+    tree = session.query(Tree).filter_by(tree=tree_name).one_or_none()
+    if not tree:
+        raise ProblemException(
+            404,
+            f"No tree {tree_name} found.",
+            "The tree does not exist.",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404",
+        )
+    session.delete(tree)
```
zzzeid, line 161: May be better to implement some type of archiving functionality, to avoid accidental deletions of trees and records (e.g., have a `Tree.is_archived` flag.)

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,619 @@
...
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
+
+    return new_tree.to_dict(), 200
+
+
+@auth.require_auth0(
+    groups=(auth.TREESTATUS_ADMIN), scopes=("lando", "profile", "email"), userinfo=True
+)
+def kill_tree(tree: str) -> tuple[str, int]:
```
zzzeid, line 538:
```suggestion
def delete_tree(tree: str) -> tuple[str, int]:
```

landoapi/api/treestatus.py
zzzeid, line 1: I think there is some general cleanup that could be done in this module, I will leave more of those comments in the second pass.

landoapi/auth.py
```diff
@@ -357,11 +371,19 @@ class require_auth0:
...
+        self,
+        groups: Optional[Iterable[str]] = None,
+        scopes: Optional[Iterable[str]] = None,
+        userinfo: bool = False,
+    ):
         assert scopes is not None, (
             "`scopes` must be provided. If this endpoint truly does not "
             "require any scopes, explicilty pass an empty tuple `()`"
         )
+        assert not groups or userinfo, "Requiring `groups` implies `userinfo`."
```
zzzeid, line 384: Shouldn't use asserts in code flow. Instead, these should be changed to if statements / exceptions raised.

landoapi/auth.py
```diff
@@ -407,6 +429,26 @@ def wrapped(*args, **kwargs):
...
 
+    def _require_groups(self, f: Callable) -> Callable:
+        """Decorator which verifies membership in required access groups."""
+
+        @functools.wraps(f)
+        def wrapped(*args, **kwargs):
+            if self.groups and not g.auth0_user.is_in_groups(self.groups):
+                missing_groups = {
+                    group for group in self.groups if group not in g.auth0_user.groups
+                }
```
zzzeid, line 440: Could calculate the missing groups once, and if they are not empty raise.

landoapi/landing_worker.py
```diff
@@ -168,7 +166,7 @@ def refresh_enabled_repos(self):
         self.enabled_repos = [
             r
             for r in self.applicable_repos
-            if treestatus_subsystem.client.is_open(repo_clone_subsystem.repos[r].tree)
+            if treestatus.is_open(repo_clone_subsystem.repos[r].tree)
```
zzzeid, line 169: Would be nice to have something like:
```python
tree = Tree.query.get(name=repo_clone_subsystem.repos[r].tree)
if tree.is_open():
    # ... do stuff
```

zzzeid (changes requested):
First pass! Will do additional passes later in the week.

landoapi/auth.py
cgsheeh: I copied this pattern from the `assert` statement that already exists for `scopes`. I'm happy to change it as I usually don't use `assert` in code flows either. However I'll note that in this case, the code flow that is caught is a programming error instead of something that would be hit due to runtime conditions. So this code path would always throw in the tests when someone on our team uses the decorator incorrectly. That might explain why an `assert` is used here instead of an `if` statement.

landoapi/landing_worker.py
cgsheeh: A class method would be nice, so the query details are close to the model definition: `Tree.is_open(repo_clone_subsystem.repos[r].tree)`

landoapi/spec/swagger.yml
```diff
@@ -423,7 +423,499 @@ paths:
...
+          required: true
+          default: 0
+          type: integer
+      responses:
+        200:
+          description: Status
+          schema:
+            type: string
+
+  /stack2/discard/{id}:
```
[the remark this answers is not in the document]
cgsheeh: I think I'm going to remove the `stack2/{discard,restore}` endpoints and just use the `PUT /stack` endpoint that takes a `revert` parameter. I'm not sure why both exist.

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+            "status": old_tree["status"],
+            "reason": old_tree["reason"],
+            "tags": old_tree["tags"],
+            "log_id": old_tree["log_id"],
+            "current_status": new_tree.status,
+            "current_reason": new_tree.reason,
+            "current_tags": new_tree.tags,
+            "current_log_id": new_tree.log_id,
+        }
+    )
```
zzzeid, line 114: Improves readability since the keys are constant throughout.

```suggestion
    params = ("status", "reason", "tags", "log_id")
    result = {}
    for key in params:
        result[key] = old_tree[key]
        result[f"current_{key}"] = getattr(new_tree, key)
    return json.dumps(result)
```

nit: the JSON serialization could be done elsewhere, either where needed or in a separate function.

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+    reason: Optional[str] = None,
+    tags: Optional[list[str]] = None,
+    message_of_the_day: Optional[str] = None,
+):
+    """Update the given tree's status.
+
+    Note that this does not commit the session.
+    """
+    if not tags:
+        tags = []
```
zzzeid, line 234:
```suggestion
    tags = tags or []
```

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+
+    # Create a log entry if the reason or status has changed.
+    if status or reason:
+        if status is None:
+            status = "no change"
+        if reason is None:
+            reason = "no change"
+        log = Log(
+            tree=tree.tree,
+            created_at=_now(),
```
zzzeid, line 253: This should be automatically filled by the `Base` model.

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+def combinedtree_as_dict(tree: CombinedTree) -> dict[str, Any]:
+    """Convert a `CombinedTree` to a dict.
+
+    Removes the `model` field as part of the conversion.
+    """
+    return {field: value for field, value in tree._asdict().items() if field != "model"}
+
+
+def result_object_wrap(f: Callable) -> Callable:
+    """Wrap the value returned from `f` in a result object.
```
zzzeid, line 87: Is this meant to be a "result dictionary"?

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+        )
+        session.add(log)
+
+    cache.delete_memoized(get_tree_by_name, tree.tree)
+
+
+@result_object_wrap
+def get_stack() -> list[dict]:
+    """Handler for `GET /stack`."""
+    return [
```
zzzeid, line 267: Would be good to add an explicit status code (200) to these methods.

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+            "Must specify status, reason and tags to remember the change.",
+            "Must specify status, reason and tags to remember the change.",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
+
+    # Update the trees as requested.
+    new_status = body.get("status")
+    new_reason = body.get("reason")
+    new_motd = body.get("message_of_the_day")
+    new_tags = body.get("tags", [])
```
zzzeid, line 434: `status`, `reason`, `message_of_the_day`, and `tags` are repeated quite a bit in this method. I think if this is refactored to reuse these keys in a more generic way, it would make this a bit easier to read. E.g.,
```suggestion
    keys = ("status", "reason", "tags")
    new_values = {k: body.get(k) for k in keys + ("message_of_the_day",)}
```

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+
+    old_trees = {}
+
+    for tree in trees:
+        old_trees[tree.tree] = {}
+        old_trees[tree.tree]["status"] = tree.status
+        old_trees[tree.tree]["reason"] = tree.reason
+        old_trees[tree.tree]["tags"] = tree.tags
+        old_trees[tree.tree]["log_id"] = tree.log_id
```
zzzeid, line 443: Continued from above comment...

```suggestion
        old_trees[tree.tree] = {key: getattr(tree, key) for key in keys}
        old_trees[tree.tree]["log_id"] = tree.log_id
```

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+
+        update_tree_status(
+            session,
+            tree.model,
+            status=new_status,
+            reason=new_reason,
+            message_of_the_day=new_motd,
+            tags=new_tags,
```
zzzeid, line 451: Continued from above...

```suggestion
            **new_values
```


Etc...

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+                continue
+
+            if reason:
+                last_state["current_reason"] = reason
+            if tags:
+                last_state["current_tags"] = tags
+
+            tree.last_state = json.dumps(last_state)
+
+    session.commit()
```
zzzeid, line 567: I've noticed this in a few places. In this method (and others), it's only being used once, and could probably just do `db.session.commit()` then and there instead.

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+
+    return None, 200
+
+
+def get_logs_for_tree(tree_name: str, limit_logs: bool = True) -> list[dict]:
```
zzzeid, line 572: Since this defaults to `True`, and uses a the default value (`TREE_SUMMARY_LOG_LIMIT`) might as well just change it to be an integer.
```suggestion
def get_logs_for_tree(tree_name: str, limit_logs: int = TREE_SUMMARY_LOG_LIMIT) -> list[dict]:
```

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,609 @@
...
+@result_object_wrap
+def get_logs_all(tree: str) -> list[dict]:
+    """Handler for `GET /trees/{tree}/logs_all`."""
+    return get_logs_for_tree(tree, limit_logs=False)
+
+
+@result_object_wrap
+def get_logs(tree: str) -> list[dict]:
+    """Handler for `GET /trees/{tree}/logs`."""
+    return get_logs_for_tree(tree, limit_logs=True)
```
zzzeid, line 603: I assume the reason we are using a different URL here is for compatibility? It would be a bit nicer to have a single endpoint (instead of this one and `get_logs_all`, and just pass a limit parameter instead.

landoapi/auth.py
```diff
@@ -29,6 +29,10 @@
 
 logger = logging.getLogger(__name__)
 
+MOZILLIANS_PREFIX = "mozilliansorg_"
+TREESTATUS_USERS = f"{MOZILLIANS_PREFIX}treestatus_users"
+TREESTATUS_ADMIN = f"{MOZILLIANS_PREFIX}treestatus_admins"
```
zzzeid, line 34:
```suggestion
TREESTATUS_PREFIX = "mozilliansorg_treestatus_"
TREESTATUS_USERS = f"{TREESTATUS_PREFIX}users"
TREESTATUS_ADMIN = f"{TREESTATUS_PREFIX}admins"
```

landoapi/auth.py
```diff
@@ -312,6 +316,16 @@ def is_in_groups(self, *args: str) -> bool:
 
         return set(args).issubset(self.groups)
 
+    def user_id(self) -> str:
+        """Return the user ID for the authenticated user."""
+        if not self._userinfo:
+            raise ValueError("User ID requested for user without userinfo.")
+
+        if "sub" not in self._userinfo:
```
zzzeid, line 324: Can you add a comment describing what `sub` means?

landoapi/auth.py
zzzeid: I think this would be better refactored to handle required arguments a little more conventionally. The above code (re: `scopes`) also does not make much sense. If an argument is required, it should not be marked as `Optional` and should be changed to a positional argument.

landoapi/landing_worker.py
zzzeid: Anything stopping us from actually including the database trees as part of `repo_clone_subsystem.repos`? I.e., mapping `repo_clone_subsystem.repos[r].lando_tree` to a `Tree` object, or something similar.

zzzeid (changes requested):
Second pass with more to come.

landoapi/auth.py
cgsheeh: The group names in Mozillians are `treestatus_users` and `treestatus_admins`, and the prefix is an implementation detail of how the groups are returned from SSO. I'd like to keep the `treestatus_` part outside of the Mozillians prefix since including it there isn't an accurate reflection of how they're represented in SSO.

I could make a variable like `USERS_GROUP = "treestatus_users"` and then do `USER_MOZILLIANS_GROUP = f"{MOZILLIANS_PREFIX}{USERS_GROUP}"` if that makes things clearer, or maybe just add a comment.

landoapi/landing_worker.py
cgsheeh: Tying the `Tree` model to a `Repo` is something I would like to do, since they are mostly referring to the same concept. I'll take a look at trying to do this, but unless we can think of a technical reason that these models couldn't be merged later, I don't think it's a hard blocker for this project.

landoapi/auth.py
cgsheeh: The `scopes` assert is fixed in #311, and I'll make the check here an `if` statement instead.

landoapi/api/treestatus.py
cgsheeh: It's easier to read in the declarative style, you can see exactly what the output will look like.

landoapi/api/treestatus.py
cgsheeh: The output from the API will be JSON, which usually uses `object`. I'll switch it to `dict`.

landoapi/api/treestatus.py
cgsheeh: Yes, it's for compatibility, since this is a publicly accessible endpoint. Once we migrate over I'm going to re-audit usage of the endpoints and hopefully we can consolidate into a single endpoint.

landoapi/api/treestatus.py
cgsheeh: This tiny change took a ton of effort to fix, since it's very hard to mock out timestamps when created by `db.func.now()`, as `db.func` is a dynamic DB-specific object in SQLAlchemy. I ended up re-writing most of the tests to use `pydantic` schema validation instead of just checking `dict == dict`, but in the end I think they're easier to understand and make changes to. :)

zzzeid (changes requested):
NOTE: Branch not up to date with `main` at time of code review.

landoapi/api/treestatus.py
```diff
@@ -0,0 +1,619 @@
...
+            404,
+            f"No tree {tree_name} found.",
+            "The tree does not exist.",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404",
+        )
+    session.delete(tree)
+
+    # Delete from logs and change stack, too.
+    Log.query.filter_by(tree=tree_name).delete()
+    StatusChangeTree.query.filter_by(tree=tree_name).delete()
```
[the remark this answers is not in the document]
cgsheeh: Good point, this was carried over from the original Treestatus but it's not really necessary or desirable.

landoapi/models/treestatus.py
```diff
@@ -6,25 +6,63 @@
...
+    RELEASE_STABILIZATION = "release_stabilization"
+    TRY = "try"
+    COMM_REPOS = "comm_repos"
+    OTHER = "other"
+
+
+class TreeStatus(enum.Enum):
+    """Allowable statuses of a tree."""
+
+    OPEN = "open"
```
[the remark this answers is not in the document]
cgsheeh: That's not a convention, the value of a enum can be anything including objects/classes etc.

The values are what the API is expecting to be returned as values for the `status` key.

landoapi/models/treestatus.py
```diff
@@ -6,25 +6,63 @@
...
+    db,
 )
 
-DEFAULT_TREE = {"reason": "New tree", "status": "closed", "tags": [], "log_id": None}
+
+class TreeCategory(enum.Enum):
+    """Categories of the various trees.
+
+    Note: the definition order is in order of importance for display in the UI.
+    Note: this class also exists in Lando-UI, and should be updated in both places.
```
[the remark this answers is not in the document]
cgsheeh: I doubt this will change between the Treestatus migration and our UI/API consolidation with the Django migration. I'm going to leave it as-is and the duplication will disappear once that work is completed.

landoapi/api/treestatus.py
cgsheeh: I added the status code to the `result_object_wrap`, so that decorator implies a 200 now.

landoapi/api/treestatus.py
```diff
@@ -110,20 +104,18 @@ def wrap_output(*args, **kwargs) -> tuple[dict[str, Any], int]:
     return wrap_output
 
 
-def serialize_last_state(old_tree: dict, new_tree: CombinedTree) -> str:
+def serialize_last_state(old_tree: dict, new_tree: CombinedTree) -> dict[str, Any]:
     """Serialize a `last_state` value for a `StatusChangeTree`."""
```
zzzeid, line 108: nit: this docstring is missing some context, and doesn't impart any information other than what is in the name of this method. Perhaps add some additional documentation.

landoapi/models/treestatus.py
```diff
@@ -53,21 +52,23 @@ def is_open(self) -> bool:
...
+        "reason": "New tree",
+        "status": TreeStatus.CLOSED,
+        "tags": [],
+        "log_id": None,
+    }
 
 
-def load_last_state(last_state_str: str) -> dict:
+def load_last_state(last_state_orig: dict) -> dict:
     """Ensure that structure of last_state is backwards compatible."""
```
zzzeid, line 69: nit: could add more context here as to what it should be backwards compatible with.

landoapi/models/treestatus.py
```diff
@@ -150,31 +143,14 @@ class Log(Base):
...
-    def __init__(self, tags: Optional[list[str]] = None, **kwargs):
-        if tags is not None:
-            kwargs["_tags"] = json.dumps(tags)
-        super(Log, self).__init__(**kwargs)
-
-    @sqlalchemy.ext.hybrid.hybrid_property
-    def tags(self) -> list[str]:
-        """Handle conversion of the `tags` column to a list."""
-        return json.loads(self._tags)
+    tags = db.Column(JSONB, nullable=False)
```
zzzeid, line 153: Does this need a default?

landoapi/models/treestatus.py
```diff
@@ -150,31 +143,14 @@ class Log(Base):
...
     # The status which the tree has been set to.
-    status = db.Column(
-        db.Enum(
-            TreeStatus,
-            # Use the values of the enum in the DB instead of the keys.
-            values_callable=lambda x: [i.value for i in x],
-        ),
-        nullable=False,
-    )
+    status = db.Column(db.Enum(TreeStatus), nullable=False)
```
zzzeid, line 146: Does this need a default?

landoapi/models/treestatus.py
```diff
@@ -198,14 +174,7 @@ class StatusChange(Base):
...
     # The status the trees were changed to.
-    status = db.Column(
-        db.Enum(
-            TreeStatus,
-            # Use the values of the enum in the DB instead of the keys.
-            values_callable=lambda x: [i.value for i in x],
-        ),
-        nullable=False,
-    )
+    status = db.Column(db.Enum(TreeStatus), nullable=False)
```
zzzeid, line 177: Does this need a default?

landoapi/models/treestatus.py
```diff
@@ -234,9 +203,9 @@ class StatusChangeTree(Base):
         db.String(64), db.ForeignKey(Tree.tree), nullable=False, index=True
     )
 
-    # A JSON encoded string containing the previous state of the tree before
+    # A JSON object containing the previous state of the tree before
     # applying this change.
-    last_state = db.Column(db.Text, nullable=False)
+    last_state = db.Column(JSONB, nullable=False)
```
zzzeid, line 208: Does this need a default?

zzzeid (changes requested):
Looking good, just a few minor comments.

landoapi/models/treestatus.py
cgsheeh: No, there shouldn't be a default here - `status` isn't set that should be considered a programming error.

landoapi/models/treestatus.py
cgsheeh: No, a missing status here is a programming error.

landoapi/models/treestatus.py
cgsheeh: No, a missing status is a programming error.

landoapi/models/treestatus.py
cgsheeh: Yes, the default should be `list`.

migrations/versions/1dfc5aa9d1b2_add_treestatus_models.py
zzzeid, line 1: Unfortunately alembic does not pick up default JSONB changes, and those have to be added manually. This would be something like what I have done in [56c6748ee7cf_add_revision_data.py](https://github.com/mozilla-conduit/lando-api/blob/main/migrations/versions/56c6748ee7cf_add_revision_data.py). I.e., 
```python
server_default="{}"
```

zzzeid (changes requested):
Look good, just the manual migration update and this is good to go.

migrations/versions/1dfc5aa9d1b2_add_treestatus_models.py
cgsheeh: Ah that's annoying. Do you have a bug report somewhere the describes this? It would be nice to add it to the `migrations/README.md` file for future reference.

zzzeid: I believe at the time I tried different approaches and this was the one that worked with our setup. Definitely worth documenting it somewhere, but I don't think there's a bug filed anywhere describing this.

requirements.txt
```diff
@@ -853,9 +853,9 @@ typing-extensions==4.8.0 \
     #   black
     #   pydantic
     #   pydantic-core
-urllib3==1.26.16 \
-    --hash=sha256:8d36afa7616d8ab714608411b4a3b13e58f463aee519024578e062e141dce20f \
-    --hash=sha256:8f135f6502756bde6b2a9b28989df5fbe87c9970cecaa69041edcce7f0589b14
+urllib3==1.26.18 \
```
zzzeid, line 856: I assume this is intentional in this PR?

zzzeid (approved):

requirements.txt
cgsheeh: That change is just from my merging the latest `main` into this PR. If you look at the full diff view (ie not just the latest commits) you can see there isn't any change to `urllib3`.

landoapi/app.py
```diff
@@ -111,3 +111,19 @@ def construct_app(config: dict[str, Any]) -> connexion.App:
...
+
+
+def construct_treestatus_app(config: dict[str, Any]) -> connexion.App:
+    app = connexion.App(__name__, specification_dir="spec/")
+
+    app.add_api(
+        "treestatus.yml",
+        resolver=RestyResolver("landoapi.api"),
+        options={"swagger_ui": False},
+    )
```
zzzeid, line 123: As mentioned elsewhere, I think it makes sense to simply add the treestatus spec to the existing app, I don't see the need to create a completely separate app instance, as this causes code duplication.

Even if you were to stay with the single API spec per app approach, you could instead just decide on which API spec to add to the app based on the command or command parameters. E.g.,

```python
def construct_app(config: dict[str, Any], spec="swagger.yml") -> connexion.App:
    app = connexion.App(__name__, specification_dir="spec/")
    app.add_api(
        spec,
        resolver=RestyResolver("landoapi.api"),
        options={"swagger_ui": False},
    )
    flask_app = app.app
    flask_app.config.update(config)
    flask_app.register_blueprint(dockerflow)
    initialize_hooks(flask_app)

    return app
```

And then you can call `constuct_app(config, "treestatus.yml")` where relevant based on the command.

zzzeid (changes requested):
Left some comments mainly around the implementation of the app/api/cli.

Since everything has to go through the nginx proxy, I do think it would be simpler to namespace all the endpoints as I mentioned previously (even if they are in a separate API spec), and determine the actual public URLs in the nginx configuration (e.g., with rewrites), if they should be different for any reason (either temporarily or permanently) on that specific container. This would remove the need to implement separate apps, clis, apis, and nginx configurations for treestatus.

I think we should be planning on combining these APIs, particularly since there aren't going to be many other public endpoints in Lando in the near future. I also don't see an actual clash of endpoint definitions (even though the naming is quite similar and can be confusing, e.g. in the case of `/stack` and `/stacks`).

landoapi/cli.py
```diff
@@ -159,5 +174,93 @@ def test(pytest_arguments):
...
+        return TreeStatus(status)
+    except ValueError:
+        if status == "approval require":
+            return TreeStatus.APPROVAL_REQUIRED
+
+    return TreeStatus.OPEN
+
+
+@cli.command("import-treestatus")
+def import_treestatus_data():
```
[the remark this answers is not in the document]
cgsheeh: Added a note and filed https://bugzilla.mozilla.org/show_bug.cgi?id=1894984 to track this. I also added it to the migration plan.

cgsheeh:
> Left some comments mainly around the implementation of the app/api/cli.
> 
> Since everything has to go through the nginx proxy, I do think it would be simpler to namespace all the endpoints as I mentioned previously (even if they are in a separate API spec), and determine the actual public URLs in the nginx configuration (e.g., with rewrites), if they should be different for any reason (either temporarily or permanently) on that specific container. This would remove the need to implement separate apps, clis, apis, and nginx configurations for treestatus.

For the initial migration, "no changes required for clients" was an explicit requirement, which is why we've done this work to undo the `/treestatus/` namespacing. Previously we did have the Treestatus API endpoints in the Swagger spec as a single app with `/treestatus/` namespacing, with a plan to add Nginx path rewriting so RelEng Treestatus client requests would be redirected to their Lando-Treestatus equivalents. After discussion with our SRE team we decided it was easier to stand up a separate Treestatus API container instead of using nginx and doing path rewriting. This also gives us the advantage of being able to update the DNS entry for `treestatus.mozilla-releng.net` and point it at the new Lando-Treestatus, and existing clients will continue to work as expected without any changes during the migration.

> I think we should be planning on combining these APIs, particularly since there aren't going to be many other public endpoints in Lando in the near future. I also don't see an actual clash of endpoint definitions (even though the naming is quite similar and can be confusing, e.g. in the case of `/stack` and `/stacks`).

I agree, having one set of Lando APIs would be much better. When many of the existing Lando API endpoints are obsoleted in Django-Lando, we could combine the APIs without namespacing.

landoapi/app.py
cgsheeh: I described why we need to load the Treestatus app as a separate, non-namespaced app in my comment reply to your overall review comment. I have added the `spec` argument to `construct_app` as suggested.

landoapi/wsgi.py
```diff
@@ -4,10 +4,19 @@
...
+import os
+
+from .app import SUBSYSTEMS, construct_app, construct_treestatus_app, load_config
+
+# Determine which app to construct by looking for a Treestatus specific env variable.
+app_constructor = (
+    construct_treestatus_app
+    if os.getenv("TREESTATUS_APP") is not None
+    else construct_app
+)
```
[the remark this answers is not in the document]
cgsheeh: Updated so `construct_app` takes an argument about which API spec to add to the app instead of a separate app constructor.

landoapi/cli.py
```diff
@@ -36,10 +44,17 @@ def get_subsystems(exclude: Optional[list[Subsystem]] = None) -> list[Subsystem]
...
 def create_lando_api_app() -> connexion.App:
-    from landoapi.app import construct_app, load_config
+    from landoapi.app import construct_app, construct_treestatus_app, load_config
+
+    # Determine which app to construct by looking for a Treestatus specific env variable.
+    app_constructor = (
+        construct_treestatus_app
+        if os.getenv("TREESTATUS_APP") is not None
+        else construct_app
+    )
```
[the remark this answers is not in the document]
cgsheeh: Part of the reason for using an environment variable here is that the `wsgi.py` loading doesn't get access to the CLI arguments. Since we need to load the spec into the app when we run `wsgi.py`, I went with an environment variable. The PR to add this config has been landed in the upstream cloudops-infra repo, where using a separate environment variable also has the advantage of allowing us to re-use the config for the existing LandoAPI app, with an extra environment variable added to the YAML configs.

landoapi/cli.py
```diff
@@ -159,5 +174,93 @@ def test(pytest_arguments):
...
+
+    db.session.flush()
+    print("Flushing.")
+
+    # Create log entries for each update in the trees file
+    for tree, logs in tree_logs.items():
+        print(f"Importing logs for {tree}.")
+        for log_entry in reversed(logs["result"]):
+            log = Log(
+                tree=log_entry["tree"],
```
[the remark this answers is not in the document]
cgsheeh: It will always be the same in this loop. You're right that it would be more correct to use `tree.tree` here, but since I've already tested it on dev and making this change now would mean wiping the dev DB to re-test and ensure this works as expected, I'd rather keep it as-is.

landoapi/wsgi.py
```diff
@@ -6,17 +6,13 @@
...
-from .app import SUBSYSTEMS, construct_app, construct_treestatus_app, load_config
+from .app import SUBSYSTEMS, construct_app, load_config
 
 # Determine which app to construct by looking for a Treestatus specific env variable.
-app_constructor = (
-    construct_treestatus_app
-    if os.getenv("TREESTATUS_APP") is not None
-    else construct_app
-)
+spec = "treestatus.yml" if os.getenv("TREESTATUS_APP") is not None else "swagger.yml"
```
zzzeid, line 12: When moving this to django lando, it would be better handled via a command vs an environment variable, but this is an acceptable compromise for now after simplifying the app construction.

However, this config variable should be implemented in `app.load_config` (e.g., in the `config` variable) and not fetched directly from the environment in this module.
```suggestion
spec = config["API_SPEC"]
```

And since that would be done there, it would be better to have it as a string (spec) rather than a boolean.

landoapi/cli.py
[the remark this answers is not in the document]
zzzeid: I think this leaves some room for potential errors, but if you've tested the script already with production data then that potential would be remediated.

landoapi/wsgi.py
zzzeid, line 12: When moving this to django lando, it would be better handled via a command vs an environment variable, but this is an acceptable compromise for now after simplifying the app construction.

However, this config variable should be implemented in `app.load_config` (e.g., in the `config` variable) and not fetched directly from the environment in this module.

```suggestion
spec = config["API_SPEC"]
```

And since that would be done there, it would be better to have it as a string (spec) rather than a boolean.

zzzeid (changes requested):
> Previously we did have the Treestatus API endpoints in the Swagger spec as a single app with /treestatus/ namespacing, with a plan to add Nginx path rewriting so RelEng Treestatus client requests would be redirected to their Lando-Treestatus equivalents. After discussion with our SRE team we decided it was easier to stand up a separate Treestatus API container instead of using nginx and doing path rewriting.

Though it may be easier in the short term, it's not necessarily easier in the long term and will likely produce at least some technical debt that we will have to deal with later -- in this case, at least with URL dispatching and the app configuration. For what its worth, having it in a separate container is not the issue as we can still do that with the path rewriting approach or this approach.

Please loop me into any Lando architectural decision meetings in the future. This will save us time from uncovering these things in the code review / later stages of a project. We'll have to set up a discussion on transitioning this to a more simplified implementation in the new Lando, but this will work fine for this transitional stage.

landoapi/wsgi.py
cgsheeh: Makes sense to move the fetching into `load_config`. Since we already have the SRE side changes looking for a `TREESTATUS_APP` environment variable, I set the `API_SPEC` based on the logic above that looks for `TREESTATUS_APP`. We can update the changes in our deployment configs to make the environment variable align with the config knob after the migration is complete, but with it being so close I'd like to make as few changes as possible.

cgsheeh:
> Please loop me into any Lando architectural decision meetings in the future. This will save us time from uncovering these things in the code review / later stages of a project. We'll have to set up a discussion on transitioning this to a more simplified implementation in the new Lando, but this will work fine for this transitional stage.

I can't remember if I explicitly invited you to that meeting, so apologies for that. I did organize the meeting openly in our Conduit team channel in early January, with an explicit comment that I wanted to discuss Lando related things (Treestatus as well as the Lando-Try landing worker). I discussed the contents of that meeting in our team meeting on Jan 15, and discussed the details of the implementation on February 5. I thought I had my bases covered in terms of communicating what was happening and why, but an earlier re-review request would have solved this.

landoapi/app.py
```diff
@@ -54,6 +54,9 @@ def load_config() -> dict[str, Any]:
         "MAIL_USE_TLS": bool(os.getenv("MAIL_USE_TLS")),
         "SQLALCHEMY_DATABASE_URI": os.getenv("DATABASE_URL"),
         "SQLALCHEMY_TRACK_MODIFICATIONS": False,
+        "APP_SPEC": "treestatus.yml"
+        if os.getenv("TREESTATUS_APP") is not None
```
zzzeid, line 58: This should be explicit, i.e.,
```suggestion
        if os.getenv("TREESTATUS_APP") == "1"
```

Otherwise values like "0" and "false" will also behave in the same way.

tests/conftest.py
```diff
@@ -254,14 +254,16 @@ def init_app(self, app, db):
 def app(request, versionfile, docker_env_vars, disable_migrations, mocked_repo_config):
     """Needed for pytest-flask."""
     module = request.module
-    spec = "swagger.yml" if not hasattr(module, "TREESTATUS_APP") else "treestatus.yml"
 
     config = load_config()
+    config["API_SPEC"] = (
+        "swagger.yml" if not hasattr(module, "TREESTATUS_APP") else "treestatus.yml"
```
zzzeid, line 260: Same thing here, we shouldn't only be checking if the attribute is there or not, we should be checking the value of the attribute. However, this also circumvents the functionality in `load_config`, which should be tested. Instead, `TREESTATUS_APP` should be monkeypatched in ` docker_env_vars`, similar to how other environment variables are set.

landoapi/wsgi.py
zzzeid: Yeah this is fine, in terms of which value is in the environment for now.

zzzeid (changes requested):

tests/conftest.py
cgsheeh: Good call, I had the config knob as `API_SPEC` in one location and `APP_SPEC` in another, so this already saved me time debugging in dev.

tests/conftest.py
```diff
@@ -177,6 +177,14 @@ def docker_env_vars(versionfile, monkeypatch):
...
     monkeypatch.delenv("MAIL_SUPPRESS_SEND", raising=False)
 
+    treestatus_app = (
+        "1"
+        if hasattr(request.module, "TREESTATUS_APP")
+        and request.module.TREESTATUS_APP is True
+        else ""
+    )
+    monkeypatch.setenv("TREESTATUS_APP", treestatus_app)
+
```
zzzeid, line 187: This is still a little awkward, but I don't think there's a way to completely remove this awkwardness with the current implementation taking the environment variable approach. Ideally things can be done cleanly in test fixtures with the behaviour unchanged from module to module. Probably the only way to do this cleanly here would be to rewrite the `app` fixture in `test_treestatus`.

However given that some of this functionality will be ported/reimplemented in django-lando, an interim clean up here would suffice -- I would just remove the `test_treestatus.TREESTATUS_APP` variable altogether and just check the module name.


```suggestion
    if request.module.__name__ == "tests.test_treestatus":
        monkeypatch.setenv("TREESTATUS_APP", treestatus_app)
```

zzzeid (changes requested):
Looking good, this is very close to the finish line.

tests/test_treestatus.py
```diff
@@ -1,113 +1,1118 @@
...
     TreeStatus,
-    TreeStatusCommunicationException,
-    TreeStatusError,
+)
+from landoapi.treestatus import (
+    is_open,
 )
 
+# Used to load the Treestatus app in tests instead of LandoAPI.
+TREESTATUS_APP = True
```
zzzeid, line 26: I don't think this is in use anymore

```suggestion
```

zzzeid (changes requested):

zzzeid (approved):

PR #285, opened 2023-03-13
configuration: don't fetch previous value if initializing (bug 1822080)

zzzeid:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1822080.


cgsheeh (approved):
It would be nice to have a log entry for when we initialize as well.

zzzeid:
> It would be nice to have a log entry for when we initialize as well.

There is one already, on line 115.

cgsheeh:
> > It would be nice to have a log entry for when we initialize as well.
> 
> There is one already, on line 115.

One line above the preview, naturally. ;)

PR #288, opened 2023-03-15
stacks: raise error if stack data has incorrect length (bug 1822588)

zzzeid:


cgsheeh (changes requested):
We should add a test for this code path.

We also call `request_extended_revision_data` in a few other places. Should we copy this pattern to those locations as well?

cgsheeh:
> > We should add a test for this code path.
> 
> Still a WIP (draft) -- will add a test for sure.
> 
> Note that I don't think we have any mechanism for adding authentication capabilities to the mock `PhabricatorClient`. Something we should add in the future, so the tests will be somewhat limited by this.
> 
> > We also call `request_extended_revision_data` in a few other places. Should we copy this pattern to those locations as well?
> 
> Not sure about this -- I think all other usages require the page to be loaded already, and are done on landing / uplift triggers? Should in theory fail long before this in those cases, but I'll look at what we've done in the past with the edge search.

Ah, I didn't notice the draft state. The fix itself looks good, FWIW.

zzzeid:
> We also call request_extended_revision_data in a few other places. Should we copy this pattern to those locations as well?

I checked the other two spots. Previously in `landoapi.uplift`, we were just raising a `ValueError` which gets caught in `landoapi.api.uplift`. Since we already raise a `ValueError` in this implementation, this should behave exactly the same --  see [here](https://github.com/mozilla-conduit/lando-api/commit/72c8655332055ee3ebaea7c7a0d026da9fdb3754#diff-4b283b5641edfe5ab9919481899d871710929836bbc444cbfd77d956b4fa5758L138). In the other instance (`landoapi.api.transplants`), this was not handled anyway as there was no try/except to start with.

cgsheeh (approved):
LGTM.

We have some `test_request_extended_revision_data*` tests, we could add one that checks `pytest.raises(ValueError)` under the expected conditions, but I'm not too fussed about it if we have an API test.

cgsheeh (approved):
r+

PR #289, opened 2023-03-22
landing_worker: move merge conflict processing to function (bug 1823898)

zzzeid:
Move merge conflict processing to its own function.

### DIFF
--- a/landoapi/workers/landing_worker.py
+++ b/landoapi/workers/landing_worker.py
@@ -149,13 +149,13 @@ class LandingWorker(Worker):
         )
 
     def process_merge_conflict(
-            self,
-            exception: PatchConflict,
-            repo: Repo,
-            hgrepo: HgRepo,
-            revision_id: int,
+        self,
+        exception: PatchConflict,
+        repo: Repo,
+        hgrepo: HgRepo,
+        revision_id: int,
     ):
-        "Extract and parse merge conflict data from exception into a usable format."""
+        "Extract and parse merge conflict data from exception into a usable format." ""
         failed_paths, reject_paths = self.extract_error_data(str(exception))
 
         # Find last commits to touch each failed path.
--- a/landoapi/workers/landing_worker.py
+++ b/landoapi/workers/landing_worker.py
@@ -148,6 +148,56 @@ class LandingWorker(Worker):
             job.requester_email, job.head_revision, job.error, job.id
         )
 
+    def process_merge_conflict(self, exception, repo, hgrepo, revision_id):
+        failed_paths, reject_paths = self.extract_error_data(str(exception))
+
+        # Find last commits to touch each failed path.
+        failed_path_changesets = [
+            (
+                path,
+                hgrepo.run_hg(
+                    [
+                        "log",
+                        "--cwd",
+                        hgrepo.path,
+                        "--template",
+                        "{node}",
+                        "-l",
+                        "1",
+                        path,
+                    ]
+                ),
+            )
+            for path in failed_paths
+        ]
+
+        breakdown = {
+            "revision_id": revision_id,
+            "content": None,
+            "reject_paths": None,
+        }
+
+        breakdown["failed_paths"] = [
+            {
+                "path": r[0],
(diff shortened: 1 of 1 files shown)


landoapi/workers/landing_worker.py
```diff
@@ -148,6 +148,56 @@ def notify_user_of_landing_failure(job):
...
+        breakdown["failed_paths"] = [
+            {
+                "path": r[0],
+                "url": f"{repo.pull_path}/file/{r[1].decode('utf-8')}/{r[0]}",
+                "changeset_id": r[1].decode("utf-8"),
+            }
+            for r in failed_path_changesets
+        ]
+        breakdown["reject_paths"] = {}
+        for r in reject_paths:
```
cgsheeh, line 189: r -> `path`, or `reject_path`

cgsheeh (changes requested):
How difficult would it be to add a test for this function? If possible we should add one now.

landoapi/workers/landing_worker.py
```diff
@@ -148,6 +148,56 @@ def notify_user_of_landing_failure(job):
             job.requester_email, job.head_revision, job.error, job.id
         )
 
+    def process_merge_conflict(self, exception, repo, hgrepo, revision_id):
```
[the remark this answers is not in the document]
zzzeid: I think it makes sense to add `flake8-docstrings` to help with these checks. I've already added it to other repos, and it makes sense to enforce this as a standard. Will file a bug.

zzzeid:
> > How difficult would it be to add a test for this function? If possible we should add one now.
> 
> There are already integration tests that test some of this code (e.g., `test_landings.py::test_landing_worker__extract_error_data`). We could add a unit test for this in the future but I think it's out of scope for this PR (as this is mainly just extracted from #224 and mainly just moves the exact code block to its own function.)

Also I am planning some improvements to this code as it will ultimately be reused by the revision worker to detect conflicts. That might be a better time to write unit tests for this.

landoapi/workers/landing_worker.py
```diff
@@ -148,6 +148,63 @@ def notify_user_of_landing_failure(job):
...
         )
 
+    def process_merge_conflict(
+            self,
+            exception: PatchConflict,
+            repo: Repo,
+            hgrepo: HgRepo,
+            revision_id: int,
+    ):
+        "Extract and parser merge conflict data from exception into a usable format."""
```
zzzeid, line 158:
```suggestion
        "Extract and parse merge conflict data from exception into a usable format."""
```

landoapi/workers/landing_worker.py
```diff
@@ -148,6 +148,63 @@ def notify_user_of_landing_failure(job):
             job.requester_email, job.head_revision, job.error, job.id
         )
 
+    def process_merge_conflict(
+        self,
+        exception: PatchConflict,
+        repo: Repo,
+        hgrepo: HgRepo,
+        revision_id: int,
+    ):
```
cgsheeh, line 157:
```suggestion
    def process_merge_conflict(
        self,
        exception: PatchConflict,
        repo: Repo,
        hgrepo: HgRepo,
        revision_id: int,
    ) -> dict[str, Any]:
```

cgsheeh (changes requested):

landoapi/workers/landing_worker.py
```diff
@@ -148,6 +149,63 @@ def notify_user_of_landing_failure(job):
...
         )
 
+    def process_merge_conflict(
+        self,
+        exception: PatchConflict,
+        repo: Repo,
+        hgrepo: HgRepo,
+        revision_id: int,
+    ) -> dict[str, Any]:
+        "Extract and parse merge conflict data from exception into a usable format." ""
```
cgsheeh, line 159:
```suggestion
        """Extract and parse merge conflict data from exception into a usable format."""
```

cgsheeh (changes requested):

landoapi/workers/landing_worker.py
zzzeid: Nice catch, interesting how this slipped through the linters, most likely black automatically added the space at the end.

PR #290, opened 2023-03-22
landing_worker: simplify exception handling when landing (bug 1823899)

zzzeid:
Refactor and simplify how exceptions are handling when pushing.

### DIFF
--- a/landoapi/workers/landing_worker.py
+++ b/landoapi/workers/landing_worker.py
@@ -322,14 +322,18 @@ class LandingWorker(Worker):
                     )
                     self.notify_user_of_landing_failure(job)
                     return True
-                except NoDiffStartLine:
-                    logger.exception("Patch without a diff start line.")
+                except NoDiffStartLine as e:
                     message = (
                         "Lando encountered a malformed patch, please try again. "
-                        "If this error persists please file a bug."
+                        "If this error persists please file a bug: "
+                        "Patch without a diff start line."
                     )
+                    logger.error(message)
                     job.transition_status(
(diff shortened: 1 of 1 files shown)


cgsheeh (changes requested):

landoapi/workers/landing_worker.py
```diff
@@ -388,43 +407,34 @@ def run_job(
                 "utf-8"
             )
 
+            temporary_exceptions = {
+                TreeClosed: f"Tree {repo.tree} is closed - retrying later.",
+                TreeApprovalRequired: f"Tree {repo.tree} requires approval - retrying later.",
+                LostPushRace: f"Lost push race when pushing to {repo.push_path}.",
+            }
+
```
cgsheeh, line 415: These error messages should be more tightly coupled to the exceptions themselves, either in the exception class definition or passed in the location they are raised (ie `raise LostPushRace(f"Lost push race when pushing to {repo.push_path}.")`).

That might mean we need to pass the whole `repo` object to `hgrepo.push` to create these error messages, or perhaps these three classes should all subclass some `TemporaryFailureException` that has a `.get_error_message(repo)` function.

cgsheeh (changes requested):

landoapi/workers/landing_worker.py
zzzeid: Changed it around a little, I think this last implementation is a good compromise. We don't need to stick with the original text since all we need is the repo info in the exception, and the exception names are pretty descriptive.

cgsheeh (approved):
LGTM! :)

PR #292, opened 2023-03-23
style: apply ruff to codebase (bug 1824234)

zzzeid:


landoapi/patches.py
```diff
@@ -0,0 +1,154 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from __future__ import annotations
```
dklawren: This file looks like something other than just ruff related refactoring. Was this intentional?

dklawren (changes requested):

landoapi/patches.py
zzzeid: Not intentional, already removed. The only functional change in this PR should be in `tests/test_py3_codequality.py`.

dklawren (changes requested):
Looks good. But running locally, `test_py3_codequality.py::test_check_python_ruff` fails with 

```python
>               raise child_exception_type(errno_num, err_msg, err_filename)
E               FileNotFoundError: [Errno 2] No such file or directory: 'ruff'
```

This was using `invoke test`. Is ruff installed in the container?

zzzeid:
> This was using `invoke test`. Is ruff installed in the container?

It should be. Did you run `docker-compose build` first?

dklawren:
> Looks good. But running locally, `test_py3_codequality.py::test_check_python_ruff` fails with
> 
> ```python
> >               raise child_exception_type(errno_num, err_msg, err_filename)
> E               FileNotFoundError: [Errno 2] No such file or directory: 'ruff'
> ```
> 
> This was using `invoke test`. Is ruff installed in the container?

Nevermind. I think the test was running against an old container. Rebuilding and then running the test again worked.

dklawren (approved):
LGTM r=dkl

cgsheeh (approved):
ruff <3

PR #293, opened 2023-03-24
requirements: pin requirements explicity in `requirements.in` (Bug 1824006)

cgsheeh:
Each of the added requirements is used at least once in Lando's
core codebase.


zzzeid (changes requested):
You should run `build-requirements` since there are packages now that are sourced from `requirements.in` vs. as dependencies of other packages, thus `requirements.txt` is no longer accurate.

zzzeid (approved):

PR #298, opened 2023-03-29
uplift: make Bugzilla status tracking flag configurable per-repo (Bug 1825091)

cgsheeh:
The status tracking flags for ESR differ slightly from beta and release.
Make the tracking flag configurable per-repo instead of hard-coded to
a specific format. On the Bugzilla side the uplift API endpoint allows
changing any `cf_status_*` flag.


dklawren (approved):
LGTM. r=dkl

zzzeid (approved):

zzzeid:
Not something to do in this PR, but I am thinking now that we have a few flags that aren't relevant to every repo, maybe we should find a better place for these flags.

PR #307, opened 2023-05-03
requirements: upgrade flask and other requirements (bug 1831185)

zzzeid:
- Update Flask and related packages
- Update test code to reflect changes in Flask


zzzeid:
> CI is red, looks like some changes to Flask internals need to be reflected in Lando.
> 
> Can you file a bug for this? I also noticed our version of Flask (via `connexion`) is quite old in Lando-UI, can we upgrade those packages also?

Yep noticed this after CI ran, though locally the test initially passed (I think this was a cache issue). Fixing this up. I do have another PR for lando-ui coming up, but it has very outdated packages and thus more changes.

tests/test_dockerflow.py
```diff
@@ -4,7 +4,7 @@
 
 import json
 
-from tests.utils import phab_url, trans_url
```
zzzeid, line 7: `trans_url` no longer used (and is deprecated functionality).

tests/test_dockerflow.py
```diff
@@ -23,9 +23,8 @@ def test_dockerflow_version_matches_disk_contents(client, versionfile):
 
 
 def test_heartbeat_returns_200(
-    client, db, phabdouble, request_mocker, redis_cache, jwks, treestatusdouble
```
zzzeid, line 26: `db` here was causing a strange app context error during teardown. Going to dig a little as to whether this is going to cause any real problems or whether we should return the db fixture here, but the error was happening on teardown.

tests/test_dockerflow.py
```diff
@@ -4,7 +4,7 @@
 
 import json
 
-from tests.utils import phab_url, trans_url
+from tests.utils import phab_url
```
cgsheeh, line 7: Can we file a bug for removing `trans_url` and other transplant related test code?

zzzeid: Sure, most of the tansplant related test code was removed in [bug 1654646](https://bugzilla.mozilla.org/show_bug.cgi?id=1654646) but there are some stragglers left like this one. Filed [bug 1831622](https://bugzilla.mozilla.org/show_bug.cgi?id=1831622).

PR #308, opened 2023-05-09
landing_worker: release lock when queue is empty (bug 1832083)

zzzeid:


landoapi/workers/landing_worker.py
```diff
@@ -104,6 +104,7 @@ def loop(self):
 
         if job is None:
             self.throttle(self.sleep_seconds)
+            db.session.commit()
```
cgsheeh, line 107: Nit: should we commit and then throttle? Seems like we're just holding the lock for `sleep_seconds` longer than we need to?

cgsheeh (approved):

landoapi/workers/landing_worker.py
zzzeid, line 107:
```suggestion
            db.session.commit()
            self.throttle(self.sleep_seconds)
```

PR #309, opened 2023-05-09
app: remove legacy config keys (bug 1832137)

zzzeid:


cgsheeh (changes requested):
Looks good! 

We should also remove `LOCALDEV_MOCK_TRANSPLANT_SUBMIT` from `docker-compose.yml`, remove `trans_url` from `tests/utils.py`, and rename `TRANSPLANT_FAILURE_EMAIL_TEMPLATE` in `landoapi/email.py`.

PR #310, opened 2023-05-09
cache: add DISABLE_CACHE, support filesystem caching (bug 1832142)

zzzeid:
- add explicit flag for initializing cache in null mode
- if cache is enabled and no host is provided, use filesystem


cgsheeh (approved):
Explicit > Implicit. :)

What do we need the filesystem mode for?

PR #313, opened 2023-05-17
tests: remove some `noqa` where strings could be moved to multi-line (Bug 1833710)

cgsheeh:


zzzeid (changes requested):
I think it is nicer to put the spaces before the line breaks instead of after (i.e., not to start lines with a space). This is also how we have it already in other parts of the code (e.g., landing_worker.py) though not sure if it's like that everywhere. This might be a good time to establish a convention.

E.g.,

```python
                error_info = (
                    'Error while reading "ids": Expected a nonempty list, '
                    "but value is an empty list."
                )
```

Instead of

```python
                error_info = (
                    'Error while reading "ids": Expected a nonempty list,'
                    " but value is an empty list."
                )
```

dklawren (approved):

zzzeid (approved):

PR #317, opened 2023-05-25
transplants: update revision_to_diff_id with legacy data (bug 1834868)

zzzeid:


cgsheeh (approved):
Makes sense to me. Could we have a test for this?

PR #320, opened 2023-05-29
stacks: add revision data to API response (bug 1835678)

zzzeid:
- add Lando revision data to API response
- add serialize method to Revision class


landoapi/models/revisions.py
```diff
@@ -86,6 +86,16 @@ def set_patch(self, raw_diff: bytes, patch_data: dict[str, str]):
         patch = build_patch_for_revision(raw_diff, **self.patch_data)
         self.patch_bytes = patch.encode("utf-8")
 
+    def serialize(self):
```
cgsheeh, line 89:
```suggestion
    def serialize(self) -> dict[str, Any]:
```

cgsheeh (approved):

PR #324, opened 2023-06-02
mocks: add bugzilla.account.search mock endpoint (bug 1836484)

zzzeid:


dklawren (approved):
Looks sane to me. r=dkl

tests/mocks.py
```diff
@@ -733,6 +733,27 @@ def comment(self, content, author=None):
     def conduit_ping(self):
         return "ip-123-123-123-123.us-west-2.compute.internal"
 
+    @conduit_method("bugzilla.account.search")
+    def bugzilla_account_search(self, phids=None, ids=None):
+        """Return a list of Bugzilla IDs and Phabricator phids, given phids or ids."""
+        if ids:
+            raise NotImplementedError("Searching by Bugzilla ID is not implemented.")
```
cgsheeh, line 740: Not implemented in the test mock, or not implemented in the Conduit API endpoint?

cgsheeh (approved):
This looks good to me, but it's a bit hard to tell since this is unused code at the moment. I think it makes more sense to add this in the patch where it will be used in a test.

tests/mocks.py
zzzeid: > Not implemented in the test mock, or not implemented in the Conduit API endpoint?

In the test mock -- I can be more clear about that.

zzzeid:
> This looks good to me, but it's a bit hard to tell since this is unused code at the moment. I think it makes more sense to add this in the patch where it will be used in a test.

This is used in https://github.com/mozilla-conduit/lando-api/pull/325, which I can land around the same time (as it depends on this one).

PR #325, opened 2023-06-02
revisions: record approvers, peers, and owners at landing (bug 1836485)

zzzeid:
- add mots to requirements
- add Revision.data field that will store meta information
- upon landing request, store list of reviewers who accepted revision
- upon landing, store list of peers and owners who accepted revision
- add tests

The new Revision.data field will be used to store information about who approved a revision and of those users, who is a peer or an owner. The data will be used for analysis down the road to determine how many revisions land with peer or owner approval.


landoapi/workers/landing_worker.py
```diff
@@ -414,6 +415,10 @@ def run_job(
             LandingJobAction.LAND, commit_id=commit_id, commit=True, db=db
         )
 
+        mots_path = Path(hgrepo.path) / "mots.yaml"
+        if mots_path.exists():
+            job.set_landed_reviewers(mots_path)
```
zzzeid, line 420: May need a catch-all exception here, in case someone breaks `mots.yaml` since we don't have an automated check that prevents someone from landing a corrupted version of mots.yaml at this time.

landoapi/models/revisions.py
```diff
@@ -69,6 +69,9 @@ class Revision(Base):
         "LandingJob", secondary=revision_landing_job, back_populates="revisions"
     )
 
+    # A general purpose data field to store arbitrary information about this revision.
```
cgsheeh, line 72: What's the advantage to using a dynamic JSON object here instead of adding proper columns?

landoapi/reviews.py
```diff
@@ -212,3 +212,17 @@ def approvals_for_commit_message(
     ]
 
     return accepted_reviewers, approvals
+
+
+def get_approved_by_ids(phab: PhabricatorClient, reviewers: dict) -> list[str]:
```
cgsheeh, line 217: The type hint for `reviewers` is wrong, looks like it should be `list[dict]`.

cgsheeh (changes requested):
Please add some tests and a commit message.

landoapi/reviews.py
zzzeid, line 217:
```suggestion
def get_approved_by_ids(phab: PhabricatorClient, reviewers: list[dict]) -> list[str]:
```

landoapi/models/revisions.py
zzzeid: Discussed this over zoom.

landoapi/models/landing_job.py
```diff
@@ -272,6 +275,20 @@ def set_landed_revision_diffs(self):
...
+        directory = Directory(FileConfig(path))
+        for revision in self.revisions:
+            approved_by = revision.data.get("approved_by")
+            if not approved_by:
+                continue
+            data = {"peers_or_owners": []}
+            for reviewer in approved_by:
+                if reviewer in directory.peers_and_owners:
+                    data["peers_or_owners"].append(reviewer)
+            data.update(revision.data)
```
cgsheeh, line 289: If `revision.data` already included a `peers_or_owners` field, wouldn't this overwrite it? Maybe we want to `copy.deepcopy(revision.data)` and then add the `peers_or_owners` key instead?

cgsheeh (changes requested):
There's still nothing in the bug or the commit message describing the use case for this new field. Please add some information about what this is going to be used for, and include the details of our Zoom discussion so we have a record of what we talked about.

landoapi/models/landing_job.py
zzzeid: Good point, this could happen for revisions that land more than once. I've reworked the way this is implemented to avoid having to create a new dictionary (which was one way to trigger a "modified" state on the JSONB field.).

cgsheeh (approved):

landoapi/models/landing_job.py
```diff
@@ -282,12 +283,17 @@ def set_landed_reviewers(self, path: Path):
...
+
+            if "peers_and_owners" not in revision.data:
+                revision.data["peers_and_owners"] = []
+
             for reviewer in approved_by:
                 if reviewer in directory.peers_and_owners:
-                    data["peers_or_owners"].append(reviewer)
-            data.update(revision.data)
-            revision.data = data
+                    revision.data["peers_and_owners"].append(reviewer)
```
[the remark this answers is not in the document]
zzzeid: True. Fixed, though this field is still not "perfect" since if there is a backout and a revision is landed again with a different approving reviewer than the original, both will be included in this field. I think that's OK for the scope/purpose of this field though.

PR #326, opened 2023-06-02
landing_job: iterate over revisions directly when applying patches (Bug 1836499)

cgsheeh:
The `all_patches` list is built in the `LandingWorker` as
the ordered set of patches to apply to the repo for a job.
The contents of this list is built entirely from the `revisions`
field, so iterate over them directly instead.

Since we no longer have a list of patches to check the length
of for autoformatting, use the length of the changeset titles
list instead, as it is a better variable to use in that context
anyways.

### DIFF
--- a/landoapi/models/landing_job.py
+++ b/landoapi/models/landing_job.py
@@ -5,11 +5,9 @@ import datetime
 import enum
 import logging
 import os
-from io import BytesIO
 from typing import (
     Any,
     Iterable,
-    Iterator,
     Optional,
 )
 
@@ -331,17 +329,6 @@ class LandingJob(Base):
 
         if commit:
             db.session.commit()
-
-    def get_patches(self) -> Iterator[tuple[int, BytesIO]]:
-        """Generator of patches for this landing job.
-
-        Yields patches in the format (patch identifier, patch buffer) for application
-        to the repository in the landing worker.
-        """
-        return (
-            (revision.revision_id, BytesIO(revision.patch_bytes))
-            for revision in self.revisions
-        )
 
     def serialize(self) -> dict[str, Any]:
         """Return a JSON compatible dictionary."""
--- a/landoapi/models/landing_job.py
+++ b/landoapi/models/landing_job.py
@@ -5,9 +5,11 @@ import datetime
 import enum
 import logging
 import os
+from io import BytesIO
 from typing import (
     Any,
     Iterable,
+    Iterator,
     Optional,
 )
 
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -426,3 +427,35 @@ def strptime(cls, date_string, fmt):
...
++++ b/test.txt
+@@ -1,1 +1,2 @@
+ TEST
++adding another line
+""".strip()
+
+
+@pytest.fixture
+def create_patch_revision(db):
+    """A fixture that fake uploads a patch"""
```
zzzeid, line 450: While we're here, this docstring should be updated (likely still refers to S3 uploads from before the migration).

landoapi/models/landing_job.py
```diff
@@ -330,6 +332,17 @@ def transition_status(
...
             db.session.commit()
 
+    def get_patches(self) -> Iterator[tuple[int, BytesIO]]:
+        """Generator of patches for this landing job.
+
+        Yields patches in the format (patch identifier, patch buffer) for application
+        to the repository in the landing worker.
+        """
+        return (
+            (revision.revision_id, BytesIO(revision.patch_bytes))
```
zzzeid, line 342: I assume that the intention here is to return a different value for try pushes in a future PR? Something like

```python
return (
    (revision.revision_id, revision.id, BytesIO(revision.patch_bytes))
...
```

However, see other comment regarding the need for this at all.

landoapi/workers/landing_worker.py
```diff
@@ -291,26 +290,18 @@ def run_job(
...
 
-            # Fetch all patches.
-            all_patches = [
-                (revision.revision_id, BytesIO(revision.patch_bytes))
-                for revision in job.revisions
-            ]
-
             # Run through the patches one by one and try to apply them.
-            for revision_id, patch_buf in all_patches:
+            for patch_id, patch_buf in job.get_patches():
```
zzzeid, line 294: Upon rethinking this a little more, I think it would make more sense to access the revisions directly here.
```suggestion
            for revision in job.revisions:
                hgrepo.apply_patch(BytesIO(revision.patch_bytes))
```

Revisions are [lazily-loaded](https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#lazy-loading) anyway.

And then down the road, access the necessary variables directly, e.g.,

```python
breakdown = self.process_merge_conflict(exc, repo, hgrepo, revision.revision_id)
```

I don't think this breakdown is necessary any more, and is just a small remnant of the old implementation.

zzzeid (changes requested):

cgsheeh:
Changed the loop to just iterate over `job.revisions` directly. Left the moving of the patch names and test fixture in for now, I can remove them if you want but it's more effort than it's worth IMO.

zzzeid (changes requested):
>Since we no longer have a list of patches to check the length
of for autoformatting, use the length of the changeset titles
list instead, as it is a more accurate number in that
context anyways.

How is it a more accurate number? Shouldn't it be the exact same number? Perhaps "a better variable to use in that context" would make more sense -- this request is just to edit the commit message.

>Left the moving of the patch names and test fixture in for now, I can remove them if you want but it's more effort than it's worth IMO.

```shell
git checkout main -- tests/conftest.py tests/test_landing_job.py tests/test_landings.py
git stash
```

I think it would be better to put them in their own PR if you still want to make this change. No harm in small, no-bug PRs that make these unrelated changes (e.g., #323)

zzzeid (approved):
lgtm!

PR #328, opened 2023-06-09
api: add an API endpoint for submission of try pushes (Bug 1835959)

cgsheeh:
Adds a new `POST /try` endpoint to Lando that accepts try patch
data and creates a landing job to submit the patches to try.
Patches are submitted as an array of base64 encoded bytes and
the base commit for the try submission in `mozilla-unified` is
passed so the patch applies as it would in the developer's local
machine.

The `Repo` object sees several changes to allow Lando to work
with the try repo. A `force_push` option is added that controls
adding the `-f` option during `hg push`. A `phabricator_repo`
flag is also added which indicates if the repo is present on
Phabricator. At the moment this flag controls pinging Phabricator
to issue a repo update API call, which is irrelevant for try.

The try repo is added to the repo configs, with `mozilla-unified`
as the `pull_url` and `try` as the `push_url`. This means patches
will be applied to `mozilla-unified` in Lando and pushed out to
`try`, in the same fashion as the current try push model on a
developer's local machine.

A decorator is added to enforce `active_scm_level_1` group
permissions with appropriate error messages.

Patches can be submitted in the `hg export` or `git format-patch`
formats. Support for `git format-patch` is added via a Git patch parser
similar to the `HgPatchHelper`. To parse Git date headers, the `python-dateutil`
package is added as a dependency.


landoapi/api/try.py
```diff
@@ -0,0 +1,77 @@
...
+            500,
+            "Could not find a `try` repo to submit to.",
+            "Could not find a `try` repo to submit to.",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500",
+        )
+
+    # Add a landing job for this try push.
+    ldap_username = g.auth0_user.email
+    revisions = [
+        # TODO do something more useful with `patch_data`, maybe gather data from
```
cgsheeh, line 59: I may add the patch data information to the request from the client, or parse it from the patch data here on the server-side. Still on the fence if it's worth adding here.

landoapi/api/try.py
```diff
@@ -0,0 +1,77 @@
...
+        # rs-parsepatch??
+        Revision(patch_bytes=base64.b64decode(patch.encode("ascii")), patch_data={})
```
zzzeid, line 61: This should ideally use `Revision.set_patch` instead, which already encodes `patch` (perhaps you can pass the encoding to that method if it should be a different one) and it uses `patch_data` when calling `build_patch_for_revision`. You can pass boilerplate `patch_data`if needed, but it is better to have this go through the conventional process of patch setting rather than directly setting the patch here.

Another reason to do that apart from the above, and reducing repetition is that in the future, there may be some restrictions as to when and how `patch_bytes` can be set -- for example, if a request to land is issued, there should be no way to modify `patch_bytes`. It would be good to keep this convention and use `set_patch` just in case any other checks or restrictions are needed in the future.

landoapi/auth.py
```diff
@@ -493,6 +493,40 @@ def __call__(self, f: Callable) -> Callable:
...
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404",
+        )
+
+
+def enforce_scm_level_1(func):
+    """Decorator to enforce `active_scm_level_1` membership with error messaging."""
+
+    @functools.wraps(func)
+    def wrap_api(*args, **kwargs):
+        assert_scm_level_1(g.auth0_user)
```
zzzeid, line 524:
```suggestion
        enforce_scm_level(g.auth0_user, SCM_LEVEL_1)
```

landoapi/hg.py
```diff
@@ -556,19 +556,26 @@ def push(self, target, bookmark=None):
             and self.patch_header("Fail HG Import") == b"LOSE_PUSH_RACE"
         ):
             raise LostPushRace()
+
+        force_args = ["-f"] if force_push else []
```
zzzeid, line 560:
```suggestion
        extra_args = ["-f"] if force_push else []
```

landoapi/hg.py
```diff
@@ -556,19 +556,26 @@ def push(self, target, bookmark=None):
...
+
         try:
             if bookmark is None:
-                self.run_hg(["push", "-r", "tip", target])
+                self.run_hg(["push", "-r", "tip", target] + force_args)
```
zzzeid, line 564:
```suggestion
                self.run_hg(["push", "-r", "tip", target] + extra_args)
```

landoapi/hg.py
```diff
@@ -556,19 +556,26 @@ def push(self, target, bookmark=None):
...
             else:
                 self.run_hg_cmds(
-                    [["bookmark", bookmark], ["push", "-B", bookmark, target]]
+                    [
+                        ["bookmark", bookmark],
+                        ["push", "-B", bookmark, target] + force_args,
```
zzzeid, line 569:
```suggestion
                        ["push", "-B", bookmark, target] + extra_args,
```

landoapi/repos.py
```diff
@@ -266,6 +296,18 @@ def phab_identifier(self) -> str:
             "/1.0/firefox_versions.json",
             autoformat_enabled=True,
         ),
+        # Try uses `mozilla-unified` as the `pull_path` as using try
```
zzzeid, line 299: Why did you choose `mozilla-unified` here? (vs say `autoland`, or `mozilla-central`?) Maybe just add a bit about that in the comment.

zzzeid (changes requested):
Looks good with first pass! Just a few small requests here and there.

landoapi/api/try.py
cgsheeh: My plan was to have the client simply `hg export` the patches so they could be applied directly. `Revision.set_patch` builds an `hg export` formatted patch from a diff and metadata, which might end up being unnecessary and slower to parse from patches on the client. I'm still tinkering with the best way to get patch data into the request.

migrations/versions/06cc1c6a04b5_set_phab_related_revision_fields_to_.py
zzzeid, line 1: I suggest renaming this file to 06cc1c6a04b5_try_related_changes.py or something more succinct as the current name of the file is cut off and you can't really tell what it is.

landoapi/api/try.py
zzzeid: > My plan was to have the client simply `hg export` the patches so they could be applied directly. `Revision.set_patch` builds an `hg export` formatted patch from a diff and metadata, which might end up being unnecessary and slower to parse from patches on the client. I'm still tinkering with the best way to get patch data into the request.

If you want to apply them as-is, then I would still go through `Revision.set_patch` but modify it so that it can take a full patch and circumvent the need for extra patch data. That way we would still be able to apply any future changes to when/how patches can be modified (e.g., a status check or a hash check). In the future we may want to do a pre-update hook instead but that is out of scope for now.

zzzeid: One other thing to keep in mind is the patch format expectations that the landing worker has (e.g., `Diff Start Line` header, etc.) which may not be as expected if using the diff as-is.

landoapi/api/try_push.py
```diff
@@ -0,0 +1,222 @@
...
+from landoapi.models.revisions import Revision
+from landoapi.repos import (
+    SCM_LEVEL_1,
+    get_repos_for_env,
+)
+
+logger = logging.getLogger(__name__)
+
+
+# Borrowed from Mercurial core.
```
zzzeid, line 35: Can this not be imported directly from the `mercurial` module?

```python
from mercurial.utils import stringutil
```

Same applies for anything else copy/pasted.

landoapi/api/try_push.py
```diff
@@ -0,0 +1,222 @@
...
+
+    return Revision.new_from_patch(
+        raw_diff=helper.get_diff().decode("utf-8"),
+        patch_data={
+            "author_name": author.decode("utf-8"),
+            "author_email": email.decode("utf-8"),
+            "commit_message": helper.commit_description().decode("utf-8"),
+            "timestamp": timestamp,
+        },
+    )
```
zzzeid, line 151: These two methods are very similar, and can be combined into one method that takes a `helper` argument, and with minor additional changes to accommodate the differences in timestamp/user headers. Just for illustration, e.g., 

```python
def build_revision_from_patch(helper: Helper, patch: bytes) -> Revision:
    """Convert a formatted `bytes` patch into a `Revision`."""
    helper = helper(io.BytesIO(patch))

    user = helper.header(user_header_name)
    if not user:
        raise ValueError("Patch does not have a `User` header.")

    author, email = parse_git_author_information(user)

    date = helper.header(date_header_name)
    if not date:
        raise ValueError("Patch does not have a `Date` header.")

    timestamp = get_timestamp_from_hg_date_header(date)

    commit_message = helper.commit_description()
    if not commit_message:
        raise ValueError("Patch does not have a commit description.")

    return Revision.new_from_patch(
        raw_diff=helper.get_diff().decode("utf-8"),
        patch_data={
            "author_name": author.decode("utf-8"),
            "author_email": email.decode("utf-8"),
            "commit_message": helper.commit_description().decode("utf-8"),
            "timestamp": timestamp.decode("utf-8"),
        },
    )
```

landoapi/auth.py
```diff
@@ -493,6 +494,43 @@ def __call__(self, f: Callable) -> Callable:
         return self._require_access_token(f)
 
 
+def ensure_user_has_scm_level(auth0_user: A0User, access_group: AccessGroup):
+    """Raise an appropriate `ProblemException` if the user is missing `scm_level_1`."""
+    # Return appropriate error message if user does not have commit access.
+    if not auth0_user.is_in_groups(access_group.membership_group):
+        raise ProblemException(
+            401,
```
zzzeid, line 502: This should probably be  a `403`, if a user has provided a valid token, but the token does not give them the correct permissions.

landoapi/hgexports.py
```diff
@@ -71,7 +71,7 @@ def _no_line_breaks(break_string: str) -> str:
     return "".join(break_string.strip().splitlines())
 
 
-class PatchHelper(object):
+class HgPatchHelper(object):
```
zzzeid, line 74: There is repeated identical code between this class and `GitPatchHelper` (e.g., `_is_diff_line`, `header`, `write`, etc.). Common code should be moved to an abstract `Helper` and that class should be extended by `GitPatchHelper` and `HgPatchHelper`.

landoapi/hgexports.py
```diff
@@ -163,6 +163,129 @@ def write_commit_description(self, f: io.BytesIO):
...
+
+    def write_diff(self, file_obj: io.BytesIO):
+        """Writes the diff to the specified file object."""
+        file_obj.write(self.get_diff())
+
+
+class GitPatchHelper(object):
+    """Helper class for parsing Mercurial patches/exports."""
+
+    def __init__(self, fileobj: io.BytesIO):
```
zzzeid, line 201: `__init__` is a little too large -- it could be split into different instance methods based on functionality. It would be better if it was easy to figure out what `__init__` is doing by glancing at a few calls to helper methods instead.

landoapi/hgexports.py
```diff
@@ -163,6 +163,129 @@ def write_commit_description(self, f: io.BytesIO):
...
+        else:
+            raise ValueError("Patch is malformed, could not find start of patch diff.")
+
+        # The diff is the remainder of the patch, except the last two lines of Git version info.
+        remaining_lines = list(self.patch)
+        self.diff += b"".join(line for line in remaining_lines[:-2])
+
+    @staticmethod
+    def _is_diff_line(line: bytes) -> bool:
+        return DIFF_LINE_RE.search(line) is not None
```
zzzeid, line 260: This is repeated code.

landoapi/hgexports.py
```diff
@@ -163,6 +163,129 @@ def write_commit_description(self, f: io.BytesIO):
...
+    def write(self, f: io.BytesIO):
+        """Writes whole patch to the specified file object."""
+        try:
+            while 1:
+                buf = self.patch.read(16 * 1024)
+                if not buf:
+                    break
+                f.write(buf)
+        finally:
+            self.patch.seek(0)
```
zzzeid, line 283: This is repeated code.

landoapi/hgexports.py
```diff
@@ -163,6 +163,129 @@ def write_commit_description(self, f: io.BytesIO):
...
+
+    def write_commit_description(self, f: io.BytesIO):
+        """Writes the commit description to the specified file object."""
+        f.write(self.commit_description())
```
zzzeid, line 287: This is repeated code.

landoapi/models/landing_job.py
```diff
@@ -166,9 +169,26 @@ def serialized_landing_path(self):
...
-        return f"D{self.revisions[-1].revision_id}"
+        if not self.revisions:
+            raise ValueError(
+                "Job must be associated with a revision to have a head revision."
+            )
+
+        head = self.revisions[-1]
+        if head.revision_id:
+            # Return the Phabricator identifier if the head revision has one.
+            return f"D{self.revisions[-1].revision_id}"
```
zzzeid, line 182:
```suggestion
            return f"D{head.revision_id}"
```

landoapi/models/landing_job.py
```diff
@@ -166,9 +169,26 @@ def serialized_landing_path(self):
...
+
+        commit_message = head.patch_data.get("commit_message")
+        if commit_message:
+            # If there is no Phabricator identifier, return the first line of the
+            # commit message for the patch.
```
zzzeid, line 187: Probably more useful to include the commit hash here? Or possibly both.

landoapi/models/landing_job.py
```diff
@@ -166,9 +169,26 @@ def serialized_landing_path(self):
...
+            return commit_message.splitlines()[0]
+
+        # Return a placeholder in the event neither exists.
+        return "unknown"
```
zzzeid, line 191: There should always be at least a commit hash, so we should return that in case nothing else is available (though that can be covered by the previous statement).

landoapi/repos.py
```diff
@@ -74,6 +80,8 @@ class Repo:
     autoformat_enabled: bool = False
     commit_flags: list[tuple[str, str]] = field(default_factory=list)
     product_details_url: str = ""
+    phabricator_repo: bool = True
```
zzzeid, line 83:
```suggestion
    is_phabricator_repo: bool = True
```

landoapi/repos.py
```diff
@@ -91,53 +99,65 @@ def __post_init__(self):
             self.short_name = self.tree
 
     @property
-    def phab_identifier(self) -> str:
+    def phab_identifier(self) -> str | None:
```
zzzeid, line 102: Out of scope for this PR, but I think the way we lay out the repos is not great, since there is a duplication of values in the keys of the dictionary of repos and `short_name`, in each entry. In the future it might be good to refactor this so that there is no duplication.

zzzeid (changes requested):
First pass. Will review tests on second pass.

landoapi/models/landing_job.py
cgsheeh: The commit hash will change after landing, so the commit title would be a more useful identifier. This property is used as the identifying string in the landing failure email.

That said it just occurred to me that the head commit won't be a useful string either since most commit messages will just be something related to the try selector used, like [this push](https://hg.mozilla.org/try/rev/443c2cbbd878df752fc6833794a15e4126154272) for example. I'll update this to use the `-2` commit instead.

landoapi/hgexports.py
```diff
@@ -163,6 +163,129 @@ def write_commit_description(self, f: io.BytesIO):
...
+                    if line_no == self.diff_start_line:
+                        diff.append(line)
+                        break
+                else:
+                    if self._is_diff_line(line):
+                        diff.append(line)
+                        break
+
+            while 1:
+                buf = self.patch.read(16 * 1024)
```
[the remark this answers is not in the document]
cgsheeh: This is just an issue with the diff display, this code is the already-existing code in `PatchHelper`. I can address this in a separate patch.

landoapi/hgexports.py
```diff
@@ -163,6 +163,129 @@ def write_commit_description(self, f: io.BytesIO):
...
+                if not buf:
+                    break
+                diff.append(buf)
+
+            return b"".join(diff)
+        finally:
+            self.patch.seek(0)
```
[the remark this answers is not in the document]
cgsheeh: This is just an issue with the diff display, this code is the already-existing code in `PatchHelper`. I can address this in a separate patch.

landoapi/api/try_push.py
cgsheeh: When you sent this it occurred to me that I can't use this code due to incompatible licenses. I found a different solution that uses functions in the standard library.

landoapi/auth.py
```diff
@@ -499,21 +499,21 @@ def ensure_user_has_scm_level(auth0_user: A0User, access_group: AccessGroup):
...
     if auth0_user.is_in_groups(
         access_group.expired_group
     ) or not auth0_user.is_in_groups(access_group.active_group):
         raise ProblemException(
-            401,
+            403,
             f"Your {access_group.display_name} has expired.",
             f"Your {access_group.display_name} has expired.",
-            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403",
```
zzzeid: This should have remained `401` (since the token is provided, but it has expired permissions and thus is invalid).

landoapi/hgexports.py
```diff
@@ -71,12 +75,92 @@ def _no_line_breaks(break_string: str) -> str:
...
+    def write(self, f: io.BytesIO):
+        """Writes whole patch to the specified file object."""
+        try:
+            while 1:
+                buf = self.patch.read(16 * 1024)
+                if not buf:
+                    break
+                f.write(buf)
+        finally:
+            self.patch.seek(0)
```
zzzeid, line 148: I've made a [previous comment](https://github.com/mozilla-conduit/lando-api/pull/328#discussion_r1270973325) about a similar implementation/piece of code, there are some improvements there that could be applied here.

landoapi/models/landing_job.py
```diff
@@ -179,12 +183,15 @@ def landing_job_identifier(self) -> str:
         head = self.revisions[-1]
         if head.revision_id:
             # Return the Phabricator identifier if the head revision has one.
-            return f"D{self.revisions[-1].revision_id}"
+            return f"D{head.revision_id}"
+
+        # If there is no Phabricator identifier, return the first line of the
+        # non-try-syntax commit's message for the patch.
+        if len(self.revisions) > 1:
+            head = self.revisions[-2]
```
zzzeid, line 191: Is there a more rigorous way of detecting a try syntax commit?

tests/test_auth.py
```diff
@@ -374,7 +374,7 @@ def test_scm_level_enforce():
     user = A0User(token, userinfo)
     with pytest.raises(ProblemException) as exc_info:
         ensure_user_has_scm_level(user, SCM_LEVEL_1)
-    assert exc_info.value.status == 401, "Expired level 1 permission should return 401."
+    assert exc_info.value.status == 403, "Expired level 1 permission should return 403."
```
zzzeid: This should have remained as `401`.

landoapi/auth.py
```diff
@@ -493,6 +494,43 @@ def __call__(self, f: Callable) -> Callable:
         return self._require_access_token(f)
 
 
+def ensure_user_has_scm_level(auth0_user: A0User, access_group: AccessGroup):
+    """Raise an appropriate `ProblemException` if the user is missing `scm_level_1`."""
+    # Return appropriate error message if user does not have commit access.
```
zzzeid, line 499: This method does not return anything.

```suggestion
```

landoapi/hg.py
```diff
@@ -556,19 +556,26 @@ def push(self, target, bookmark=None):
             and self.patch_header("Fail HG Import") == b"LOSE_PUSH_RACE"
         ):
             raise LostPushRace()
+
+        extra_args = ["-f"] if force_push else []
```
zzzeid, line 560: nit: This would be a little more future proof if `extra_args` was initialized as an empty list, and appended to as needed.

landoapi/hgexports.py
```diff
@@ -71,12 +75,92 @@ def _no_line_breaks(break_string: str) -> str:
...
+def get_timestamp_from_hg_date_header(date_header: bytes) -> bytes:
+    """Return the first part of the `hg export` date header.
+
+    >>> get_timestamp_from_hg_date_header(b"1686621879 14400")
+    b"1686621879"
+    """
+    return date_header.split(b" ")[0]
+
+
 class PatchHelper(object):
```
zzzeid, line 102: While we're here.

```suggestion
class PatchHelper:
```

landoapi/hgexports.py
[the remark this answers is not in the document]
zzzeid: Noted, RE: `16 * 1024`. However, the other comment still applies (about the large try/except and the while loop). I think since this code is being modified, it makes sense to clean it up while you're here. Parts of it are also copied to the base class definition of `write`, so at the very least that definition could benefit from a refactoring.

landoapi/hgexports.py
```diff
@@ -71,12 +75,92 @@ def _no_line_breaks(break_string: str) -> str:
...
 
     def __init__(self, fileobj: io.BytesIO):
         self.patch = fileobj
         self.headers = {}
+
+    @staticmethod
+    def _is_diff_line(line: bytes) -> bool:
+        return DIFF_LINE_RE.search(line) is not None
+
+    def header(self, name: bytes | str) -> Optional[bytes]:
```
zzzeid, line 113: Might be better to rename this to `get_header` to make things a little more consistent.

landoapi/hgexports.py
```diff
@@ -148,41 +223,136 @@ def commit_description(self) -> bytes:
...
+
+            return b"".join(diff)
         finally:
             self.patch.seek(0)
+
+    def parse_author_information(self) -> tuple[bytes, bytes]:
+        """Return the author name and email from the patch."""
+        user = self.header("User")
+        if not user:
+            raise ValueError("Patch does not have a `User` header.")
```
zzzeid, line 257: More accurately stated as
```suggestion
            raise ValueError("Could not determine patch author information from header.")
```

Though if you want you could raise `ValueError` in `get_header`, and keep this same message.

landoapi/hgexports.py
```diff
@@ -148,41 +223,136 @@ def commit_description(self) -> bytes:
...
+
+        return parse_git_author_information(user)
+
+    def get_timestamp(self) -> bytes:
+        """Return an `hg export` formatted timestamp."""
+        date = self.header("Date")
+        if not date:
+            raise ValueError("Patch does not have a `Date` header.")
```
zzzeid, line 265: Similar comment as `parse_author_information`.

landoapi/hgexports.py
zzzeid: The intention of my comment was to encourage breaking down the functionality into separate methods, not moving it all to a single method -- since that does not alleviate the fact that there are a few too many things happening in `parse_patch` that could be broken down into different methods, and the readability (and code flow) should be improved at that point.

zzzeid (changes requested):
Looking good! Few more comments on this pass related to your most recent changes.

landoapi/hgexports.py
cgsheeh: I'll handle this change in [bug 1846330](https://bugzilla.mozilla.org/show_bug.cgi?id=1846330), it's out of scope for this PR. This code is pre-existing in Lando and is just being moved to the new base `PatchHelper`.

landoapi/models/landing_job.py
cgsheeh: There are quite a few different incantations of try commit syntax, so it isn't quite that simple. For the MVP I think this is a simple enough solution, and in the future we could add a check for specific types of try syntax. The key thing here is to just have some string that helps developers identify which patch this landing job is associated with.

landoapi/hgexports.py
cgsheeh: As discussed during our team meeting, while updating this patch to address these comments I realized my existing solution of looking for specific headers in a specific order was not robust enough, as the user's underlying config files may add more headers to the produced patch than I was seeing in local testing.

To fix this I have changed the initial parsing of the patch to use the `email` package, as `git format-patch` patches are formatted in Unix mbox format. They are essentially email messages. So the updated parsing uses the `EmailMessage` object to get the values of the appropriate headers, and we parse the email body content for extended commit description and the contents of the diff.

I updated the `PatchHelper` class and subclasses to use `str` instead of `bytes` everywhere. I also attempted to fix this inconsistency in other parts of the code (ie make places like `HgRepo.apply_patch` no longer take an `io.BytesIO`) but the problem is a bit deeper than that and I decided it is out of scope. Despite using `str` in most places in the code, the patch itself is `LargeBinary` in the database, so we are still expecting `str` in some areas and `bytes` in others. I will file a bug to track the work of fleshing that out and being more consistent with the types we use to represent patches in the future.

landoapi/hgexports.py
```diff
@@ -274,85 +287,83 @@ def __init__(self, fileobj: io.BytesIO):
...
-
-        author_from_line = next(self.patch)
-        if not author_from_line.startswith(b"From: "):
-            raise ValueError("Patch is malformed, second line is not `From:`.")
-        self.set_header(
-            "From", author_from_line.removeprefix(b"From: ").removesuffix(b"\n")
-        )
+    def get_header(self, name: bytes | str) -> Optional[str]:
+        """Get the headers from the message."""
+        name = name.decode("utf-8") if isinstance(name, bytes) else name
```
zzzeid, line 292: Slightly less redundant:
```suggestion
        if isinstance(name, bytes):
            name = name.decode("utf-8")
```

landoapi/hgexports.py
```diff
@@ -274,85 +287,83 @@ def __init__(self, fileobj: io.BytesIO):
...
-        # Clean up the first subject line.
-        subject_line = (
-            subject_line.removeprefix(b"Subject: ")
-            .removeprefix(b"[PATCH] ")
-            .removesuffix(b"\n")
-        )
+        # `EmailMessage` will return `None` if the header isn't found.
+        return self.message[name]
+
+    def verify_content(self, content: str) -> str:
```
zzzeid, line 297: It would be good if this was renamed -- since it is doing a more than verifying (i.e., setting `self.commit_message` and returning `diff_lines`.) Maybe something like `verify_and_set_diff_content` or something with a verb implying things are being modified.

Same comment applies to the docstring.

landoapi/hgexports.py
```diff
@@ -135,54 +217,153 @@ def commit_description(self) -> bytes:
...
+        )
+        self.diff = self.verify_content(self.message.get_content())
+
+    def commit_description(self) -> str:
+        """Returns the commit description."""
+        return self.commit_message
+
+    def get_diff(self) -> str:
+        """Return the patch diff."""
+        return self.diff
```
zzzeid, line 353: What is the point of these methods? If it's to make naming more consistent, perhaps `commit_description` should be `get_commit_description`.

zzzeid (changes requested):
Looking good! Just a few more minor comments.

landoapi/hgexports.py
cgsheeh: `commit_description` is carried over from the existing `PatchHelper` code.

`get_diff` has been added to retrieve the diff from the patch since we need it for adding to the landing job. The existing `PatchHelper` code had a `write_diff` method which I refactored to instead return the diff, and made `write_diff` just write the value returned from `get_diff`.

landoapi/spec/swagger.yml
```diff
@@ -366,6 +366,45 @@ paths:
           schema:
             allOf:
               - $ref: '#/definitions/Error'
+  /try/patches:
+    post:
+      operationId: landoapi.api.try_push.post
```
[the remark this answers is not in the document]
cgsheeh: The `patches` nomenclature is supposed to represent the format of the push content, that we are sending the commits to be pushed as patches in the given VCS format. In the future if we wanted to submit bundles or some other format we would update the URL. I'll update the handler function names to reflect this.

landoapi/hgexports.py
cgsheeh: This is just carried over from the existing `PatchHelper` implementation. I'll update it.

tests/test_try.py
```diff
@@ -0,0 +1,209 @@
...
+        "id" in response.json
+    ), "Response should include the ID of the new landing job."
+
+    queue_items = LandingJob.job_queue_query(
+        repositories=["try"], grace_seconds=0
+    ).all()
+    assert len(queue_items) == 1, "Try push should have created 1 landing job."
+
+    # Run the landing job.
+    job = queue_items[0]
```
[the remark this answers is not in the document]
cgsheeh: Can you be more specific about what you think should be checked?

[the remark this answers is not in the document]
zzzeid: Checking `job.revisions` length, that revisions were created as expected (this would be coverage for `Revision.new_from_patch`), etc... This would also help people reading this integration test understand the code flow.

landoapi/api/try_push.py
```diff
@@ -53,38 +63,52 @@ def build_revision_from_patch_helper(helper: PatchHelper) -> Revision:
...
+            return [
+                build_revision_from_patch_helper(HgPatchHelper(patch))
+                for patch in patches_bytes
+            ]
+
+        if patch_format == PatchFormat.GitFormatPatch:
+            return [
+                build_revision_from_patch_helper(GitPatchHelper(patch))
+                for patch in patches_bytes
+            ]
```
zzzeid, line 96: - Ideally the try block will only wrap the specific calls that would cause an exception
- The two if statements should go outside the try block

zzzeid (changes requested):
Looks great, just a couple more comments and this should be good to go.

landoapi/api/try_push.py
```diff
@@ -40,6 +40,18 @@ class PatchFormat(enum.Enum):
...
+        """Return the `PatchHelper` subclass associated with this patch format."""
+        if self == PatchFormat.GitFormatPatch:
+            return GitPatchHelper(patch_io)
+
+        if self == PatchFormat.HgExport:
+            return HgPatchHelper(patch_io)
+
+        raise ValueError(
+            f"Patch format {self.value} has not been mapped to a patch helper."
+        )
```
zzzeid, line 53: nit: I think it would be cleaner if this was moved outside the enum class. Maybe also rename it to `get_patch_helper`.

landoapi/api/try_push.py
```diff
@@ -82,18 +94,13 @@ def parse_revisions_from_request(
         io.BytesIO(convert_json_patch_to_bytes(patch)) for patch in patches
     )
 
+    patch_helpers = (patch_format.patch_helper(patch) for patch in patches_bytes)
+
     try:
-        if patch_format == PatchFormat.HgExport:
```
zzzeid: It looks like `patch_format` is now unused?

zzzeid (changes requested):

landoapi/api/try_push.py
cgsheeh: It's used to generated `patch_helpers` just above.

zzzeid: Thanks, never mind!

zzzeid: I think if `patch_helper` is moved outside the Enum it would make this also clearer, even though this is fuctional. I will bundle this along with the nit though, since I don't think it's a blocker.

landoapi/api/try_push.py
cgsheeh: I renamed the function but I think this makes more sense as an instance method on `PatchFormat`. The function is just a mapping of the patch format to the associated helper class. Writing this as a standalone function turns the usage from `patch_format.get_patch_helper(patch)` to `get_patch_helper(patch_format, patch)`, and moves the mapping outside the enum where it is more relevant.

landoapi/hgexports.py
```diff
@@ -4,21 +4,31 @@
...
 )
 
-HEADER_NAMES = (
-    b"User",
-    b"Date",
-    b"Node ID",
-    b"Parent",
-    b"Diff Start Line",
-    b"Fail HG Import",
+from dateutil.parser import parse as dateparse
```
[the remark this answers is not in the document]
cgsheeh: I didn't know the exact format of the `Date` header at the time so I went with `dateutil` since it can parse string dates in various formats.

zzzeid (changes requested):

zzzeid (approved):
lgtm!

PR #330, opened 2023-06-19
typing: add types to `hgexports.py` and fix `set_patch` types (Bug 1759890)

cgsheeh:
While testing the try API endpoint I noticed my very simple
patch was failing to apply. This is because the type hint for
`Revision.set_patch` is incorrect and the raw diff returned
from Phabricator is actually a `str`, not `bytes`, and thus the
type I was using was incorrect. The timestamp is also a `str`
and not an `int`. Update the function signature to match reality,
and while we're here add type hints to all of `hgexports.py`.

### DIFF
--- a/landoapi/hgexports.py
+++ b/landoapi/hgexports.py
@@ -1,6 +1,10 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from __future__ import annotations
+
+import io
 import re
 from typing import (
     Optional,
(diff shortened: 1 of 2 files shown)


zzzeid (approved):
nit: RE: commit message, It is inaccurate to state that the patch was failing to apply because of an incorrect type hint (albeit it may have been misleading). Rewording this commit message to be more accurate (e.g., "fixed incorrect type hint") should suffice here.

>While testing the try API endpoint I noticed my very simple
patch was failing to apply. This is because the type hint for
Revision.set_patch is incorrect and the raw diff returned
from Phabricator is actually a str, not bytes. The timestamp
is also a str and not an int. Update the function signature
to match reality, and while we're here add type hints to all of
hgexports.py.

cgsheeh:
> nit: RE: commit message, It is inaccurate to state that the patch was failing to apply because of an incorrect type hint (albeit it may have been misleading). Rewording this commit message to be more accurate (e.g., "fixed incorrect type hint") should suffice here.
> 
> > While testing the try API endpoint I noticed my very simple
> > patch was failing to apply. This is because the type hint for
> > Revision.set_patch is incorrect and the raw diff returned
> > from Phabricator is actually a str, not bytes. The timestamp
> > is also a str and not an int. Update the function signature
> > to match reality, and while we're here add type hints to all of
> > hgexports.py.

Fixed.

PR #336, opened 2023-08-10
hgexports: refactor patch parsing code (Bug 1846330)

cgsheeh:
Refactor some patch parsing code in the `HgPatchHelper` to
clarify what is taking place. Use `enumerate` instead of a
manually incremented looping variable. Flatten if-statement
logic to reduce repeated behaviour and add a comment describing
what is taking place within the loop.


zzzeid (approved):
lgtm

PR #337, opened 2023-08-28
uplift: strip `Depends on` text from commit message (Bug 1850387)

cgsheeh:
Adds a `strip_depends_on_from_commit_message` function that removes
`Depends on` lines from commit messages. Pass the commit message
of the original revision through this function before creating the
uplift revision to avoid linking uplift revisions to the original
revision.

### DIFF
--- a/landoapi/uplift.py
+++ b/landoapi/uplift.py
@@ -4,6 +4,7 @@
 
 import json
 import logging
+import re
 import time
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


zzzeid (approved):
lgtm.

zzzeid:
> [Here is an example of this in action.](https://phabricator-dev.allizom.org/D3556)

Link does not work due to permissions issue.

PR #343, opened 2023-09-19
hg: retry after Mercurial repo lock waiting timeouts (Bug 1853965)

cgsheeh:
When pushes fail due to timeouts waiting for the repo lock,
Lando should wait and retry the push. Add a new exception for
this condition that includes snippets of the expected output
from Mercurial, and add the exception to the block of temporary
failure exceptions when trying to push.

### DIFF
--- a/landoapi/hg.py
+++ b/landoapi/hg.py
@@ -43,7 +43,13 @@ class HgException(Exception):
             err.decode(errors="replace"),
         ).rstrip()
 
-        for cls in (LostPushRace, PatchConflict, TreeClosed, TreeApprovalRequired):
+        for cls in (
+            LostPushRace,
+            PatchConflict,
+            TreeClosed,
+            TreeApprovalRequired,
+            TryPushTimeoutException,
+        ):
             for s in cls.SNIPPETS:
                 if s in err or s in out:
                     return cls(msg)
@@ -78,6 +84,12 @@ class LostPushRace(HgException):
         b"abort: push creates new remote head",
         b"repository changed while pushing",
     )
+
+
+class TryPushTimeoutException(HgException):
+    """Exception when pushing failed due to a timeout on the Try repo."""
+
+    SNIPPETS = (b"abort: working directory of /repo/hg/mozilla/try: timed out waiting",)
 
 
 class PatchApplicationFailure(HgException):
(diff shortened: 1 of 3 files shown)


zzzeid (changes requested):
Wondering if this should be a more generic exception. Something like `PushTimeoutException`? That might be cleaner than hardcoding the error output (which includes the path).

cgsheeh:
> Wondering if this should be a more generic exception. Something like `PushTimeoutException`? That might be cleaner than hardcoding the error output (which includes the path).

My intention was to have this be an exception which is only considered for try-related pushes. Do you think this is something we should retry for in all cases? I could change the snippet to `timed out waiting for lock held by` in that case.

zzzeid:
> My intention was to have this be an exception which is only considered for try-related pushes. Do you think this is something we should retry for in all cases? I could change the snippet to `timed out waiting for lock held by` in that case.

I think a broader check would fit better with the general conventions we have in those push exceptions (most of them are general). Since there is no extra cost (these checks are running on every push) I don't see any downside to making it more general. If we want to keep it specific to try pushes, maybe you can replace the hardcoded path with a path fetched from the repo config instead?

landoapi/hg.py
```diff
@@ -80,6 +86,12 @@ class LostPushRace(HgException):
     )
 
 
+class PushTimeoutException(HgException):
+    """Exception when pushing failed due to a timeout on the Try repo."""
```
zzzeid, line 90:
```suggestion
    """Exception when pushing failed due to a timeout on the repo."""
```

zzzeid (changes requested):

cgsheeh:
> > My intention was to have this be an exception which is only considered for try-related pushes. Do you think this is something we should retry for in all cases? I could change the snippet to `timed out waiting for lock held by` in that case.
> 
> I think a broader check would fit better with the general conventions we have in those push exceptions (most of them are general). Since there is no extra cost (these checks are running on every push) I don't see any downside to making it more general. If we want to keep it specific to try pushes, maybe you can replace the hardcoded path with a path fetched from the repo config instead?
> 
> Edit: bonus advantage is more robust testing, since the paths can be different on different environments.

I made the check apply to all repos, not just try. :+1:

zzzeid (approved):

PR #344, opened 2023-09-27
landing_jobs: raise a `ProblemException` for improper date format (Bug 1855566)

cgsheeh:
Verify the value passed in the query string is indeed a datetime
object.

### DIFF
--- a/landoapi/api/landing_jobs.py
+++ b/landoapi/api/landing_jobs.py
@@ -74,6 +74,22 @@ def put(landing_job_id: str, data: dict):
         )
 
 
+def try_convert_datetime(date: str) -> datetime:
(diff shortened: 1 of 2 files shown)


zzzeid (approved):
lgtm, however, I also don't think we're using this endpoint any more, and can probably just remove it altogether from the API.

landoapi/api/landing_jobs.py
```diff
@@ -74,6 +74,22 @@ def put(landing_job_id: str, data: dict):
         )
 
 
+def try_convert_datetime(date: str) -> datetime:
```
zzzeid, line 77:
```suggestion
def datetime_from_string(date: str) -> datetime:
```

I think the word `try` is not necessary and could be slightly confusing.

PR #347, opened 2023-10-06
landings: use `StringIO` instead of `BytesIO` everywhere (Bug 1855328)

cgsheeh:
Transition from using `io.BytesIO` to `io.StringIO` for content buffers.
Update uses of `NamedTemporaryFile` to open in `w+` mode with a specified
encoding.

### DIFF
--- a/landoapi/models/revisions.py
+++ b/landoapi/models/revisions.py
@@ -82,6 +82,11 @@ class Revision(Base):
         )
         return f"<{self.__class__.__name__}: {self.id}{phab_identifier}>"
 
+    @property
+    def patch_string(self) -> str:
+        """Return the patch as a UTF-8 encoded string."""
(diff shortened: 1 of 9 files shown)


landoapi/models/revisions.py
```diff
@@ -82,6 +82,11 @@ def __repr__(self):
         )
         return f"<{self.__class__.__name__}: {self.id}{phab_identifier}>"
 
+    @property
+    def patch_text(self):
```
zzzeid, line 86: nit: missing type hint, perhaps rename to `patch_string`.

zzzeid (approved):
nit: I think it would be better to separate the changes that fix this bug, from any additional changes to the implementation.

zzzeid (approved):

PR #348, opened 2023-10-10
hgexports: read buffer to EOF instead of in chunks (Bug 1855328)

cgsheeh:
Remove the buffer read size argument from calls to `BytesIO.read`
to instead read to EOF. Parametrize the integrated landing job
with a test case that would fail on the current behaviour due
to an improperly decoded string.


zzzeid (approved):
lgtm!

PR #353, opened 2023-10-30
uplift: set `sourceControlBaseRevision` as latest commit on target repo (Bug 1843041)

cgsheeh:
Add the `metrics` attachment to the Conduit API call for the target
repo, which adds the most recent commit hash from that repo to the
returned data. Use this commit hash as the `sourceControlBaseRevision`
value when uploading the diff to Phabricator. This mirrors how the
value is set in `moz-phab`. This fixes the mangled diff appearance
on Phabricator, and will fix the code review bot runs on uplift
revisions.


zzzeid (approved):
lgtm.
nit: maybe add some test coverage if that's possible.

PR #355, opened 2023-11-08
emails: make landing failure emails less Phab-specific (Bug 1855851)

cgsheeh:
Make the Phabricator landing emails less Phabricator-specific.
Update the landing job identifier to include the context that
the job belongs to a try push with the included commit message
title. Add a check of the landing job identifier to only display
a link to a Lando-UI URL when the identifier is a Phab revision
ID.

### DIFF
--- a/landoapi/email.py
+++ b/landoapi/email.py
@@ -4,12 +4,14 @@
 import logging
 from email.message import EmailMessage
 
+from landoapi.validation import REVISION_ID_RE
+
 logger = logging.getLogger(__name__)
 
 LANDING_FAILURE_EMAIL_TEMPLATE = """
 Your request to land {landing_job_identifier} failed.
 
-See {lando_revision_url} for details.
+See {error_details_location} for details.
 
 Reason:
 {reason}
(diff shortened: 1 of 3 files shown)


landoapi/email.py
```diff
@@ -4,12 +4,14 @@
...
 
+from landoapi.validation import REVISION_ID_RE
+
 logger = logging.getLogger(__name__)
 
 LANDING_FAILURE_EMAIL_TEMPLATE = """
 Your request to land {landing_job_identifier} failed.
 
-See {lando_revision_url} for details.
+See {error_details_location} for details.
```
zzzeid, line 14: Maybe we don't need this line at all, if there is no link? Instead of "see below". This would also simplify the logic in the rest of this file.

zzzeid (changes requested):

landoapi/email.py
```diff
@@ -38,19 +36,17 @@ def make_failure_email(
...
+    revision_status_url = ""
     if REVISION_ID_RE.match(landing_job_identifier):
         # If the landing job identifier looks like a Phab revision,
         # link to the relevant view page.
         error_details_location = f"{lando_ui_url}/{landing_job_identifier}/"
-    else:
-        # If there is no relevant URL to show the user in the email, the error
-        # details in the bottom of the email is all we have to show them for now.
-        error_details_location = "below"
+        revision_status_url = f"\nSee {error_details_location} for details.\n"
```
zzzeid, line 44: nit: having a variable named `revision_status_url` would imply it contains a URL, but that is not the case here.

zzzeid (approved):

PR #359, opened 2023-11-15
landing-worker: add an `exception` level logging call when jobs do not succeed (Bug 1864912)

cgsheeh:


landoapi/workers/landing_worker.py
```diff
@@ -422,6 +422,7 @@ def run_job(
                 return False  # Try again, this is a temporary failure.
             except Exception as e:
                 message = f"Unexpected error while pushing to {repo.push_path}.\n{e}"
+                logger.exception(message)
```
zzzeid, line 425: The error related to this change happened in the section above this one (where `False` is returned, not `True`.)

zzzeid (changes requested):

landoapi/workers/landing_worker.py
```diff
@@ -416,12 +418,14 @@ def run_job(
                     f"`Temporary error ({e.__class__}) "
                     f"encountered while pushing to {repo_push_info}"
                 )
+                logger.exception(message)
```
zzzeid, line 421: nit: I'm not sure we need the full stack trace for all of these errors, since all except `HgmoInternalServerError` are very specific. Might make more sense to move `except HgmoInternalServerError` to its own line instead.

zzzeid (approved):

landoapi/workers/landing_worker.py
cgsheeh: I think having more info in the logs as a default makes sense, we can always filter that out if the full stack trace is too much context. It would be worse to have an incident in progress and have to re-deploy Lando to add more logging so we could fix the problem. :)

PR #362, opened 2023-11-20
landing-job: add a type hint to `revisions` (Bug 1759890)

cgsheeh:
Add a type hint for `LandingJob.revisions`. This fixes LSP features on this field,
which don't work at the moment as the LSP can't infer the type from SQLAlchemy's
string-based type hints.

### DIFF
--- a/landoapi/models/landing_job.py
+++ b/landoapi/models/landing_job.py
@@ -132,7 +132,7 @@ class LandingJob(Base):
     # Identifier of the published commit which this job should land on top of.
     target_commit_hash = db.Column(db.Text(), nullable=True)
 
-    revisions = db.relationship(
+    revisions: list[Revision] = db.relationship(
         "Revision",
         secondary=revision_landing_job,
         back_populates="landing_jobs",


cgsheeh:
Note: I wanted to add a type hint for `Revision.landing_jobs` as well but since they are defined in separate modules it would be circular dependency to add them both. I suppose I could add the landing jobs type hint as `list["LandingJob"]`, but the coupling of those two models makes me wonder if they should be defined in the same file.

landoapi/models/landing_job.py
```diff
@@ -132,7 +132,7 @@ class LandingJob(Base):
     # Identifier of the published commit which this job should land on top of.
     target_commit_hash = db.Column(db.Text(), nullable=True)
 
-    revisions = db.relationship(
+    revisions: list[Revision] = db.relationship(
```
zzzeid, line 135: nit: `InstrumentedList` would be more accurate here? Although The type is also ambiguous (could be `RelationshipProperty`, `InstrumentedAttribute`, or `InstrumentedList` depending on how you're looking at this object).

What benefit does adding this type hint provide us, and why just here and nowhere else?

zzzeid (changes requested):

zzzeid (approved):
> What benefit does adding this type hint provide us, and why just here and nowhere else?

Ah, just noticed the reference to the LSP in the description, never mind. I am not really sure as to whether we should be making specific change to support features that are dependent on a particular developer environment/preferences, vs. say something that is more universally beneficial to the development environment or workflow.

It can end up shaping the code in a biased way (for example, in this PR which kind of looks out of place to someone who doesn't have the same development environment as you do.)

landoapi/models/landing_job.py
cgsheeh: The issue right now is that using language features on any subfields of revision don't work, because the LSP can't infer that `revisions` is a list of `Revision` objects due to the dynamic nature of `db.relationship`. Adding this allows using those features on the `revisions` field. For example, try using "get all references" on `Revision.revision_id`, it won't return the usage in `LandingWorker.run_job`. Running it on that usage won't return any references either. 

We could add type hints to the other fields too, but this specific type hint adds tooling features, so I made a point of adding it.

cgsheeh: Also, we could make the type hint reflect the actual ORM type, but since we don't typically use any of those features I think `list[Revision]` is fine (that's really how we think of the `revisions` attribute anyway). We're planning to tear all this out soon enough, so having this in the short term while we're still on Flask-Lando seems reasonable.

zzzeid (approved):

cgsheeh:
> > What benefit does adding this type hint provide us, and why just here and nowhere else?
> 
> Ah, just noticed the reference to the LSP in the description, never mind. I am not really sure as to whether we should be making specific change to support features that are dependent on a particular developer environment/preferences, vs. say something that is more universally beneficial to the development environment or workflow.
> 
> It can end up shaping the code in a biased way (for example, in this PR which kind of looks out of place to someone who doesn't have the same development environment as you do.)

Which LSP are you using? This should be a benefit for most development environments.

PR #363, opened 2023-11-28
hgexports: empty Git commits aren't parsed correctly (Bug 1867036)

cgsheeh:
When empty Git commits are sent to Lando in `git format-patch`
format, Lando fails to parse the commit correctly. Currently
Lando will fail to find the `---` line used to separate the
commit message from the diff, and assume all lines in the
patch body are part of the commit message. The parsing will
then fail later down the line as Lando will not find a valid
diff line in the body.

Add an `else` clause to the loop which looks for the `---` line
separating the extended commit message and the diff, which when
reached indicates that no diff is present on the commit. In this
case we strip the last two lines of the gathered commit message
and return an empty diff. Add a test for parsing a patch generated
from an empty commit to `test_hgexports.py`.

### DIFF
--- a/landoapi/hgexports.py
+++ b/landoapi/hgexports.py
@@ -319,6 +319,13 @@ class GitPatchHelper(PatchHelper):
                 commit_message_lines += [""]
 
             commit_message_lines.append(line)
+        else:
+            # We never found the end of the commit message body, so this change
+            # must be an empty commit. Discard the last two lines of the
+            # constructed commit message which are Git version info and return
+            # an empty diff.
+            commit_message = "\n".join(commit_message_lines[:-2])
+            return commit_message, ""
 
         commit_message = "\n".join(commit_message_lines)
 
--- a/tests/test_hgexports.py
+++ b/tests/test_hgexports.py
@@ -122,6 +122,21 @@ index f56ba1c..33391ea 100644
          500,
 """.rstrip()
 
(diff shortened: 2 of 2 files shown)


landoapi/hgexports.py
```diff
@@ -319,6 +319,13 @@ def parse_email_body(self, content: str) -> tuple[str, str]:
                 commit_message_lines += [""]
 
             commit_message_lines.append(line)
+        else:
```
zzzeid, line 322: Are there any more specific/explicit conditions we could check here?

zzzeid (approved):

landoapi/hgexports.py
cgsheeh: Unfortunately I don't think so. The format essentially has subject/date/author etc information in the Unix Mbox style headers, and then the message body. The message body looks like:

- the extended commit message if one is available (ie a commit message longer than 1 line).
- a `---` line to signify the end of the commit message.
- the `--stat` output from Git (ie like you would find for `git log --stat`).
- a line beginning with `diff --git` that signifies the start of the diff, followed by all the lines of the diff.
- the final two lines of the patch body are `--` on one line, followed by the Git version used to generate the patch on the last line.

If the commit is empty, the `---` isn't present to signal the end of the commit message, and only the `--` + Git version is found. So I think it makes sense to consider commits where we don't find `---`  as empty commits.

PR #364, opened 2023-12-12
uplifts: query `differential.querydiffs` to find the latest non-`commit` diff (Bug 1709608)

cgsheeh:
Extend `get_uplift_conduit_state` to also query `differential.querydiffs`
and return all known diffs associated with the selected uplift
revisions. When creating the uplift revisions, add a function to
parse through the diffs and return the latest diff that does not
have a `creationMethod` of `commit`. This works around an issue
in Phabricator where diffs are created with empty binary files
when a new diff is created after parsing a commit from the observed
version control repository.

### DIFF
--- a/landoapi/uplift.py
+++ b/landoapi/uplift.py
@@ -2,10 +2,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import collections
 import json
 import logging
 import re
 import time
+from operator import itemgetter
 from typing import (
     Any,
     Optional,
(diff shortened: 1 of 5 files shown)


landoapi/uplift.py
```diff
@@ -143,9 +160,25 @@ def get_uplift_conduit_state(
...
+    rev_ids = [revision["id"] for revision in stack_data.revisions.values()]
+    rev_ids_to_all_diffs = get_rev_ids_to_diffs(phab, rev_ids)
+
     stack = RevisionStack(set(stack_data.revisions.keys()), edges)
 
-    return stack_data, stack, target_repo
+    return stack_data, stack, target_repo, rev_ids_to_all_diffs
+
+
+def get_latest_good_binary_diff(diffs: list[dict]) -> dict:
```
zzzeid, line 171: The method name and docstring can be improved here. "good" doesn't convey any meaning in this context, and referencing a bug in the docstring first line is not ideal. There is no way to tell what this method is trying to do by reading the docstring or method name.

landoapi/uplift.py
```diff
@@ -143,9 +160,25 @@ def get_uplift_conduit_state(
...
+    """Given a list of diff dicts, return the latest diff that avoids bug 1865760."""
+    # Iterate through the diffs in order of the latest IDs.
+    for diff in sorted(diffs, key=itemgetter("id"), reverse=True):
+        # Diffs with a `creationMethod` of `commit` may have bad binary data.
+        if diff["creationMethod"] == "commit":
+            continue
+
+        return diff
```
zzzeid, line 179: nit: I think it would be a little easier to follow what is happening here if we first filter out unwanted records, then return the first/last desired one left, outside the for loop.
```suggestion
    diffs = sorted(diffs, key=itemgetter("id"))
    diffs = [diff for diff in diffs if diff["creationMethod"] != commit]
    if diffs:
        return diffs[-1]
```

zzzeid (changes requested):

landoapi/uplift.py
```diff
@@ -143,9 +160,25 @@ def get_uplift_conduit_state(
...
+    rev_ids_to_all_diffs = get_rev_ids_to_diffs(phab, rev_ids)
+
     stack = RevisionStack(set(stack_data.revisions.keys()), edges)
 
-    return stack_data, stack, target_repo
+    return stack_data, stack, target_repo, rev_ids_to_all_diffs
+
+
+def get_latest_non_commit_diff(diffs: list[dict]) -> dict:
+    """Given a list of diff dicts, return the latest diff that avoids bug 1865760."""
```
zzzeid, line 172: This still references a bug number, it would be better to actually summarize the behaviour here instead. The context of the bug is nice to have as an explanation of why this method is implemented, but ideally the docstring describes what this method is doing, not just why.

I.e.
```python
"""Return a non-commit diff given a list of diff dicts."""
```

You can elaborate more in additional lines in the docstring.

zzzeid (changes requested):

PR #368, opened 2023-12-22
testing: add test for partial stack request

cgsheeh:
While debugging bug 1871425 we orignally had a suspicion that
the issue was resulting from Lando erroneously applying later
patches in the stack when it should only have applied the
requested patches. I noticed we don't have a test for this
case, so I added one myself, and it passed as expected. We
might as well keep this simple test around just to provide
extra test coverage of this use case.

### DIFF
--- a/tests/test_transplants.py
+++ b/tests/test_transplants.py
@@ -716,6 +716,60 @@ def test_integrated_transplant_simple_stack_saves_data_in_db(
     assert job.landed_revisions == {1: 1, 2: 2, 3: 3}
 
 
+def test_integrated_transplant_simple_partial_stack_saves_data_in_db(
+    db,
+    client,
+    phabdouble,
+    auth0_mock,
+    release_management_project,
+    register_codefreeze_uri,
+):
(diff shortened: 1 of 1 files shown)


tests/test_transplants.py
```diff
@@ -716,6 +716,60 @@ def test_integrated_transplant_simple_stack_saves_data_in_db(
...
+
+    d2 = phabdouble.diff()
+    r2 = phabdouble.revision(diff=d2, repo=phabrepo, depends_on=[r1])
+    phabdouble.reviewer(r2, user)
+
+    d3 = phabdouble.diff()
+    r3 = phabdouble.revision(diff=d3, repo=phabrepo, depends_on=[r2])
+    phabdouble.reviewer(r3, user)
+
+    # Request a transplant, but only for the first revision in the stack.
```
zzzeid, line 743: This comment does not match what this test is doing. Perhaps you meant to only include one revision, or this should say "first two revisions"?

zzzeid (changes requested):

zzzeid (approved):

PR #369, opened 2023-12-29
uplift: use `differential.querydiffs` data to populate new diff changes (Bug 1843041)

cgsheeh:
Currently a Lando based uplift request populates the required fields
in Phabricator by pulling down the raw diff from `differential.getrawdiff`,
passing the diff through `rs-parsepatch` and populating a dict in
the format Phabricator expects. This has worked decently well to date,
however this method doesn't include the full file context, leading
to diff display issues on Phabricator.

The `differential.querydiffs` endpoint returns data about a diff in
the expected Phabricator format. We have already started using this
data to populate the binary files metadata for uplift requests.
Taking a closer look at the response we can see that the full file
context is included in this format, and the diff itself is the
exact same diff we expect to be uploaded as the uplift request on
Phabricator.

Remove the code which parses the `differential.getrawdiff` response
into the expected Phabricator format and instead simply re-use the
data from the `differential.querydiffs` response for the uplift
revision. This will provide a more accurate view of the diff that
will land on the uplift train. The content of the file itself may
be different than what exists on the target train, however this
problem already exists in the existing Lando uplift implementation.


zzzeid (approved):
I assume this change only affects uplifts -- I guess this was partially implemented before you fully implemented the uplift flow.

I also noticed that the `differential.querydiffs` endpoint is deprecated. I'm not sure what impact that would have given that we're on our own fork at this point, but something to keep in mind.

cgsheeh:
> I assume this change only affects uplifts -- I guess this was partially implemented before you fully implemented the uplift flow.

Yes it did work for the main use case, creating a new diff that is landable with Lando. However now that we've stress tested it, we're hitting all the various edge cases.

> I also noticed that the `differential.querydiffs` endpoint is deprecated. I'm not sure what impact that would have given that we're on our own fork at this point, but something to keep in mind.

I doubt anyone on our team is going to be removing deprecated Phabricator endpoints, and Phacility certainly won't be, so I think we're safe. :sweat_smile:

PR #375, opened 2024-02-13
transplants: use `values()` instead of `items()` since key is unused

cgsheeh:
Also rename the `r` variable to `reviewers` while we are here.


dklawren (approved):
LGTM

PR #379, opened 2024-02-27
transplants: use `any` instead of `for`/`else` to check valid landing

cgsheeh:


zzzeid (approved):
Was this just to improve readability?

cgsheeh:
> Was this just to improve readability?

Yes, `for`/`else` with a simple break clause is really just an `any`. :)

PR #382, opened 2024-03-06
revisions: block revisions with `needs-data-classification` tag from landing (Bug 1880861)

cgsheeh:
Add a blocking check that enforces revisions with the `needs-data-classification`
tag cannot land. The check is similar to the testing policy project tag checks
except it simply looks for the `needs-data-classification` tag and
blocks landing if it is present on the revision. The tag requires the
PHID of the project tag so it is generated with a factory function in a
similar manner to the release manager approval blocking check.

### DIFF
--- a/landoapi/revisions.py
+++ b/landoapi/revisions.py
@@ -132,7 +132,7 @@ def check_diff_author_is_known(*, diff: dict, **kwargs) -> Optional[str]:
     )
 
 
-def revision_has_needs_data_classification(
+def revision_has_needs_data_classification_tag(
     revision: dict, data_policy_review_phid: str
 ) -> bool:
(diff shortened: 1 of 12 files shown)


landoapi/revisions.py
```diff
@@ -132,6 +132,28 @@ def check_diff_author_is_known(*, diff: dict, **kwargs) -> Optional[str]:
     )
 
 
+def revision_has_needs_data_classification(
```
zzzeid, line 135: nit
```suggestion
def revision_has_needs_data_classification_tag(
```

zzzeid (approved):
lgtm. I left out any comments related to the way this is implemented, since it follows the current conventions and we plan on reworking that after porting Lando.

PR #383, opened 2024-03-06
workers: add landing worker warning logs

zzzeid:
- add warning log for large queue size (bug 1879896)
- add warning log for too many attempts (bug 1880310)

### DIFF
--- a/landoapi/workers/base.py
+++ b/landoapi/workers/base.py
@@ -152,7 +152,6 @@ class Worker:
             for r in self.applicable_repos
             if treestatus_subsystem.client.is_open(repo_clone_subsystem.repos[r].tree)
         ]
-        logger.info(f"{len(self.enabled_repos)} enabled repos: {self.enabled_repos}")
 
     def start(self, max_loops: int | None = None):
         """Run setup sequence and start the event loop."""
--- a/landoapi/workers/landing_worker.py
+++ b/landoapi/workers/landing_worker.py
@@ -86,10 +86,36 @@ class LandingWorker(Worker):
         self.last_job_finished = None
         self.refresh_enabled_repos()
 
-    def loop(self):
-        logger.debug(
-            f"{len(self.applicable_repos)} applicable repos: {self.applicable_repos}"
+    def check_landing_worker_warnings(self):
+        """Log messages that show various important statistics about the landing worker."""
+        TOO_MANY_ATTEMPTS_THRESHOLD = 10
+        # TODO: should this threshold be different for try landing worker?
+        QUEUE_SIZE_THRESHOLD = 20
+        ACTIVE_STATUSES = (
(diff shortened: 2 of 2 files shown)


landoapi/workers/landing_worker.py
```diff
@@ -86,11 +86,37 @@ def __init__(self, *args, **kwargs):
...
+    def check_landing_worker_warnings(self):
+        """Log messages that show various important statistics about the landing worker."""
+        TOO_MANY_ATTEMPTS_THRESHOLD = 10
+        # TODO: should this threshold be different for try landing worker?
+        QUEUE_SIZE_THRESHOLD = 20
+        ACTIVE_STATUSES = (
+            LandingJobStatus.SUBMITTED,
+            LandingJobStatus.DEFERRED,
+            LandingJobStatus.IN_PROGRESS,
+        )
```
cgsheeh, line 98: This tuple is a copy of the `applicable_statuses` variable in `job_queue_query`. Let's move them to a variable where they can be shared (maybe a class-level `LandingJobStatus.active_statuses` or something along those lines).

landoapi/workers/landing_worker.py
```diff
@@ -86,11 +86,37 @@ def __init__(self, *args, **kwargs):
...
+        runaway_jobs = LandingJob.query.filter(
+            LandingJob.status.in_(ACTIVE_STATUSES),
+            LandingJob.attempts >= TOO_MANY_ATTEMPTS_THRESHOLD,
         )
 
+        if runaway_jobs.count() > 0:
+            job = runaway_jobs.all()[0]
+            logger.warning(
+                f"Active landing job ({job}) has too many attempts ({job.attempts})"
+            )
```
cgsheeh, line 115: Should we cancel the job in that case?

cgsheeh (changes requested):

landoapi/workers/landing_worker.py
zzzeid: That would be a good idea, but I think it would be outside the scope of this PR since it would require us to also send a notification email. Maybe it can be done as a follow up.

landoapi/workers/base.py
```diff
@@ -152,7 +152,6 @@ def refresh_enabled_repos(self):
             for r in self.applicable_repos
             if treestatus_subsystem.client.is_open(repo_clone_subsystem.repos[r].tree)
         ]
-        logger.info(f"{len(self.enabled_repos)} enabled repos: {self.enabled_repos}")
```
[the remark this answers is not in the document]
zzzeid: Yes, I am not sure if it is a useful thing to have in the logs constantly, since this is basically the repos in the environment variable, minus any trees that are closed. Perhaps it makes more sense to add it as a warning instead, if the two don't match?

PR #384, opened 2024-03-06
landing_job: fix ordering of jobs in queue (bug 1878666)

zzzeid:
- order first by priority, then by status and creation time
- decrease job priority by 1 when it is deferred

### DIFF
--- a/landoapi/models/landing_job.py
+++ b/landoapi/models/landing_job.py
@@ -248,11 +248,7 @@ class LandingJob(Base):
             grace_cutoff = now - datetime.timedelta(seconds=grace_seconds)
             q = q.filter(cls.created_at < grace_cutoff)
 
-        # Any `LandingJobStatus.IN_PROGRESS` job is first and there should
-        # be a maximum of one (per repository). For
-        # `LandingJobStatus.SUBMITTED` jobs, higher priority items come first
-        # and then we order by creation time (older first).
-        q = q.order_by(cls.status.desc(), cls.priority.desc(), cls.created_at)
+        q = q.order_by(cls.priority.desc(), cls.status.desc(), cls.created_at)
 
         return q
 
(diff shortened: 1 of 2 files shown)


landoapi/models/landing_job.py
```diff
@@ -376,6 +372,8 @@ def transition_status(
 
         if action in (LandingJobAction.FAIL, LandingJobAction.DEFER):
             self.error = kwargs["message"]
+            if action == LandingJobAction.DEFER:
+                self.priority -= 1
```
cgsheeh, line 376: The way this is written now, any job which has its priority lowered will always be at the back of the queue, meaning the patch effectively needs to wait until the queue is empty before it will be considered for landing. This could cause a situation where someone queues their revision for landing, it doesn't land the first time due to some transient issue, then their landing is stuck in limbo because it has a lower priority than newer revisions that are added to the queue. If the queue is empty often enough it's probably not a big deal, but worth noting here.

As an alternative we could note the last attempt time in the landing job, and filter out revisions that have been deferred for a certain duration before attempting to land again. For example we defer a landing job, and then in the job query we filter out all deferred revisions with a last attempted time within the last 15-20 minutes.

landoapi/models/landing_job.py
```diff
@@ -248,11 +248,7 @@ def job_queue_query(
             grace_cutoff = now - datetime.timedelta(seconds=grace_seconds)
             q = q.filter(cls.created_at < grace_cutoff)
 
-        # Any `LandingJobStatus.IN_PROGRESS` job is first and there should
-        # be a maximum of one (per repository). For
-        # `LandingJobStatus.SUBMITTED` jobs, higher priority items come first
-        # and then we order by creation time (older first).
-        q = q.order_by(cls.status.desc(), cls.priority.desc(), cls.created_at)
+        q = q.order_by(cls.priority.desc(), cls.status.desc(), cls.created_at)
```
cgsheeh, line 251: Do we need this change? We already get the highest priority items with the highest priority statuses, right?

cgsheeh (changes requested):

PR #385, opened 2024-03-20
tests: use correct loop variable and rename to `revision`

cgsheeh:
In the list comprehension where we set `job.revision_order` in
`create_landing_job_with_no_linked_revisions`, we iterate over
`revisions` and set the iteration variable as `r`, but the
value that ends up in the list uses the leftover `revision`
variable from the loop iteration above. Rename `r` to `revision`
so the correct variable is used.


zzzeid (approved):
nice.

PR #388, opened 2024-04-10
stacks: unify and cleanup landing checks system (Bug 1888188)

cgsheeh:
This commit begins the process of unifying and refactoring the
landing checks system, which blocks stacks and revisions from
landing and issues warnings about various issues in the landing
that are not blockers. The state of this revision gives us a
base to work on for extending the lints system to add Mercurial
commit hooks, clean up some unnecessary data fields, and improve
the user experience around presentation of landing lints.

The existing checks are split up into three categories. First are
stack-level blocker checks, which inspect the entire state of the
stack and block landings entirely. The second are revision-level
blocker checks which inspect the state of individual revisions and
block landing of each revision and it's successor nodes. Finally
are the revision-level warnings, which do not block landings but
will present information about sub-optimal landing conditions that
must be acknowledged before landing.

Each piece of state used by individual checks is added to a new
`TransplantAssessmentState` class. An instance of this class is
passed to each lint to keep the API consistent and discoverable.
Some fields on this class have been moved directly from existing
checks, and future refactors should reduce the number of fields on
this class.

The transplant assessment API and the `calculate_landable_subgraphs`
API are combined into a single API as part of this commit. Since
the previous setup had two different use cases, some state is not
required to run landing checks on non-transplant assessments. All
pieces of state required to check for a valid landing request are
encapsulated in `LandingAssessmentState`, which is an optional field
on `TransplantAssessmentState`. Checks which should only run on
landing assessments can look for the existance of a `LandingAssessmentState`
on the `TransplantAssessmentState`.

The `calculate_landable_subgraphs` API and the custom graph walking
code is replaced by the new APIs. When building the `TransplantAssessmentState`
we create a deep copy of the `RevisionStack` object and name it the
`landing_stack`, where blocked revisions are removed from this stack.
Then all landable paths in the `landing_stack` from the root revisions
to the leaf revisions are considered to be landable subgraphs. A new
`leaf_revisions` method is added which returns all nodes that have
no successors in the graph. Each node in the original `stack` object has a
`blocked` attribute which is initialized to an empty list. When a
blocking check fails on a revision the reason is appended to the
list in the full revision stack.


landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
+        reviewers[revision["phid"]] = get_collated_reviewers(revision)
+
+    # Get more Phabricator data.
+    involved_phids = list(involved_phids)
+    users = user_search(phab, involved_phids)
+    projects = project_search(phab, involved_phids)
+
+    secure_project_phid = get_secure_project_phid(phab)
+    testing_tag_project_phids = get_testing_tag_project_phids(phab)
+    testing_policy_phid = get_testing_policy_phid(phab)
```
zzzeid, line 895: I think this should also be done elsewhere, perhaps in the same place as relman_group_phid and data_policy_review_phid. Maybe they can all be put in a dataclass somewhere and fetched directly from there. Slightly out of scope, but at least we should not be introducing more of these calls in these methods.

landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
+        supported_repos=supported_repos,
+        reviewers=reviewers,
+        users=users,
+        projects=projects,
+        data_policy_review_phid=data_policy_review_phid,
+        relman_group_phid=relman_group_phid,
+        secure_project_phid=secure_project_phid,
+        testing_tag_project_phids=testing_tag_project_phids,
+        testing_policy_phid=testing_policy_phid,
+        landing_assessment=landing_assessment,
```
zzzeid, line 926: Too many arguments are being passed here that we could be fetching or calculating elsewhere. This should at most be:

```suggestion
        stack_data=stack_data,
        stack=stack,
        statuses=statuses,
        landable_stack=landable_stack,
        reviewers=reviewers,
        users=users,
        projects=projects,
        landing_assessment=landing_assessment,
```
For example, `landable_repos` isn't used in this method, even though it is calculated from two different variables, both which can be used in the definition of `TransplantAssessmentState`.

landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
+    )
+
+    # Where we check the landing blockers.
+    assessment = check_landing_lints(transplant_state)
```
zzzeid, line 930: Referring to things as blockers and lints interchangeably here, I think we should pick a more appropriate term and stick with it (e.g., "landing checks" as suggested in other comments).

landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
-    return DEFAULT_OTHER_BLOCKER_CHECKS + [
-        # Configure uplift check with extra data.
-        check_uplift_approval(relman_group_phid, repositories, stack_data),
-        # Configure data policy check with the project PHID.
-        check_revision_data_classification(data_policy_review_phid),
-    ]
+    landing_assessment: Optional[LandingAssessmentState] = None,
+) -> tuple[TransplantAssessment, TransplantAssessmentState]:
+    """Assess the transplant request."""
+    # Get data from Lando.
```
zzzeid, line 879: This comment seems out of place.

landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
+    users = user_search(phab, involved_phids)
+    projects = project_search(phab, involved_phids)
+
+    secure_project_phid = get_secure_project_phid(phab)
+    testing_tag_project_phids = get_testing_tag_project_phids(phab)
+    testing_policy_phid = get_testing_policy_phid(phab)
+
+    # Create a copy of the stack where we will remove revisions that are blocked from
+    # landing, leaving a graph where each path is landable.
+    landable_stack = copy.deepcopy(stack)
```
zzzeid, line 899: I don't think we should be creating a copy here, that is passed elsewhere for processing. If we need to do something with a copy, the copy should be made then and there (i.e., in `TransplantAssessmentState` in this case.)

The comment is also describing what is happening elsewhere and not here (i.e., "...where we will remove revisions...")

landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
+
+    # We only want to consider paths starting from the open revisions
+    # do grab the status for all revisions.
```
zzzeid, line 902: This comment is unclear. Perhaps this also all belongs in `TransplantAssessmentState` instead. 

```suggestion
    # Grab the status for all revisions, since we only want to consider paths
    # starting from the open revision when assessing.
```

landoapi/transplants.py
```diff
@@ -51,16 +65,115 @@
...
+            to_land=to_land,
+            revision_path=revision_path,
+            revision_to_diff_id=revision_to_diff_id,
+            auth0_user=auth0_user,
+        )
+
+
+@dataclass
+class TransplantAssessmentState:
+    """Handles the state of a Transplant for assessment.
```
zzzeid, line 120: Docstring should describe what is meant by a "state of a Transplant" (also note that capitalizing "Transplant" is a little confusing here, because Transplant used to be a legacy model that is no longer used).

landoapi/transplants.py
```diff
@@ -51,16 +65,115 @@
 CODE_FREEZE_OFFSET = "-0800"
 
 
+@dataclass
+class LandingAssessmentState:
+    """Encapsulates the state of a landing request for assessment.
```
zzzeid, line 70: Docstring should describe what a "state of a landing request" means. Also, perhaps what a "landing request" is and how it is different than a transplant.

I think "landing request" might be a bit confusing because technically a landing job is a landing request (i.e., a request to land something). Is there a better name for this?

landoapi/transplants.py
```diff
@@ -51,16 +65,115 @@
...
+                ],
+            )
+            for revision in to_land
+        ]
+
+        revision_path = []
+        revision_to_diff_id = {}
+        for revision_phid, diff_id in landing_path_phid:
+            revision_path.append(revision_phid)
+            revision_to_diff_id[revision_phid] = diff_id
```
zzzeid, line 108: This code block seems redundant. Why not just pass `landing_path_phid` directly to `LandingAssessmentState`?

landoapi/transplants.py
```diff
@@ -51,16 +65,115 @@
...
+
+    def revision_check_pairs(self) -> list[tuple[dict, dict]]:
+        """Return the appropriate list of `revision, diff` pairs for assessing.
+
+        If this state has an associated `LandingAssessmentState`, return the pairs
+        for the landing path being assessed. Otherwise, return the pairs for every
+        revision in the stack.
+        """
+        if self.landing_assessment:
+            # Only check the revisions in `to_land` for landing assessments.
```
zzzeid, line 152:
```suggestion
            # Only return the revisions in `landing_assessment.to_land` for checking.
```

landoapi/transplants.py
```diff
@@ -51,16 +65,115 @@
...
+            return self.landing_assessment.to_land
+
+        # Check all revisions for full stack assessments.
```
zzzeid, line 155:
```suggestion
        # Return all revisions for checking.
```

landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
+REVISION_BLOCKER_LINTS = [
+    user_block_scm_level,
+    blocker_unsupported_repo,
+    blocker_open_parents,
+    blocker_closed_revisions,
+    blocker_latest_diffs,
+    check_author_planned_changes,
+    check_diff_author_is_known,
+    check_uplift_approval,
+    check_revision_data_classification,
```
zzzeid, line 788: may be good to keep the `blocker_` prefix/pattern for ease of identifying which methods are intended as blockers, or perhaps change everything to say `check_*`.

landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
+    warning_code_freeze,
+    warning_unresolved_comments,
+]
+
+
+def check_landing_lints(
+    transplant_state: TransplantAssessmentState,
+    revision_blockers=REVISION_BLOCKER_LINTS,
+    stack_blockers=STACK_BLOCKER_LINTS,
+    revision_warnings=WARNING_LINTS,
```
zzzeid, line 811: Is it necessary to pass these arguments here? They are defined in this module.

landoapi/transplants.py
```diff
@@ -495,43 +606,330 @@ def check_landing_blockers(
...
+
+    # Run revision-level warning checks.
+    for revision, diff in revision_check_pairs:
+        for check in revision_warnings:
+            reason = check(
+                revision=revision, diff=diff, transplant_state=transplant_state
+            )
+
+            if reason is not None:
+                assessment.warnings.append(reason)
```
zzzeid, line 864: nit (can be applied to above as well).

```suggestion
            if reason := check(
                revision=revision, diff=diff, transplant_state=transplant_state
            ) is not None:
                assessment.warnings.append(reason)
```


Another slightly simpler approach may be:
```python
results = [check(revision, diff, transplant_state) for check in revision_warnings]
assessment.warnings += [result for result in results if result is not None]
```

zzzeid (changes requested):
Nice bug #. This is a first pass and there are some portions of the code that I still need to review.

There are definitely some nice additions in this PR. 

Few quick notes:
- I think that it could use a bit more consistency / improvement when it comes to how differentiable the various concepts are (e.g., transplant vs. landing, etc...)
- I think "lints" in "landing lints" is a bit of a misnomer, since many of the blockers/warnings have nothing to do with the code itself (e.g., stack layout, revision status related, etc...). "Pre-landing checks" might be a better name to encapsulate everything. I.e., having "blocking checks" and "warning checks".
- I think we should try and move away from the deprecated language around "transplants" for new things or things that are being reimplemented, whenever possible.

landoapi/stacks.py
```diff
@@ -135,6 +131,22 @@ def root_revisions(self) -> Iterator[str]:
...
+        """Iterate over the set of root revisions in the stack.
+
+        A leaf revision is a revision in a graph with no successors.
+
+        For example in this stack, where A has no successors:
+        A
+        |\
+        B C
+        | |
+        D E
```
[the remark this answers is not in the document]
cgsheeh: Updated so it is obvious which revision is considered the tip, and made the same change for the docstring of `root_revisions`. I can change the function name from `leaf_revisions` to `tip_revisions` too if you think that would be an improvement.

landoapi/transplants.py
cgsheeh: For the most part, I've been trying to make as few changes to the existing checks as possible. `landable_repos` is used by multiple checks, so I've carried it forward as something that is contained in the `TransplantAssessmentState` so both checks have access to it after it is built once. I'm going to create a custom constructor for the `TransplantAssessmentState` and build shared objects there instead of in this method. At this point the `assess_transplant_request` function is just a constructor for `TransplantAssessmentState` that also calls `check_landing_lints` anyways, so it seems like a reasonable cleanup to take now.

landoapi/transplants.py
cgsheeh: These calls were already a part of `assess_transplant_request`, but they were all happening in the `check_landing_warnings` function call. They've just been moved to an earlier location since we need them to build the `TransplantAssessmentState`. I'm going to leave them as-is for now, because what I would like to do is come up with a unified API for "retrieve and store all the required project PHIDs we need to inspect in Lando", and move them all into that API in a separate PR.

cgsheeh:
Updated a few of the terms used here:

- `TransplantAssessment` is now `StackAssessment`, and `TransplantAssessmentState` is now `StackAssessmentState`.
- All of the blockers/warnings are referred to as `checks`.
- Expanded docstrings to add more detail where applicable.
- Tried to make it clearer that adding the `LandingAssessmentState` to the `StackAssessmentState` will run extra checks to determine if a landing job can be created.

landoapi/transplants.py
```diff
@@ -70,12 +70,13 @@ class LandingAssessmentState:
     """Encapsulates the state of a landing request for assessment.
 
     Holds fields that are necessary to assess if a landing is blocked,
-    but are not necessary to run lints on the entire stack.
+    but are not necessary to run checks on the entire stack. This includes
+    checks for permissions on a user to initiate a landing, whether the
+    requested path can land at the current time, etc.
```
zzzeid, line 75:
```suggestion
    requested path can land at the current time, etc...
```

landoapi/transplants.py
```diff
@@ -70,12 +70,13 @@ class LandingAssessmentState:
     """Encapsulates the state of a landing request for assessment.
```
zzzeid, line 70:
```suggestion
    """Encapsulates the state of a stack for assessment.
```

landoapi/transplants.py
```diff
@@ -70,12 +70,13 @@ class LandingAssessmentState:
...
 
     Holds fields that are necessary to assess if a landing is blocked,
```
zzzeid, line 72:
```suggestion
    Holds fields that are necessary to assess if a landing will be blocked,
```

landoapi/transplants.py
```diff
@@ -70,12 +70,13 @@ class LandingAssessmentState:
...
-    but are not necessary to run lints on the entire stack.
+    but are not necessary to run checks on the entire stack. This includes
+    checks for permissions on a user to initiate a landing, whether the
+    requested path can land at the current time, etc.
     """
 
     auth0_user: A0User
-    revision_path: list[str]
-    revision_to_diff_id: dict[str, int]
+    landing_path_phid: list[tuple[str, int]]
```
zzzeid, line 79: I don't think `landing_path_phid` is a good descriptor. This is a list of tuples, and the variable name implies a single value.

landoapi/transplants.py
```diff
@@ -101,25 +102,21 @@ def from_landing_path(
...
-class TransplantAssessmentState:
-    """Handles the state of a Transplant for assessment.
+class StackAssessmentState:
+    """Handles the state of a stack for assessment.
 
-    Holds the fields necessary to run landing lints.
+    Holds all data relevant to a stack such that the stack blocker/warning checks
+    can be run against them. This includes the optional `LandingAssessmentState`
+    which includes extra fields related to an attempt to create a landing job
+    for patches in the stack.
```
zzzeid, line 119:
```suggestion
    which includes extra fields related to the ability to create a landing job
    for patches in the stack.
```

landoapi/transplants.py
```diff
@@ -141,6 +138,60 @@ class TransplantAssessmentState:
...
+        testing_policy_phid: str,
+        landing_assessment: Optional[LandingAssessmentState] = None,
+    ) -> StackAssessmentState:
+        """Build a `StackAssessmentState` from passed arguments.
+
+        Build any fields that are shared between checks but are derived from
+        existing fields.
+        """
+        # Create a copy of the stack where we will remove revisions that are blocked from
+        # landing, leaving a graph where each path is landable.
```
zzzeid, line 165:
```suggestion
        # Create a copy of the stack so that revisions that are blocked from landing
        # can be removed from it when parsing `StackAssessmentState`.
```

landoapi/transplants.py
```diff
@@ -763,34 +803,34 @@ def blocker_single_landing_repo(
...
     blocker_closed_revisions,
     blocker_latest_diffs,
-    check_author_planned_changes,
-    check_diff_author_is_known,
-    check_uplift_approval,
-    check_revision_data_classification,
+    block_author_planned_changes,
+    block_diff_author_is_known,
+    block_uplift_approval,
+    block_revision_data_classification,
```
zzzeid, line 828: Could be more consistent with the naming of other "blocker" checks (i.e., prefix with `blocker_` instead of `block_`, or rename everything to start with `block_`.

zzzeid (changes requested):
Few more small comments.

landoapi/transplants.py
zzzeid: Is there a bug tracking this work?

landoapi/transplants.py
zzzeid: I do think that it might be good to at least group the "phids" in their own variable, since it is quite tedious to keep passing multiple variables multiple times. Makes sense to punt refactoring this work to remove most (if not all) the dependency injections into its own PR. We should file a bug for that as well.

landoapi/transplants.py
cgsheeh: This is the landing assessment state, not the stack assessment.

landoapi/transplants.py
cgsheeh: This is the name that currently exists in Lando. It's just a landing path but instead of an `int` ID (ie `123` for `D123`) this variable uses the PHID. I'll update it to clarify.

landoapi/transplants.py
cgsheeh: Makes sense to me. :+1:

landoapi/transplants.py
cgsheeh: Updated this but kept the note about paths in this stack being landable.

landoapi/transplants.py
cgsheeh: No, it's just an idea at the moment. The main goal of the refactor was to have a more consistent interface for blockers/warnings than the scattered state we were in before. Cleaning up the scattered calls to Phabricator is something I want to do as well, but it's out of scope for this PR. I can file a bug for this if need be, it's just not a blocker for hg->git and can probably be done when we have some spare cycles.

cgsheeh: Filed [bug 1893498](https://bugzilla.mozilla.org/show_bug.cgi?id=1893498) for this.

zzzeid (approved):
Looks good to me! I think given that we don't have coverage checks running, and there are a few significant changes in this PR, it would be worth doing more thorough manual testing for all the different checks. Glad to see that the code is more consistent and easy to read after this PR.

zzzeid (commented):
Note that the title still says "landing lints" vs. "landing checks".

cgsheeh:
> Looks good to me! I think given that we don't have coverage checks running, and there are a few significant changes in this PR, it would be worth doing more thorough manual testing for all the different checks. Glad to see that the code is more consistent and easy to read after this PR.

Yes I plan to deploy this to dev and run through some tests to make sure it's working as expected. :)

PR #393, opened 2024-04-29
repos: add `large-repo` to dev repo configs (Bug 1821153)

cgsheeh:
Add the existing `large-repo` Phabricator repo to Lando so patches can
be landed there for testing.


zzzeid (approved):
Keep in mind that any commits that will land in `large-repo` (in Lando) will end up propagating to all other repos in phabricator-dev (and Lando) that are pulling from that repo, which might have unintended consequences.

PR #399, opened 2024-05-08
checks: add a check to reject symlinks in Lando (Bug 1895762)

cgsheeh:
This commit adds a blocking check to reject the introduction of
symlinks into the repo. It is a port of the `prevent_symlinks.py`
Mercurial server-side hook. First we re-introduce the `rs-parsepatch`
library as a dependency. We add a `get_parsed_diffs` function to
query Conduit's `differential.getrawdiff` API endpoint to retrieve
the raw diff content for each Phabricator diff, pass the contents
through `rs_parsepatch.get_diffs` and add the response to the
`StackAssessmentState` object. This will allow us to share the
parsed diff contents between this check and future diff-inspecting
checks that will be introduced shortly. We then add a revision-wise
check which inspects the `new` file modes for each file in the diff and
asserts the mode does not match the known symlink decimal notation.


landoapi/transplants.py
```diff
@@ -896,6 +929,21 @@ def run_landing_checks(stack_state: StackAssessmentState) -> StackAssessment:
...
 
 
+def get_parsed_diffs(
+    phab: PhabricatorClient, stack_data: RevisionData
+) -> dict[int, dict]:
+    """Return a mapping of diff PHID to raw `diff --git` content."""
+    raw_diffs = {}
+    for diff in stack_data.diffs.values():
+        diff_id = phab.expect(diff, "id")
+        raw_diff = phab.call_conduit("differential.getrawdiff", diffID=diff_id)
```
zzzeid, line 939: I think it makes sense to cache this on first request. It could significantly slow things down and increase traffic to Phabricator otherwise. It should be easy enough to cache by diff ID since that should be unique. I think as a follow up (perhaps in the new lando, when we will have better access to cron jobs) we should have a cache-warming job that prepopulates the cache.

In the future, we should also cache this on the Phabricator end (maybe with nginx) since I don't think Phabricator caches the `getrawdiff` response either. That might help speed up the first request from Lando in the future.

zzzeid (changes requested):

landoapi/transplants.py
cgsheeh: Makes sense to me, added caching in the next iteration.

zzzeid (approved):
Looks good to me. I do think we should start thinking about a more modular/integrated way of caching things from Phabricator since that's kind of scattered all over the place now (e.g., various project "phids", uplift stuff, etc...)

cgsheeh:
> Looks good to me. I do think we should start thinking about a more modular/integrated way of caching things from Phabricator since that's kind of scattered all over the place now (e.g., various project "phids", uplift stuff, etc...)

Agreed! My first thought is that we could add methods for specific API endpoints to the `PhabricatorClient` instead of calling `call_conduit` in a dynamic fashion in most places. For example this call could become `phab.differential_getrawdiff(diffID=diff_id)`, and then the caching logic all takes place within the Phabricator abstraction.

PR #401, opened 2024-05-09
checks: add a check to reject `try_task_config.json` outside of `try` (Bug 1895931)

cgsheeh:
Add a blocking check that asserts the `try_task_config.json` file is not
introduced erroneously to a non-try repo.

### DIFF
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -842,6 +842,26 @@ def blocker_prevent_symlinks(
         )
 
 
+def blocker_try_task_config(
+    revision: dict, diff: dict, stack_state: StackAssessmentState
+) -> Optional[str]:
+    """Block revisions which contain the `try_task_config.json` file."""
+    if (
+        stack_state.landing_assessment
+        and stack_state.landing_assessment.landing_repo
+        and stack_state.landing_assessment.landing_repo.tree == "try"
+    ):
+        # `try_task_config.json` is allowed to be pushed to try.
+        return
+
+    diff_id = PhabricatorClient.expect(diff, "id")
+    parsed_diff = stack_state.parsed_diffs[diff_id]
+
+    for parsed in parsed_diff:
+        if parsed["filename"] == "try_task_config.json":
+            return "Revision introduces the `try_task_config.json` file into a non-try repo."
+
+
 STACK_BLOCKER_CHECKS = [
     # This check needs to be first.
     blocker_stack_landable,
@@ -861,6 +881,7 @@ REVISION_BLOCKER_CHECKS = [
     blocker_uplift_approval,
     blocker_revision_data_classification,
     blocker_prevent_symlinks,
+    blocker_try_task_config,
     # This check needs to be last.
     blocker_open_ancestor,
 ]
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -876,7 +876,7 @@ def blocker_try_task_config(
 
     for parsed in parsed_diff:
         if parsed["filename"] == "try_task_config.json":
-            return "Revision introduces the `try_task_config.json` file into a non-try repo."
+            return "Revision introduces the `try_task_config.json` file."
 
 
 STACK_BLOCKER_CHECKS = [
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -871,14 +871,6 @@ def blocker_try_task_config(
     revision: dict, diff: dict, stack_state: StackAssessmentState
 ) -> Optional[str]:
     """Block revisions which contain the `try_task_config.json` file."""
-    if (
-        stack_state.landing_assessment
-        and stack_state.landing_assessment.landing_repo
-        and stack_state.landing_assessment.landing_repo.tree == "try"
-    ):
-        # `try_task_config.json` is allowed to be pushed to try.
-        return
-
     diff_id = PhabricatorClient.expect(diff, "id")
     parsed_diff = stack_state.parsed_diffs[diff_id]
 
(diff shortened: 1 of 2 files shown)


tests/test_transplants.py
```diff
@@ -1748,3 +1749,110 @@ def test_blocker_prevent_symlinks(phabdouble):
...
+        attachments={"reviewers": True, "reviewers-extra": True, "projects": True},
+    )
+    diff = phabdouble.diff(revision=revision, rawdiff=TRY_TASK_CONFIG_DIFF)
+
+    phab_client = phabdouble.get_phabricator_client()
+    stack_data = request_extended_revision_data(phab_client, [revision["phid"]])
+
+    # Parse diffs into `rs_parsepatch` format.
+    parsed_diffs = get_parsed_diffs(phab_client, stack_data)
+    stack_state = create_state(stack_data=stack_data, parsed_diffs=parsed_diffs)
```
zzzeid, line 1787: I'm a little concerned that we are circumventing the creation of stack state through the actual implementation and just using the `create_state` method that is defined in this test module (see my broader comment on the PR). It looks like the only way a `StackAssessmentState` is ever created is via the `from_assessment` class method, so why not try and use that for these tests as well? 

If there are optional arguments that aren't needed, even for testing, then they should be optional in `from_assessment`. But if they aren't optional, they can be included via other fixtures (e.g., `relman_project`, `mocked_repo_config`, etc...)

I think ideally this would all be done in `create_state` similar to how you have it, but though the use of fixtures, and by avoiding adding any random values if they aren't actually used for testing. There are fixtures for almost all if not all the things required by `StackAssessmentState.from_assessment`, and those that do not have fixtures are things that can be created using fixtures.

Here's a quick example of what that could look like.

```python
@pytest.fixture
def create_state(phabdouble, mocked_repo_config, ...):
    def _create_state(revision_phid, *kwargs):
        phab = phabdouble.get_phabricator_client()
        stack_data = request_extended_revision_data(phab, [revision["phid"]])
        supported_repos = get_repos_for_env("test")

        # ... leverage other available fixtures here to fetch all required parameters
        return StackAssessmentState.from_assessment(...)
    return _create_state
```

then

```suggestion
    stack_state = create_state(revision["phid"], ...)
```

In summary, these tests exist withing the ecosystem of the other tests, which have various dependencies that are provided by existing fixtures.

I should have caught this problem in the previous PR, but this PR made it more obvious that these tests are not leveraging existing fixtures as much as they should be.

zzzeid (changes requested):
While checking on `create_state` and `create_landing_state`, I noticed that you aren't using any test fixtures (e.g., for phabricator or auth related things), which can cause problems and headaches down the road.

Also, at least in `create_state`, the Phabricator client appears to be redundant and not used, and in the tests (e.g., the one added in this PR), you are manually fetching what could have been fetched by the client elsewhere. This could be cleaned up by changing how `create_state` and `create_landing_state` are implemented.

See more elaborate comment on `test_transplants`.

tests/test_transplants.py
cgsheeh: These tests are just supposed to be unit tests for the checks, so we only really need to populate the fields required to test the check function logic. This is what was done to test the warnings/blockers in the pre-refactor state. For example, previously even though the function `warning_previously_landed` took all the kwargs passed to all the warnings, we only populate the `revision` and `diffs` arguments in `test_warning_previously_landed_*` test functions, since those are the only arguments required to run the warning.

Having a more robust test fixture is a good idea and I'll look into adding it as a follow-up, but given it isn't regressing anything to keep the tests as-is, I don't think it should block this PR.

cgsheeh: Filed [bug 1896501](https://bugzilla.mozilla.org/show_bug.cgi?id=1896501) for this and will have a few PRs up shortly, once those land we can move forward with this review.

cgsheeh:
Converting to draft while I fix the tests to work with the new `create_state` fixture.

landoapi/transplants.py
```diff
@@ -871,14 +871,6 @@ def blocker_try_task_config(
...
 ) -> Optional[str]:
     """Block revisions which contain the `try_task_config.json` file."""
-    if (
-        stack_state.landing_assessment
-        and stack_state.landing_assessment.landing_repo
-        and stack_state.landing_assessment.landing_repo.tree == "try"
-    ):
-        # `try_task_config.json` is allowed to be pushed to try.
-        return
-
```
zzzeid: Is this being removed because we don't trigger landings to try via the UI? Will it be needed when we run checks on API-triggered landing jobs?

cgsheeh: Yes that's correct, these checks will only run on repos which exist on Phabricator. Once I add the API to run checks on API-triggered jobs I will add the check for `try` which skips the check.

landoapi/transplants.py
```diff
@@ -867,6 +867,18 @@ def blocker_prevent_symlinks(
...
+def blocker_try_task_config(
+    revision: dict, diff: dict, stack_state: StackAssessmentState
+) -> Optional[str]:
+    """Block revisions which contain the `try_task_config.json` file."""
+    diff_id = PhabricatorClient.expect(diff, "id")
+    parsed_diff = stack_state.parsed_diffs[diff_id]
+
+    for parsed in parsed_diff:
+        if parsed["filename"] == "try_task_config.json":
+            return "Revision introduces the `try_task_config.json` file into a non-try repo."
```
zzzeid, line 879: I don't think this error string is accurate any more, given that we are not checking the repo anywhere.

```suggestion
            return "Revision introduces the `try_task_config.json` file."
```

zzzeid (changes requested):

zzzeid (approved):

PR #404, opened 2024-05-13
checks: split `assess_stack_state` into a state builder and check runner (Bug 1896501)

cgsheeh:
The `assess_stack_state` function is doing two things at the same time.
Firstly, it is creating the `StackAssessmentState` from some arguments.
Then, it runs the checks against the state and returns a `StackAssessment`.
Split the `assess_stack_state` function into a function that returns a
`StackAssessmentState`, and have callers call `run_landing_checks`
directly. This will enable us to move the `create_state` function into
a proper fixture in a later PR.

### DIFF
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -962,7 +962,7 @@ def get_parsed_diffs(
     }
 
 
-def assess_stack_state(
+def build_stack_assessment_state(
     phab: PhabricatorClient,
     supported_repos: dict[str, Repo],
     stack_data: RevisionData,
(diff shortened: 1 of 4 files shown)


landoapi/transplants.py
```diff
@@ -962,21 +962,16 @@ def get_parsed_diffs(
     }
 
 
-def assess_stack_state(
+def build_stack_assessment_state(
```
cgsheeh, line 965: I realize this could probably be a little cleaner (maybe even a classmethod/constructor on `StackAssessmentState` directly?) but I'd like to defer making it optimal for now, since the actual goal is to make `create_state` an integrated test fixture, and we have already discussed doing some more cleanups after more of the checks/hooks migration work is complete.

cgsheeh: I had a thought that the `StackAssessmentState` could become a `StackAssessor`, and we could turn `run_landing_checks` into a `.run_assessment() -> StackAssessment` method. Just a thought for future improvements, it would make it more obvious that the `run_landing_checks` function does perform mutations on the `StackAssessmentState` object.

zzzeid (approved):
lgtm

PR #407, opened 2024-05-15
hooks: exclude treestatus API endpoints from maintenance mode (Bug 1896696)

cgsheeh:
Add an exception for the Treestatus API endpoints to the
maintenance mode hook. Unlike the dockerflow endpoints, the
Treestatus endpoints are defined with Connexion, and the
derived value of `request.endpoint` ends up being an underscore
separated represenation of the function namespace. Add a check
to exclude any endpoints that are within that namespace.

### DIFF
--- a/landoapi/hooks.py
+++ b/landoapi/hooks.py
@@ -30,11 +30,14 @@ def check_maintenance() -> Optional[Response]:
         "dockerflow.version",
         "dockerflow.lbheartbeat",
     )
+    if request.endpoint in excepted_endpoints:
+        return
+
     treestatus_api_prefix = "landoapi_api_treestatus"
-    if request.endpoint in excepted_endpoints or request.endpoint.startswith(
-        treestatus_api_prefix
-    ):
-        logger.info(f"Skipping maintenance mode check for {request.endpoint} endpoints")
+    if request.endpoint.startswith(treestatus_api_prefix):
+        logger.info(
+            f"Skipping maintenance mode check for Treestatus endpoint {request.endpoint}."
+        )
         return
 
     in_maintenance = ConfigurationVariable.get(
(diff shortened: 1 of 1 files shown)


landoapi/hooks.py
```diff
@@ -30,7 +30,10 @@ def check_maintenance() -> Optional[Response]:
         "dockerflow.version",
         "dockerflow.lbheartbeat",
     )
-    if request.endpoint in excepted_endpoints:
+    treestatus_api_prefix = "landoapi_api_treestatus"
+    if request.endpoint in excepted_endpoints or request.endpoint.startswith(
+        treestatus_api_prefix
+    ):
```
zzzeid, line 36: nit: I think it would be better to be explicit about all the endpoints that we are excluding here. This would help with debugging.

zzzeid (approved):

landoapi/hooks.py
zzzeid: This would basically log all requests to those endpoints all the time. I meant basically listing out all the individual endpoints (similar to how we don't do `request.endpoint.startswith("dockerflow.")`).

This will add unnecessary logs, since the dockerflow endpoints are checked regularly.

landoapi/hooks.py
```diff
@@ -30,7 +30,11 @@ def check_maintenance() -> Optional[Response]:
...
+        logger.info(f"Skipping maintenance mode check for {request.endpoint} endpoints")
```
zzzeid, line 37: This would basically log all requests to those endpoints all the time. I meant basically listing out all the individual endpoints (similar to how we don't do `request.endpoint.startswith("dockerflow.")`).

This will add unnecessary logs, since the dockerflow endpoints are checked regularly.

zzzeid (approved):
As mentioned in my previous comment, I think it would be better to be explicit about the endpoints, rather than check what the string starts with (i.e., list all the endpoints in a list or tuple, similar to what we do with the dockerflow endpoints.) This would also be consistent with how we handle other endpoints.

PR #417, opened 2024-05-29
hooks: add a diff inspection API for landings and try pushes (Bug 1899584)

cgsheeh:
Add a `DiffAssessor` class as an internal API for running
diff-based checks. Move existing diff-inspecting checks
into the API and update landing checks to use them. Add
a `run_diff_checks` method which runs each check and
returns a list of string errors for display on failure.
Hook the `run_diff_checks` API into the try push revision
building function, so the currently ported checks are
run on try pushes. Add integration tests to show that
pushes are accepted/rejected as appropriate based on
these diff checks.

Since some checks will require inspection of the landing
repo to determine if the check is valid, we add a
`get_repo_for_revision` method that returns the repo for
a revision based on the passed `revision` Phabricator API
object. The logic is derived from the uplift blocking
check where it is updated, and is re-used in the
`try_task_config.json` blocker test.

Also, the type hint for `parsed_diffs` was noted to be incorrect,
and has been fixed as a ridealong in this PR.

### DIFF
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -22,6 +22,7 @@ from flask import current_app
 
 from landoapi.auth import A0User
 from landoapi.cache import DEFAULT_CACHE_KEY_TIMEOUT_SECONDS, cache
+from landoapi.hgexports import DiffAssessor
 from landoapi.models.landing_job import LandingJob, LandingJobStatus
 from landoapi.models.revisions import DiffWarning, DiffWarningStatus
 from landoapi.phabricator import (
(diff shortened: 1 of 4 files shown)


landoapi/transplants.py
```diff
@@ -226,6 +227,16 @@ def revision_check_pairs(self) -> list[tuple[dict, dict]]:
...
 
+    def get_repo_for_revision(self, revision: dict) -> Optional[Repo]:
+        """Given a revision object, return the associated `Repo` in Lando."""
+        repo_phid = PhabricatorClient.expect(revision, "fields", "repositoryPHID")
+
+        repo = self.stack_data.repositories.get(repo_phid)
+        if not repo or not self.supported_repos:
+            return None
+
+        return self.supported_repos.get(repo["fields"]["shortName"])
```
zzzeid, line 238: Any reason this isn't `self.supported_repos[repo["fields"]["shortName"]]`? Could fail sooner here if there is something wrong or missing.

zzzeid (changes requested):
Few minor comments.

landoapi/transplants.py
cgsheeh: It's possible to have repos on Phabricator that aren't supported in Lando, and in that case we want to just return `None` and not fail.

zzzeid (approved):

PR #420, opened 2024-06-14
hgexports: add a commit message validation check to `DiffAssessor` (Bug 1902545)

cgsheeh:
Port the commit message validation hook from hgmo into a Lando
`DiffAssessor` check. Most of the code and accompanying regexes
are ported directly from version-control-tools, with a few minor
readability cleanups and removals of some unused features around
vendored directories and deprecated workflow checks. Tests from
v-c-t are ported as well and some additional checks are added as
required.

Not ported here is the `IGNORE BAD COMMIT MESSAGES` override, which
will be added at a later date.

### DIFF
--- a/landoapi/hgexports.py
+++ b/landoapi/hgexports.py
@@ -489,6 +489,8 @@ def parse_bugs(commit_desc: str, conservative: bool = False) -> list[int]:
     bugs_seen = set()
     bugs_seen_add = bugs_seen.add
     bugs = [x for x in bugs_with_duplicates if not (x in bugs_seen or bugs_seen_add(x))]
+
+    # Filter out very large numbers since those are certainly not actual bugs.
     return [bug for bug in bugs if bug < 100000000]
 
 
--- a/landoapi/hgexports.py
+++ b/landoapi/hgexports.py
@@ -20,6 +20,12 @@ from typing import (
     Optional,
 )
 
+from landoapi.commit_message import (
+    ACCEPTABLE_MESSAGE_FORMAT_RES,
+    INVALID_REVIEW_FLAG_RE,
+    is_backout,
+    parse_backouts,
+)
 from landoapi.repos import Repo
 
 HG_HEADER_NAMES = (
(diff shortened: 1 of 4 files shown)


landoapi/hgexports.py
```diff
@@ -393,6 +393,147 @@ def get_timestamp(self) -> str:
...
+    """Returns True if commit description indicates the changeset is a backout.
+
+    Backout commits should always result in is_backout() returning True,
+    and parse_backouts() not returning None.  Malformed backouts may return
+    True here and None from parse_backouts().
+    """
+    return BACKOUT_KEYWORD_RE.match(commit_desc) is not None
+
+
+def parse_bugs(commit_desc: str, conservative: bool = False) -> list[int]:
```
zzzeid, line 470: Missing docstring.

landoapi/hgexports.py
```diff
@@ -393,6 +393,147 @@ def get_timestamp(self) -> str:
         return get_timestamp_from_git_date_header(date)
 
 
+ACCEPTABLE_MESSAGE_FORMAT_RES = [
```
zzzeid, line 396: I think it's a bit odd to have all these definitions (396:456) in the middle of this module. May be better in a different supporting module, or perhaps at the top. Could also define the ones that are ultimately used in the methods, in those methods that are using them, since it doesn't appear that anything is used in more than one method.

landoapi/hgexports.py
```diff
@@ -393,6 +393,147 @@ def get_timestamp(self) -> str:
...
+    if commit_desc.startswith("Bumping gaia.json"):
+        conservative = True
+
+    bugzilla_re = BUG_CONSERVATIVE_RE if conservative else BUG_RE
+
+    bugs_with_duplicates = [int(m[1]) for m in bugzilla_re.findall(commit_desc)]
+    bugs_seen = set()
+    bugs_seen_add = bugs_seen.add
+    bugs = [x for x in bugs_with_duplicates if not (x in bugs_seen or bugs_seen_add(x))]
+    return [bug for bug in bugs if bug < 100000000]
```
zzzeid, line 487: Might be a good opportunity to add a comment describing what 100000000 is referring to here.

landoapi/hgexports.py
```diff
@@ -393,6 +393,147 @@ def get_timestamp(self) -> str:
...
+            if expected != len(nodes):
+                return None
+        return nodes, parse_bugs(commit_desc)
+
+    # Multiple backouts, with nodes listed on the first line
+    m = BACKOUT_MULTI_ONELINE_RE.match(first_line)
+    if m:
+        return SHORT_NODE_RE.findall(m.group("nodes")), parse_bugs(first_line)
+
+    return None
```
zzzeid, line 534: There is a mix of `return` and `return None` in this PR. Would be good to pick one and stick with it, at least for this PR.

zzzeid (changes requested):
Minor comments. Might be good to add a reference to the original code in v-c-t in the PR. I am assuming that functionally there is no difference between the old and ported code.

landoapi/hgexports.py
cgsheeh: Now that I look at these, they make more sense in `commit_message.py`.

PR #423, opened 2024-06-17
hgexports: add a WPTSync bot check to `DiffAssessor` (Bug 1903130)

cgsheeh:
Add a `check_wpt_sync` check to the `DiffAssessor` class which
is a port of the `prevent_wptsync_changes.py` hook in v-c-t.
The hook uses a regex to assert that the WPTSync bot user cannot
make changes to files outside of a specific directory in the main
tree. The bot is allowed to push any changes to `try`, but is
otherwise only allowed to make changes to files matching the regex.


cgsheeh:
This is combined with https://github.com/mozilla-conduit/lando-api/pull/425.

PR #431, opened 2024-07-17
hg: convert errors from `hg id` to `HgException` (Bug 1908492)

cgsheeh:
Since `hg id <repo_url>` hits the network to connect to hgmo,
it can fail due to internal server errors and other connectivity
issues. In this case, we want the landing job to be retried in
the same fashion as when failing to pull/push. Map the exception
raised by `run_hg` to an `HgException` so the failure will be
considered a retry in the landing worker.

### DIFF
--- a/landoapi/hg.py
+++ b/landoapi/hg.py
@@ -641,7 +641,10 @@ class HgRepo:
 
     def get_remote_head(self, source: str) -> bytes:
         # Obtain remote head. We assume there is only a single head.
-        cset = self.run_hg(["identify", source, "-r", "default", "--id"]).strip()
+        try:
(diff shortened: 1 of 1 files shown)


cgsheeh:
> lgtm. Should we also do this for `get_current_node`? The code is repeated in a few different places, I wonder if this should just be in `run_hg` instead, or passed as a flag.

`get_current_node` shouldn't touch the network, so for this specific instance we don't need to wrap it. Adding this to `run_hg` makes sense, but is a slightly more involved change since we would need to update places where we `except hglib.error.CommandError`. I'll look into that when I have some time tomorrow.

PR #432, opened 2024-09-03
checks: add a push-wide check API and implement bug references check (Bug 1915671)

cgsheeh:
Adds an API for push-wide checks and implements the bug references check
using it. The API defines a `PushCheck` interface which initializes the
check with a repo, defines a `relevant()` function which is called to
determine if the check should be run on the given repo, a `next_diff()`
function which accepts each diff parsed as a `PatchHelper`, and a
`result()` function which returns the error message for the check,
or `None` if the check passed.

The `BugReferencesCheck` is the first check to be implemented using this
new API. The check is only relevant on `try` repositories. Each diff is
passed via `next_diff`, and if `SKIP_BMO_CHECK` is found in the commit
message of any diff, a flag is set which indicates the check will be
skipped, otherwise the bug numbers found in the commit message are
stored. On `result()` the `skip_check` flag will cause an unconditional
pass, otherwise BMO is pinged to check the bug numbers in the commit
message are all referencing public bugs.

Since the `author` field was always assumed to be an email address, it was
renamed to the more accurate `email` name.

### DIFF
--- a/landoapi/api/try_push.py
+++ b/landoapi/api/try_push.py
@@ -8,7 +8,6 @@ import enum
 import io
 import logging
 
-import rs_parsepatch
 from connexion import ProblemException
 from flask import (
     current_app,
@@ -17,10 +16,10 @@ from flask import (
 
 from landoapi import auth
 from landoapi.hgexports import (
-    DiffAssessor,
     GitPatchHelper,
     HgPatchHelper,
     PatchHelper,
+    PushAssessor,
 )
 from landoapi.models.landing_job import (
     LandingJobStatus,
@@ -60,14 +59,6 @@ def build_revision_from_patch_helper(helper: PatchHelper, repo: Repo) -> Revisio
 
     raw_diff = helper.get_diff()
 
-    # Check diff for errors.
-    parsed_diff = rs_parsepatch.get_diffs(raw_diff)
-    errors = DiffAssessor(
-        author=author, commit_message=commit_message, parsed_diff=parsed_diff, repo=repo
-    ).run_diff_checks()
-    if errors:
-        raise ValueError(f"Patch failed checks: {' '.join(errors)}")
-
     return Revision.new_from_patch(
         raw_diff=raw_diff,
         patch_data={
@@ -98,7 +89,39 @@ def parse_revisions_from_request(
     """Convert a set of base64 encoded patches to `Revision` objects."""
     patches_io = (io.StringIO(decode_json_patch_to_text(patch)) for patch in patches)
 
-    patch_helpers = (PATCH_HELPER_MAPPING[patch_format](patch) for patch in patches_io)
+    try:
+        patch_helpers = [
+            PATCH_HELPER_MAPPING[patch_format](patch) for patch in patches_io
+        ]
+    except ValueError as exc:
+        raise ProblemException(
+            400,
+            "Improper patch format.",
+            f"Patch does not match expected format `{patch_format.value}`: {str(exc)}",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
+
+    try:
+        errors = PushAssessor(patch_helpers=patch_helpers, repo=repo).run_push_checks()
+    except ValueError as exc:
+        raise ProblemException(
+            400,
+            "Improper patch format.",
+            f"Patch does not match expected format `{patch_format.value}`: {str(exc)}",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
+
+    if errors:
+        error_message = (
+            f"Patch does not match expected format `{patch_format.value}`: "
+            f"Patch failed checks: {' '.join(errors)}"
+        )
+        raise ProblemException(
+            400,
+            "Improper patch format.",
+            error_message,
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
 
     try:
         return [
--- a/landoapi/api/try_push.py
+++ b/landoapi/api/try_push.py
@@ -19,8 +19,8 @@ from landoapi.hgexports import (
     BugReferencesCheck,
     GitPatchHelper,
     HgPatchHelper,
+    PatchCollectionAssessor,
     PatchHelper,
-    PushAssessor,
 )
 from landoapi.models.landing_job import (
     LandingJobStatus,
@@ -103,9 +103,9 @@ def parse_revisions_from_request(
         )
 
     try:
-        errors = PushAssessor(patch_helpers=patch_helpers, repo=repo).run_push_checks(
-            push_checks=[BugReferencesCheck]
-        )
+        errors = PatchCollectionAssessor(
+            patch_helpers=patch_helpers, repo=repo
+        ).run_patch_collection_checks(push_checks=[BugReferencesCheck])
     except ValueError as exc:
         raise ProblemException(
             400,
--- a/landoapi/api/try_push.py
+++ b/landoapi/api/try_push.py
@@ -16,6 +16,7 @@ from flask import (
 
 from landoapi import auth
 from landoapi.hgexports import (
+    BugReferencesCheck,
     GitPatchHelper,
     HgPatchHelper,
     PatchHelper,
@@ -102,7 +103,9 @@ def parse_revisions_from_request(
         )
 
     try:
-        errors = PushAssessor(patch_helpers=patch_helpers, repo=repo).run_push_checks()
+        errors = PushAssessor(patch_helpers=patch_helpers, repo=repo).run_push_checks(
+            push_checks=[BugReferencesCheck]
+        )
     except ValueError as exc:
         raise ProblemException(
             400,
--- a/landoapi/api/try_push.py
+++ b/landoapi/api/try_push.py
@@ -109,19 +109,17 @@ def parse_revisions_from_request(
     except ValueError as exc:
         raise ProblemException(
             400,
-            "Improper patch format.",
-            f"Patch does not match expected format `{patch_format.value}`: {str(exc)}",
+            "Error running checks on patch collection.",
+            f"Error running checks on patch collection: {str(exc)}",
             type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
         )
 
     if errors:
-        error_message = (
-            f"Patch does not match expected format `{patch_format.value}`: "
-            f"Patch failed checks: {' '.join(errors)}"
-        )
+        bulleted_errors = "\n  - ".join(errors)
+        error_message = f"Patch failed checks:\n\n  - {bulleted_errors}"
         raise ProblemException(
             400,
-            "Improper patch format.",
+            "Errors found in pre-submission patch checks.",
             error_message,
             type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
         )
--- a/landoapi/bmo.py
+++ b/landoapi/bmo.py
@@ -69,7 +69,7 @@ def get_status_code_for_bug(bug_id: int) -> int:
 
 
 def uplift_get_bug(params: dict) -> dict:
-    """Retrieve bug information from the BMO REST API endpoint."""
+    """Retrieve bug information from the Lando Uplift Automation endpoint."""
     resp_get = api_request("GET", "lando/uplift", authenticated=True, params=params)
     resp_get.raise_for_status()
 
@@ -77,7 +77,7 @@ def uplift_get_bug(params: dict) -> dict:
 
 
 def uplift_update_bug(json: dict) -> requests.Response:
-    """Update a BMO bug."""
+    """Update a BMO bug via the Lando Uplift Automation endpoint."""
     if "ids" not in json or not json["ids"]:
         raise ValueError("Need bug values to be able to update!")
 
--- a/landoapi/bmo.py
+++ b/landoapi/bmo.py
@@ -3,24 +3,41 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import requests
+from typing import Optional
 from flask import current_app
 
 
-def bmo_search_bug_endpoint() -> str:
-    """Returns the BMO bug search endpoint URL."""
-    return f"{current_app.config['BUGZILLA_URL']}/rest/bug"
+def bmo_default_headers() -> dict[str, str]:
+    """Returns a `dict` containing the default REST API headers."""
+    return {
+        "User-Agent": "Lando-API",
+        "X-Bugzilla-API-Key": current_app.config["BUGZILLA_API_KEY"],
+    }
+
+
+def bugzilla(
+    method: str, path: str, *args, headers: Optional[dict] = None, **kwargs
+) -> requests.Response:
+    """Send an HTTP `GET` request to BMO."""
+    url = f"{current_app.config['BUGZILLA_URL']}/rest/{path}"
+
+    common_headers = bmo_default_headers()
+    if headers:
+        common_headers.update(headers)
+
+    return requests.request(method, url, *args, headers=headers, **kwargs)
 
 
 def search_bugs(bug_ids: set[int]) -> set[int]:
     """Search for bugs with given IDs on BMO."""
-    bug_search_endpoint = bmo_search_bug_endpoint()
     params = {
         "id": ",".join(str(bug) for bug in sorted(bug_ids)),
         "include_fields": "id",
     }
 
-    resp = requests.get(
-        bug_search_endpoint,
+    resp = bugzilla(
+        "GET",
+        "bug",
         headers=bmo_default_headers(),
         params=params,
     )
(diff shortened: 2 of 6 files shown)


landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+    skip_check: bool = False
+
+    def relevant(self) -> bool:
+        if self.repo.tree == "try":
+            return True
+
+        return False
+
+    def next_diff(self, patch_helper: PatchHelper):
+        commit_message = patch_helper.get_commit_description()
```
zzzeid, line 654: Missing doc string (parent class doesn't really explain what is happening here either as nothing is being "run").

I'm not sure this method is doing what it's intended to do? Or maybe it just has the wrong name. What is being done in this method may be better done when initializing this check instance.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+
+        # Check a single bug only.
+        bug_id = invalid_bugs.pop()
+        try:
+            status_code = get_status_code_for_bug(bug_id)
+        except requests.exceptions.RequestException as exc:
+            return (
+                "Could not contact BMO to check for security bugs referenced in commit message. "
+                f"Use `SKIP_BMO_CHECK` in your commit message to push anyways. Error: {str(exc)}."
+            )
```
zzzeid, line 687: Repeated string.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+    def result(self) -> Optional[str]:
+        if self.skip_check or not self.bug_ids:
+            return
+
+        try:
+            found_bugs = search_bugs(self.bug_ids)
+        except requests.exceptions.RequestException as exc:
+            return (
+                "Could not contact BMO to check for security bugs referenced in commit message. "
+                f"Use `SKIP_BMO_CHECK` in your commit message to push anyways. Error: {str(exc)}."
```
zzzeid, line 672:
```suggestion
                f"Use `SKIP_BMO_CHECK` in your commit message to push anyway. Error: {str(exc)}."
```

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+        checks = [
+            check(repo=self.repo)
+            for check in push_checks
+            if check(repo=self.repo).relevant()
+        ]
+
+        for patch_helper in self.patch_helpers:
+            # Pass the patch information into the push-wide check.
+            for check in checks:
+                check.next_diff(patch_helper)
```
zzzeid, line 737: This is a little confusing here. If the purpose of this call is to set `check.skip_check` and `check.bug_ids`, maybe there's a clearer way of doing that elsewhere more explicitly.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+        commit_message = patch_helper.get_commit_description()
+
+        # Skip the check if the `skip_check` flag is set.
+        if self.skip_check or "SKIP_BMO_CHECK" in commit_message:
+            self.skip_check = True
+            return
+
+        self.bug_ids |= set(parse_bugs(commit_message))
+
+    def result(self) -> Optional[str]:
```
zzzeid, line 663: Missing doc string. Parent class docstring also doesn't cover this method well.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+        if status_code == 404:
+            return (
+                f"Your commit message references bug {bug_id}, which does not exist. Please check your "
+                "commit message and try again."
+            )
+
+        return (
+            f"While checking if bug {bug_id} in your commit message is a security bug, an error occurred "
+            "and the bug could not be verified."
+        )
```
zzzeid, line 705: Would be cleaner to put all the error strings into a dictionary and pass an "error code" to fetch the right string here, instead of scattering the strings across this method. Error codes could look like `ERROR_BUGZILLA_UNAVAILABLE`, or `ERROR_BUG_DOES_NOT_EXIST`.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+    ) -> list[str]:
+        """Execute the set of checks on the diffs, returning a list of issues.
+
+        `push_checks` specifies the push-wide checks to run on the push, otherwise
+        all checks will be run.
+        """
+        issues = []
+
+        if not push_checks:
+            push_checks = [BugReferencesCheck]
```
zzzeid, line 726: Not sure if this makes sense here. we could just call `PushAssessor` explicitly with the relevant checks in `api_try_push`. Maybe there could be a validation section that checks that an irrelevant check is not accidentally added to a repo.

landoapi/api/try_push.py
```diff
@@ -98,7 +89,39 @@ def parse_revisions_from_request(
...
+    except ValueError as exc:
+        raise ProblemException(
+            400,
+            "Improper patch format.",
+            f"Patch does not match expected format `{patch_format.value}`: {str(exc)}",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400",
+        )
+
+    try:
+        errors = PushAssessor(patch_helpers=patch_helpers, repo=repo).run_push_checks()
```
zzzeid, line 105: I think it makes more sense to initialize the `PushAssessor` object with the relevant checks, then call `run_push_checks`. This would also remove the need for the `check.relevant` property, unless there are more specific checks of relevancy in the future (though the docstring you have implies that repo-relevancy is the only thing being checked.)

```python
assessor = PushAssessor(patch_helpers=patch_helpers, repo=repo, checks=[BugReferencesCheck])
errors = assessor.run_push_checks()
```

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+                "Could not contact BMO to check for security bugs referenced in commit message. "
+                f"Use `SKIP_BMO_CHECK` in your commit message to push anyways. Error: {str(exc)}."
+            )
+
+        invalid_bugs = self.bug_ids - found_bugs
+        if not invalid_bugs:
+            return
+
+        # Check a single bug only.
+        bug_id = invalid_bugs.pop()
```
zzzeid, line 680: Can you add a comment as to why we are ignoring the rest of the bugs?

landoapi/api/try_push.py
```diff
@@ -60,14 +59,6 @@ def build_revision_from_patch_helper(helper: PatchHelper, repo: Repo) -> Revisio
...
     raw_diff = helper.get_diff()
 
-    # Check diff for errors.
-    parsed_diff = rs_parsepatch.get_diffs(raw_diff)
-    errors = DiffAssessor(
-        author=author, commit_message=commit_message, parsed_diff=parsed_diff, repo=repo
-    ).run_diff_checks()
-    if errors:
-        raise ValueError(f"Patch failed checks: {' '.join(errors)}")
-
```
zzzeid, line 70: It looks like this was moved to be inside `PushAssessor`. I think it would be cleaner if the two were kept separate, and their results combined after each one is called independently. This also changes the logic of when/how diff errors are raised. This might be OK but it's not explicitly called out or explained in this PR.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+
+@dataclass
+class PushCheck:
+    """Provides an interface to implement push-wide checks."""
+
+    repo: Repo
+
+    def relevant(self) -> bool:
+        """Return `True` if the check is relevant for the given repo."""
+        raise NotImplementedError()
```
zzzeid, line 629: See other comments in this PR regarding why this may not be needed.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+
+    def next_diff(self, patch_helper: PatchHelper):
+        """Pass the next diff to be checked into the check and run."""
```
zzzeid, line 632: It isn't clear what is meant by this docstring.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+        raise NotImplementedError()
+
+    def result(self) -> Optional[str]:
+        """Return the result of the check."""
```
zzzeid, line 636:
```suggestion
        """Calculate the result of the check and return it."""
```

tests/test_hgexports.py
```diff
@@ -604,7 +608,7 @@ def test_check_wpt_sync_irrelevant_user(mocked_repo_config):
         GIT_DIFF_FILENAME_TEMPLATE.format(filename="somefile.txt")
     )
     diff_assessor = DiffAssessor(
-        author="sheehan@mozilla.com",
+        email="sheehan@mozilla.com",
```
zzzeid, line 611: What was this change for? Maybe just add something to the commit message.

zzzeid (changes requested):
First pass

landoapi/bmo.py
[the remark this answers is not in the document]
cgsheeh: Please add more detail about the disadvantages you see with the current implementation. There's not enough information in this comment, or the comment below, for me to understand what sort of technical debt is accumulating.

As a counterpoint, the current implementation benefits from being functional, stateless, and straightforward. It is very easy to test since the methods can be overwritten with a monkeypatch. Rewriting it as a class might introduce unnecessary complexity without providing clear benefits, and make future changes to the API more difficult to implement due to abstractions that are not strictly necessary, on top of requiring more time investment from myself to re-implement.

landoapi/bmo.py
```diff
@@ -6,6 +6,43 @@
 from flask import current_app
 
 
+def bmo_search_bug_endpoint() -> str:
+    """Returns the BMO bug search endpoint URL."""
+    return f"{current_app.config['BUGZILLA_URL']}/rest/bug"
```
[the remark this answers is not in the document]
cgsheeh: The reason we have a function for the URLs is because `current_app.config['BUGZILLA_URL']` may not exist at module import time, but will exist at function call time.

The current use of the API code is already mostly the same as your example, except we don't need to initialize any objects since it is a stateless operation:

```python
from landoapi.bmo import search_bugs
found_bugs = search_bugs(...)
```

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
                 issues.append(issue)
 
         return issues
+
+
+@dataclass
+class PushCheck:
+    """Provides an interface to implement push-wide checks."""
```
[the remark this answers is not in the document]
cgsheeh: We do already have a concept of "push-to-Lando", which will expand as Git support is added since try/automation/sheriffs will all need to submit sets of patches in one go, which write to the pushlog as a single entry. Most of our users are familiar with the idea of a push being a set of commits/changesets that are sent to a repo as a single unit, so I fear introducing a separate internal naming for this concept might be a bit confusing. If we do want to avoid using `Push...`, I would vote for `PatchCollectionCheck`, since we already have a `StackAssessor` which does something different from this API.

landoapi/bmo.py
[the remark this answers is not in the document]
zzzeid: The main issue is that there is repetition of code -- e.g., passing headers each time, implementing a new method for each path URL (really just as a workaround), calling `.json` on each response, handling any communication errors, etc., all which could be handled by the wrapper itself (i.e., a `.get` method).

Common handling of requests/responses should be consolidated into common methods (they don't _have_ to be in a class, you can do something similar with just methods). A class is just one conventional (and convenient) way to do that. We already do this for Phabricator in this repo (see `PhabricatorClient`), as well as for Lando API in the Lando UI repo), among other places. So this is really just following those conventions.

> It is very easy to test since the methods can be overwritten with a monkeypatch

The same can be done with a wrapper. The issue isn't necessarily having the custom methods that you currently have (e.g., `search_bugs`), it's just that there is too much being done in those methods. If you want to keep your tests mostly the same, you could do this:

```python
def search_bugs(bug_ids: set[int]) -> set[int]:
    """Search for bugs with given IDs on BMO."""
    bug_search_endpoint = "/bugs"
    params = {
        "id": ",".join(str(bug) for bug in sorted(bug_ids)),
        "include_fields": "id",
    }

    result = bmo.get(bug_search_endpoint, params=params)
    bugs = result["bugs"]
    return {int(bug["id"]) for bug in bugs}
```

> on top of requiring more time investment from myself to re-implement

That's the cost/benefit of implementing things one way or another from the get go. My comment in the original PR was largely in anticipation for this situation, however as mentioned already the ask is not for a major reimplementation. It's mainly just to consolidate the common API functionality into one place since we're already repeating it in different places and it is somewhat breaking existing conventions.

If you don't want to have a class (which is also fine), you could still implement these improvements in a method. It's just that we already use the class approach elsewhere.

```python
def call_bugzilla(path, method="get", params=None, data=None):
    headers = {...}
    base_url = ...
    if method == "get":
        try:
            result = getattr(requests, method)(f"{base_url}/{path}", params, data).json()
        except ...:
            ...
    return result.json() 
```

landoapi/bmo.py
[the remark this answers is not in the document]
zzzeid: I responded to this in the main comment on this file. But to reiterate, the objective would be mainly to remove code that is unrelated to the business logic from the `search_bug` and other similar methods (e.g., `get_status_code_for_bug`), and remove code repetition. I.e., implementing only the high level functionality in those methods.

landoapi/hgexports.py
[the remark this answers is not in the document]
zzzeid: I am OK with `PatchCollectionCheck`, makes sense to avoid `StackCheck` since it is too similar to `StackAssessor`. I guess we would also rename `PushAssessor` to `PatchCollectionAssessor`.

landoapi/hgexports.py
cgsheeh: This is where all the push checks are passed each individual diff. Setting `skip_check` and `bug_ids` will only happen for the `BugReferencesCheck`, and other push-wide checks will have different behaviour.

landoapi/bmo.py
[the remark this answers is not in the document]
cgsheeh: Thanks for the clarification. I've removed the repeated code issues and it is much cleaner now, this will be included in the next iteration.

landoapi/hgexports.py
```diff
@@ -605,3 +616,144 @@ def run_diff_checks(self) -> list[str]:
...
+    """Prevent commit messages referencing non-public bugs from try."""
+
+    bug_ids: set[int] = field(default_factory=set)
+    skip_check: bool = False
+
+    def relevant(self) -> bool:
+        if self.repo.tree == "try":
+            return True
+
+        return False
```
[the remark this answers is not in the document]
cgsheeh: I think putting the checks in `Repo` makes sense. We could enable/disable checks in the admin interface without needing a code change. I'm going to do that part in a follow-up because the most logical approach will require some changes to `DiffAssessor` and I want to keep this PR focused on finishing the last 2 checks.

landoapi/api/try_push.py
cgsheeh: Since the push assessor is iterating over each diff and performing checks, and the `DiffAssessor` is essentally doing the same thing, I think it makes sense to run the `DiffAssessor` in the same loop for now. They both end up returning error strings anyways.

landoapi/hgexports.py
cgsheeh: What would make that approach cleaner? I'd rather not move the error messages away from the code that is triggering them. Also it seems like an unnecessary abstraction because the messages are only used in 1-2 locations within a single function.

tests/test_hgexports.py
cgsheeh: Added extra context to commit message.

landoapi/bmo.py
```diff
@@ -2,26 +2,54 @@
...
 
 
-def bmo_search_bug_endpoint() -> str:
-    """Returns the BMO bug search endpoint URL."""
-    return f"{current_app.config['BUGZILLA_URL']}/rest/bug"
+def api_request(
+    method: str,
+    path: str,
+    *args,
+    authenticated: bool = False,
```
zzzeid, line 15: nit: this could be renamed to be more obvious, e.g. "use_api_key".

landoapi/hgexports.py
```diff
@@ -620,37 +620,42 @@ def run_diff_checks(self) -> list[str]:
 
 @dataclass
 class PushCheck:
```
zzzeid, line 622: I guess you didn't end up renaming this? Your comment seemed to imply that you were going to do that.

zzzeid (approved):
Looks good to me. I think the implementation of `api_request` deviates from our convention of putting this sort of thing in a class, and the introduction of the method parameter is less robust than having equivalent class methods, but it is still better than what was there before. I think in the future we should consider implementing this in a class, like we do for PhabricatorClient.

It would also be good to add some additional human-readable documentation outlining how to use this new checks API? (not as part of this PR necessarily). That's something I remember finding challenging with the previous checks when I first started using them.

landoapi/hgexports.py
cgsheeh: My mistake, it's renamed now.

ahochheiden (approved):
LGTM.

PR #435, opened 2024-10-15
autoformat: remove `eslint` from supported formatters list (Bug 1924891)

cgsheeh:
`eslint` can make more than just formatting fixes, so we should disable it.


shtrom (approved):
LGTM.

PR #438, opened 2024-10-22
checks: add a warning when a revision has diffs from multiple authors

cgsheeh:
Add a warning that checks when a revision has diffs from multiple
authors associated with it. First we add support to specify a diffs
author in the test suite. Then we update `request_extended_revision_data`
to pull diffs based on the `revisionPHID` instead of the diff `id`,
which returns all diffs for a given revision. We update
`gather_involved_phids` to take the list of diffs for a revision and
gather all diff authors as involved in the revision. We then add
the warning which gathers all the PHIDs that have authored diffs
on a revision, and emit a warning when there are multiple authors,
including the Phabricator username of each author.

### DIFF
--- a/landoapi/revisions.py
+++ b/landoapi/revisions.py
@@ -25,7 +25,7 @@ from landoapi.secapproval import (
 logger = logging.getLogger(__name__)
 
 
-def gather_involved_phids(revision: dict, diffs: dict[str, dict]) -> set[str]:
+def gather_involved_phids(revision: dict, revision_diffs: list[dict]) -> set[str]:
     """Return the set of Phobject phids involved in a revision.
 
     At the time of writing Users and Projects are the type of Phobjects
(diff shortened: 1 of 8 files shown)


landoapi/revisions.py
```diff
@@ -25,7 +25,7 @@
 logger = logging.getLogger(__name__)
 
 
-def gather_involved_phids(revision: dict) -> set[str]:
+def gather_involved_phids(revision: dict, diffs: dict[str, dict]) -> set[str]:
     """Return the set of Phobject phids involved in a revision.
```
zzzeid, line 29: I think this docstring needs to be updated to reflect the new `diff` parameter. The current one was scoped for a particular "revision" dict.

landoapi/revisions.py
```diff
@@ -34,6 +34,19 @@ def gather_involved_phids(revision: dict) -> set[str]:
     attachments = PhabricatorClient.expect(revision, "attachments")
 
     entities = {PhabricatorClient.expect(revision, "fields", "authorPHID")}
+    revision_diffs = [
+        diff
+        for diff in diffs.values()
+        if PhabricatorClient.expect(revision, "phid")
+        == PhabricatorClient.expect(diff, "fields", "revisionPHID")
```
zzzeid, line 41: I think we should be only passing the relevant diffs to this method, rather than filtering out irrelevant diffs here.

zzzeid (changes requested):

zzzeid (approved):

PR #439, opened 2024-10-22
commit_message: change `parse_bugs` to use `BUG_CONSERVATIVE_RE` (Bug 1926431)

cgsheeh:
Change `parse_bugs` to use `BUG_CONSERVATIVE_RE` to avoid incorrect parsing
of bug numbers during the `BugReferencesCheck`. Other code paths use the
`parse_bugs` function as well, but their tests pass after this change, so
we fully remove `BUG_RE` since it is no longer used.

Also remove the accidental double-definition of `BUG_RE` and `BUG_CONSERVATIVE_RE`
from `commit_message.py`.

### DIFF
--- a/landoapi/commit_message.py
+++ b/landoapi/commit_message.py
@@ -12,7 +12,7 @@ REVISION_URL_TEMPLATE = "Differential Revision: {url}"
 
 # Like BUG_RE except it doesn't flag sequences of numbers, only positive
 # "bug" syntax like "bug X" or "b=".
-BUG_CONSERVATIVE_RE = re.compile(r"""((?:bug|b=)(?:\s*)(\d+)(?=\b))""", re.I | re.X)
+BUG_CONSERVATIVE_RE = re.compile(r"""(\b(?:bug|b=)\b(?:\s*)(\d+)(?=\b))""", re.I | re.X)
 
 SPECIFIER = r"\b(?:r|a|sr|rs|ui-r)[=?]"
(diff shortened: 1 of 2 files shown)


landoapi/commit_message.py
```diff
@@ -109,25 +89,6 @@
 )
 RE_SOURCE_REPO = re.compile(r"^Source-Repo: (https?:\/\/.*)$", re.MULTILINE)
 RE_SOURCE_REVISION = re.compile(r"^Source-Revision: (.*)$", re.MULTILINE)
-# Like BUG_RE except it doesn't flag sequences of numbers, only positive
-# "bug" syntax like "bug X" or "b=".
-BUG_CONSERVATIVE_RE = re.compile(r"""(\b(?:bug|b=)\b(?:\s*)(\d+)(?=\b))""", re.I | re.X)
```
zzzeid, line 114: This string is slightly different than the one on line 35, and this is the line that was prior to this change would have overridden the original value. Is there any functional difference between the two? Just wanted to flag this in case there is.

zzzeid (changes requested):

landoapi/commit_message.py
cgsheeh: Looks like this version has an extra fix from your patch in [D69608](https://phabricator.services.mozilla.com/D69608). I will switch this to removing the other copy of `BUG_CONSERVATIVE_RE`. :)

zzzeid (approved):

PR #440, opened 2024-10-25
mocks: respect `limit` parameter in mock Conduit API endpoints (Bug 1927013)

cgsheeh:
The `PhabricatorDouble` mock API calls take the `limit` parameter,
but do not actually limit the number of entries in the response data.
Add a check if `limit` is set and slice the `items `list at that index.

### DIFF
--- a/tests/mocks.py
+++ b/tests/mocks.py
@@ -836,6 +836,9 @@ class PhabricatorDouble:
 
             items = [i for i in items if i["slug"] in constraints["slugs"]]
 
+        if limit:
+            items = items[:limit]
+
         return {
             "data": [to_response(i) for i in items],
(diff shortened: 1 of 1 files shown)


shtrom (approved):
LGTM in isolation.

Should the mock also honour `before` and `after`, e.g., something like 

```items = items[after:min(before,after+limit)]```

cgsheeh:
> LGTM in isolation.
> 
> Should the mock also honour `before` and `after`, e.g., something like
> 
> `items = items[after:min(before,after+limit)]`

Yes, we should probably support the cursor methods as well. It's slightly more involved since the `before` and `after` values are `None` by default, so we can do that in a separate PR. :+1:

PR #442, opened 2024-11-20
checks: refactor `DiffAssessor` check to more closely resemble `PatchCollectionCheck` (Bug 1932507)

cgsheeh:
This commit refactors the `DiffAssessor` API to more closely
match the `PatchCollectionCheck` API. A `PatchCheck` abstract
class is introduced which has a similar interface as the
`PatchCollectionCheck`, where each check can store state
as each part of the patch is iterated over using `next_diff`,
and `result()` is called to determine the final result of
the check. Each check in the `DiffAssessor` class is moved
to a class which subclasses `PatchCheck`.

To make the patch checks more closely resemble the patch
collection checks, each check to be run is passed as an
argument to `run_patch_collection_checks`, which allows
us to run the relevant checks based on the code path
or known relevant repository instead of unconditionally
running all checks. As a result, all logic around inspecting
the repo and determining if a check is relevant is removed
from the checks, simplifying the APIs.

### DIFF
--- a/landoapi/transplants.py
+++ b/landoapi/transplants.py
@@ -22,7 +22,10 @@ from flask import current_app
 
 from landoapi.auth import A0User
 from landoapi.cache import DEFAULT_CACHE_KEY_TIMEOUT_SECONDS, cache
-from landoapi.hgexports import DiffAssessor
+from landoapi.hgexports import (
+    PreventSymlinksCheck,
+    TryTaskConfigCheck,
+)
 from landoapi.models.landing_job import LandingJob, LandingJobStatus
 from landoapi.models.revisions import DiffWarning, DiffWarningStatus
 from landoapi.phabricator import (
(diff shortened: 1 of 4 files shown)


landoapi/transplants.py
```diff
@@ -878,7 +881,11 @@ def blocker_prevent_symlinks(
     diff_id = PhabricatorClient.expect(diff, "id")
     parsed_diff = stack_state.parsed_diffs[diff_id]
 
-    return DiffAssessor(parsed_diff=parsed_diff).check_prevent_symlinks()
+    symlink_check = PreventSymlinksCheck()
+    for diff in parsed_diff:
+        symlink_check.next_diff(diff)
+
+    return symlink_check.result()
```
shtrom, line 888: I think we should keep using the DiffAssessor here, even if we are making it run a single check. This allows us to keep control of the `for` loop and the `return` in a single location, rather than duplicate it.

landoapi/transplants.py
```diff
@@ -888,11 +895,11 @@ def blocker_try_task_config(
...
 
-    local_repo = stack_state.get_repo_for_revision(revision)
+    try_task_config_check = TryTaskConfigCheck()
+    for diff in parsed_diff:
+        try_task_config_check.next_diff(diff)
 
-    return DiffAssessor(
-        parsed_diff=parsed_diff, repo=local_repo
-    ).check_try_task_config()
+    return try_task_config_check.result()
```
shtrom, line 902: Use DiffAssessor

shtrom (changes requested):
I think we should continue using the DiffAssessor in the transplants, even if running a single check.

shtrom (approved):
lgtm 👍🏼

PR #443, opened 2024-11-28
uplift: skip Bugzilla updates for landings without bugs (Bug 1934003)

cgsheeh:
Currently when an uplift without a bug number hits the
Bugzilla update code path, an exception is thrown which
is turned into a failure email indicating Bugzilla could
not be updated. This commit changes the behaviour to
avoid trying to update Bugzilla if no bug numbers are
detected.


zzzeid (approved):

zzzeid:
Would be good to add tests for these in the future.

PR #446, opened 2024-12-10
landing-worker: add a try/except block around `job.set_landed_reviewers` calls (Bug 1936373)

cgsheeh:
Catch any exceptions thrown while setting reviewers from `mots.yaml`, so
they do not take down the landing worker and cause lag in the queue.


dklawren (approved):
LGTM

PR #454, opened 2025-01-29
hgexports: fix CRLF handling in GitPatchHelper (bug 1944462)

glandium:
HgPatchHelper doesn't have this problem because it uses io.StringIO,
which yields the line endings.

Some tests need to be accommodated to the fact that the last line of a
diff now does contain newlines, which was accidentally not the case
before, and while doing that, adjust a little more test strings to be
more fitting with the reality that a patch will not terminate without
a newline at the end.

### DIFF
--- a/landoapi/hgexports.py
+++ b/landoapi/hgexports.py
@@ -349,15 +349,13 @@ class GitPatchHelper(PatchHelper):
             raise ValueError("No valid subject header for commit message.")
 
         # Start the commit message from the stripped subject line.
-        commit_message_lines = [
-            subject_header.removeprefix("[PATCH] ").removesuffix("\n")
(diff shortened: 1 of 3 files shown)


cgsheeh (changes requested):
Looks great. Could you add a test case for this to `test_hgexports.py`? Something similar to `test_git_formatpatch_helper_utf8` showing the diff is parsed with correct newlines should suffice.

zzzeid (changes requested):
Wondering if you could add a comment in the commit message about why the tests were altered? Otherwise looks good.

cgsheeh (approved):

zzzeid (approved):

PR #459, opened 2025-03-14
api: Allow fetching landed commit_id of a landing job (bug 1954116)

moztcampbell:
When tools push to try, they only get the landing job id in return, while they often want the commit_id that TreeHerder uses. If the job isn't landed yet, a simple status is returned instead. Landed commit_id or job status isn't sensitive so an auth check isn't required.

### DIFF
--- a/landoapi/api/landing_jobs.py
+++ b/landoapi/api/landing_jobs.py
@@ -13,6 +13,33 @@ from landoapi.storage import db
 logger = logging.getLogger(__name__)
 
 
+def get(landing_job_id: str):
+    """Return status about a landing job
+
+    Args:
+        landing_job_id (str): The unique ID of the LandingJob object.
+
+    Raises:
+        ProblemException: If a LandingJob object corresponding to the landing_job_id
+            is not found.
+    """
+    landing_job = LandingJob.query.get(landing_job_id)
+
+    if not landing_job:
+        raise ProblemException(
+            404,
+            "Landing job not found",
+            f"A landing job with ID {landing_job_id} was not found.",
+            type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404",
+        )
+
+    return {
+        "id": landing_job.id,
+        "status": landing_job.status.value,
+        "commit_id": landing_job.landed_commit_id,
+    }
+
+
 @auth.require_auth0(scopes=("lando", "profile", "email"), userinfo=True)
 def put(landing_job_id: str, data: dict):
     """Update a landing job.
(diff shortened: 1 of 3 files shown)


moztcampbell:
@zzzeid , I haven't set up a local test environment for this, but this is what I was imagining for a simple api for getting commit id. Then I can local try pushes generate a treeherder link that uses this API to resolve.

shtrom:
@moztcampbell it might be worth working against https://github.com/mozilla-conduit/lando now, which is the new version of Lando we're going to switch to soon as part of the migration to Git.

moztcampbell:
> @moztcampbell it might be worth working against https://github.com/mozilla-conduit/lando now, which is the new version of Lando we're going to switch to soon as part of the migration to Git.

The tip I got from Zeid, was that since I am only interested in Try and that will stick around on lando-api after the initial git migration that putting it here is probably what we want (and then we can think about if we still want to migrate this to the django lando).

shtrom:
> The tip I got from Zeid, was that since I am only interested in Try and that will stick around on lando-api after the initial git migration

Ah, right you are. Yep, that makes sense.

zzzeid:
> @zzzeid , I haven't set up a local test environment for this, but this is what I was imagining for a simple api for getting commit id. Then I can local try pushes generate a treeherder link that uses this API to resolve.

@moztcampbell sounds good, I will take a look at this PR today. Do you mind filing a bug for this and mentioning it in the PR title?

landoapi/api/landing_jobs.py
```diff
@@ -13,6 +13,36 @@
...
+        )
+
+    if landing_job.status in (LandingJobStatus.LANDED,):
+        return {
+            "id": landing_job.id,
+            "status": landing_job.status.value,
+            "commit_id": landing_job.landed_commit_id,
+        }
+    else:
+        return {"id": landing_job.id, "status": landing_job.status.value}
```
zzzeid, line 43: We could also disregard the status and return everything here, since `landing_job.landed_commit_id` will be an empty string in cases when there is a failure. This will keep the response more consistent.
```suggestion
    return {
        "id": landing_job.id,
        "status": landing_job.status.value,
        "commit_id": landing_job.landed_commit_id,
    }
```

zzzeid (changes requested):
Looks good, just a minor comment. Please file a bug also and mention it in the PR title so that it will be included in the commit message when we squash and merge.

I can also deploy this to our dev server so we can test it out with a real job.

landoapi/api/landing_jobs.py
moztcampbell: agreed

zzzeid (approved):

PR #460, opened 2025-03-14
phabricator: use POST method to be HTTP spec-compliant (Bug 1952541)

jbuck:
Some CDNs and WAFs block HTTP GET requests with bodies as security
measure. This patch changes the HTTP method to POST, which is spec
compliant. https://www.rfc-editor.org/rfc/rfc9110#name-get


zzzeid (approved):
I think we still need to do an end-to-end test with this as it has not yet been done. Can probably do this on https://hg.mozilla.org/conduit-testing/test-repo/.

PR #463, opened 2025-04-29
try_push: add support for passing Git base SHAs (Bug 1960791)

cgsheeh:
Add support for passing a Git base SHA in the try push
API endpoint. A new `base_commit_format` field is added
to the request body, where the client can pass `git` or `hg`
as the base commit format.

When a landing job is processed with a `target_commit_hash_vcs`
of `git`, it will use `git cinnabar git2hg` to convert the
Git SHA to the equivalent Mercurial SHA. To facilitate this
lookup, we install `git-cinnabar` in the Docker image and
add basic Git repo handling capabilities to the `HgRepo`
class. The class will add a Git repo from a `native_git_source`
specified on the `Repo` object, and when passed it will clone
a Git repo in a sibling directory to the main Mercurial repo
from the source. After the initial cloning, we will fetch
data using Cinnabar grafting to enable lookups of SHAs with
`git cinnabar git2hg`.


install_git-cinnabar.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+curl https://raw.githubusercontent.com/glandium/git-cinnabar/master/download.py -o download.py
+chmod u+x download.py
+./download.py --exact 0.7.0beta2
```
shtrom, line 5: There's a version 7.2 now.

tests/test_try.py
```diff
@@ -501,3 +501,97 @@ def test_try_api_success_gitformatpatch(
...
+    monkeypatch.setattr(HgRepo, "update_cinnabar_repo", lambda self: None)
+
+    response = client.post(
+        "/try/patches", json=try_push_json, headers=auth0_mock.mock_headers
+    )
+
+    assert response.status_code == 201
+    job = LandingJob.query.get(response.json["id"])
+    assert job.target_commit_hash == fake_git_sha
+    assert job.target_commit_hash_vcs == "git"
```
shtrom, line 579: Is there a way we can `assert` that the `convertedhgsha1234567890` is where we want it to be?

install_git-cinnabar.sh
cgsheeh: @shtrom I copied this from [the `git-hg-sync` repo](https://github.com/mozilla-conduit/git-hg-sync/blob/main/install_git-cinnabar.sh), we may want to update this there as well.

tests/test_try.py
cgsheeh: I removed this test, it wasn't adding any value. It will be hard to test the converted SHA without actually creating a cinnabar clone of the repo.

install_git-cinnabar.sh
shtrom: I know you did (:

and, yes.

shtrom (approved):
We may want to have a happy-path test of the git -> hg hash conversion, but it's not a blocker to land this now.

PR #465, opened 2025-05-12
hooks: add CORS header (bug 1965576)

zzzeid:


landoapi/hooks.py
```diff
@@ -82,6 +82,7 @@ def set_app_wide_headers(response: Response) -> Response:
         csp.append("report-uri {}".format(report_uri))
 
     response.headers["Content-Security-Policy"] = "; ".join(csp)
+    response.headers["Access-Control-Allow-Origin"] = "*"
```
globau, line 85: Shouldn't this be limited to just the REST api endpoints?

globau: I missed this is `lando-api`, not `lando` - everything's an API here :)

PR #466, opened 2025-05-27
cinnabar: check if `git2hg` returns the null SHA and defer jobs (Bug 1964576)

cgsheeh:
When converting the passed Git SHA to a Mercurial SHA, Lando-API
currently does nothing to assert the returned SHA is valid.
Add a check that the returned SHA is not the null SHA, and
raise a custom exception if it is. If a landing job hits this
exception, consider it a temporary failure and defer the job
until later.

In testing this patch I realized we do not install `git-cinnabar`
in the `Dockerfile-dev` file, so add that now to allow `git2hg`
to actually fail in the test.

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -33,6 +33,11 @@ RUN addgroup --gid 10001 app \
         --gecos "app,,," \
         app
 
+# Install `git-cinnabar`.
+COPY install_git-cinnabar.sh .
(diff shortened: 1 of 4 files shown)


shtrom (approved):

zzzeid (approved):
We'll need to also port this over to the new Lando.

PR #469, opened 2026-03-10
Add ability to specify try repo

dandarnell:
With Thunderbird's transition over to git, we'd like to add support for using `mach try` on Thunderbird patches.

https://phabricator.services.mozilla.com/D278840

### DIFF
--- a/landoapi/api/try_push.py
+++ b/landoapi/api/try_push.py
@@ -153,7 +153,7 @@ def post_patches(data: dict):
     patch_format = PatchFormat(data["patch_format"])
 
     environment_repos = get_repos_for_env(current_app.config.get("ENVIRONMENT"))
-    try_repo = environment_repos.get("try")
+    try_repo = environment_repos.get(data.get("try_repo"), "try")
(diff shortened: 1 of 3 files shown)


zzzeid:
Note that we have actually started using the new (non-legacy) [Lando](https://github.com/mozilla-conduit/[lando](https://github.com/mozilla-conduit/lando/)/) for try pushes, and it would be better to add this functionality to the new Lando instead of the legacy one. Please file a bug for us to track configuration changes. Any code changes should also go in the new Lando repo.

dandarnell:
> Note that we have actually started using the new (non-legacy) [Lando](https://github.com/mozilla-conduit/%5Blando%5D(https://github.com/mozilla-conduit/lando/)/) for try pushes, and it would be better to add this functionality to the new Lando instead of the legacy one. Please file a bug for us to track configuration changes. Any code changes should also go in the new Lando repo.

I've rewritten the patch (hopefully) correctly for the current version of lando: https://github.com/mozilla-conduit/lando/pull/1019

PR #470, opened 2026-06-25
treestatus: add request mode to redirect reads to new-Lando (Bug 1984161)

cgsheeh:
Add a `TREESTATUS_REQUEST_MODE` configuration variable controlling how
old-Lando serves Treestatus GET requests during the migration to new-Lando:
allow the request, redirect it to the namespaced new-Lando API, or hard-block
it.


landoapi/api/treestatus.py
```diff
@@ -38,6 +43,77 @@
 
 TREE_SUMMARY_LOG_LIMIT = 5
 
+# Base URL of the new-Lando Treestatus API. The new API mirrors the old one,
+# served under a namespaced path on the new host (see bug 1984161).
+TREESTATUS_NEW_BASE_URL = "https://lando.moz.tools/treestatus"
```
shtrom: Could we set this via the env instead?

landoapi/api/treestatus.py
```diff
@@ -38,6 +43,77 @@
...
+
+    The current request path and query string are preserved and appended to the
+    new-Lando base URL.
+    """
+    target = f"{TREESTATUS_NEW_BASE_URL}{request.path}"
+    if request.query_string:
+        target = f"{target}?{request.query_string.decode()}"
+
+    logger.info(f"Redirecting Treestatus request for {request.path} to {target}.")
+    return redirect(target, code=302)
```
shtrom, line 114: Should this be a 308, to retain the method and request payload?

Unless we only expect GETs

shtrom (changes requested):
LGTM, but requesting change so the `TREESTATUS_NEW_BASE_URL` is configurable.

landoapi/api/treestatus.py
zzzeid: Env, or even DB configuration variable which would be more versatile. We should implement this in a way that we can deploy and test on lower environments.

landoapi/api/treestatus.py
```diff
@@ -38,6 +43,77 @@
...
+
+
+def block_treestatus_request():
+    """Raise a `ProblemException` that hard-blocks the current request."""
+    logger.info(f"Hard-blocking Treestatus request for {request.path}.")
+    raise ProblemException(
+        503,
```
zzzeid, line 121: Not sure `503` is correct here, the description implies that this is a permanent state.

landoapi/api/treestatus.py
```diff
@@ -38,6 +43,77 @@
...
+        f"This Treestatus API has migrated to {TREESTATUS_NEW_BASE_URL}.",
+        type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503",
+    )
+
+
+def treestatus_request_mode(handler: Callable) -> Callable:
+    """Apply the configured `TreestatusRequestMode` to a request handler.
+
+    Depending on the `TREESTATUS_REQUEST_MODE` configuration variable, allow the
+    request through to `handler`, redirect it to new-Lando, or hard-block it.
```
zzzeid, line 132: Can you elaborate on "hard-block"? Seems like this is implying a terminal state based on some of the hints but it's not very clear.

landoapi/api/treestatus.py
```diff
@@ -38,6 +43,77 @@
...
+    logger.info(f"Hard-blocking Treestatus request for {request.path}.")
+    raise ProblemException(
+        503,
+        "Treestatus has moved.",
+        f"This Treestatus API has migrated to {TREESTATUS_NEW_BASE_URL}.",
+        type="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503",
+    )
+
+
+def treestatus_request_mode(handler: Callable) -> Callable:
```
zzzeid: nit: This could be named better, maybe `deprecation_handler` or something more descriptive.

landoapi/models/configuration.py
```diff
@@ -25,6 +25,24 @@ class ConfigurationKey(enum.Enum):
...
+    """
+
+    # Allow the request to be served by old-Lando.
+    ALLOW = "allow"
+
+    # Redirect the request to the new-Lando Treestatus API.
+    REDIRECT = "redirect"
+
+    # Hard-block the request with an error response.
+    BLOCK = "block"
```
zzzeid: This is in the wrong spot, it should go in `api.treestatus` or somewhere else. This module is for things specific to configuration variables. See also: other comment about removing this.

landoapi/api/treestatus.py
```diff
@@ -38,6 +43,77 @@
...
+        if mode == TreestatusRequestMode.REDIRECT:
+            return redirect_to_new_treestatus()
+
+        if mode == TreestatusRequestMode.BLOCK:
+            return block_treestatus_request()
+
+        return handler(*args, **kwargs)
+
+    return wrapper
+
```
zzzeid: This (along with `get_treestatus_mode`) is a little over-engineered, and along with the comment about configuring the treestatus URL, we could have two variables: `TREE_STATUS_REDIRECT_URL` and `TREESTATUS_ON` and deal with all the cases with those two variables. We also don't need an enum in that case since everything is already parsed as a boolean within the `ConfigurationVariable` class.

```python
def treestatus_request_mode(handler: Callable) -> Callable:
    @functools.wraps(handler)
    def wrapper(*args, **kwargs):
        is_on = ConfigurationVariable.get(
            ConfigurationKey.TREESTATUS_ON,
            True,
        )
        redirect_url = ConfigurationVariable.get(
            ConfigurationKey.TREESTATUS_REDIRECT_URL,
            "",
        )

        if not is_on:
            return block_treestatus_request()
        elif redirect_url:
            return redirect_to_new_treestatus()
        return handler(*args, **kwargs)   
    return wrapper
```

landoapi/api/treestatus.py
```diff
@@ -38,6 +43,77 @@
...
+        TreestatusRequestMode.ALLOW.value,
+    )
+    try:
+        return TreestatusRequestMode(raw_mode)
+    except ValueError:
+        logger.error(
+            f"Unrecognized Treestatus request mode {raw_mode!r}, defaulting to "
+            f"{TreestatusRequestMode.ALLOW.value!r}."
+        )
+        return TreestatusRequestMode.ALLOW
```
zzzeid, line 100: This can be removed by implementing the other comments.

zzzeid (changes requested):
There should be a command that sets the variables via the CLI based on input params, since we don't have an admin in legacy.

EDIT: to clarify, command should be able to do the following:
- Turn on treestatus
- Turn on "redirect mode" (i.e., turn on treestatus and set redirect URL)
- Turn off "redirect mode" (i.e., un-set redirect URL)*
- Turn off treestatus

\* not sure if this will ever be needed, but if the migration goes bad and we need to roll back, it would be useful.

landoapi/api/treestatus.py
cgsheeh: Makes sense, moved the config to be configurable via the DB, with the prod URL as a fallback. This way we can test on dev in advance and we don't need to deploy to update or change the URL.

landoapi/api/treestatus.py
cgsheeh: We are only using this on GET endpoints.

landoapi/api/treestatus.py
cgsheeh: By "hard-blocks" we mean the request isn't processed and isn't redirected. We explicitly return an error to the requestor. This is the "temporary hard-shutdown" state I mentioned in the bug. We flip the mode to `BLOCK` once the main Treestatus consumers are migrated over, and if anything breaks, we can flip back to "redirect" and ask them to update the Treestatus base URL on their end.

landoapi/models/configuration.py
cgsheeh: Done.

landoapi/api/treestatus.py
cgsheeh: I think the three-state enum is worth keeping: it lets us switch behavior with a single variable that is separate from the config, and the states map directly to how we've been describing the migration steps in the bug. With the configurable redirect URL now in place via the DB, I think the only real difference left is enum-vs-boolean. Happy to dig into specific drawbacks you see, but for a value checked in one 3-branch if, the mode variable suits this cleanly.

landoapi/api/treestatus.py
cgsheeh: Done.

landoapi/api/treestatus.py
zzzeid: Implementation details don't have to map to the bug steps/instructions (at least I would consider that too low level/not relevant) especially with a command (i.e., front-end) that matches what the bug steps are. I would maintain that it adds unnecessary complexity and code.

landoapi/models/configuration.py
```diff
@@ -25,6 +25,8 @@ class ConfigurationKey(enum.Enum):
     LANDING_WORKER_STOPPED = "LANDING_WORKER_STOPPED"
     API_IN_MAINTENANCE = "API_IN_MAINTENANCE"
     WORKER_THROTTLE_SECONDS = "WORKER_THROTTLE_SECONDS"
+    TREESTATUS_REQUEST_MODE = "TREESTATUS_REQUEST_MODE"
+    TREESTATUS_NEW_BASE_URL = "TREESTATUS_NEW_BASE_URL"
```
zzzeid: nit: I think using "redirect" in the name is more clear, "new" is a bit ambiguous. You also use redirect in most other parts of the new code.

zzzeid (approved):
> Added a command to make things foolproof, and since this service is dying soon it seems worth it anyways. :)

> ... I think the three-state enum is worth keeping: it lets us switch behavior with a single variable that is separate from the config ...

This is adding complexity and unnecessary codepaths that can be wrapped/dealt with more elegantly in the cli command, but this is all getting decommissioned soon so I won't block on this.

shtrom (approved):
lgtm!

